PR #14, opened 2024-08-29
Setup trunk and some formatting

ClementWalter:


pyproject.toml
```diff
@@ -0,0 +1,15 @@
+[tool.poetry]
+name = "cairo_programs"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
```
tcoratger, line 6: Here we keep it random at the moment?

tcoratger (dismissed):

pyproject.toml
ClementWalter: authors is required in pyproject.toml, but I don't like to put authors in plain text as authors are de facto contributors to me. So I propose to just leave it empty. Fixed the description to match the cargo.toml one

tcoratger (approved):

tcoratger (approved):

Issue #39, opened 2024-09-25
feat: implement missing hints in rust


ClementWalter:
See corresponding python implementation, to go from a rust/python `Block` and `State` to a cairo `model.Block*` and `model.State*`

PR #40, opened 2024-09-25
Add Block and State deserialization

ClementWalter:
# Add BlockHeader deserialization

- Add BlockHeader deserialization to the os module
- Add tests for the BlockHeader deserialization

The rust test has been skipped for now, as the hints are not yet implemented in the cairo vm.
Issue to follow up: https://github.com/kkrt-labs/keth/issues/39


crates/exex/src/exex.rs
```diff
@@ -321,38 +187,6 @@ mod tests {
...
-                max_priority_fee_per_gas: 0x4903a597,
-                to: reth_primitives::TxKind::Call(
-                    Address::from_str("0xf3de3c0d654fda23dad170f0f320a92172509127").unwrap(),
-                ),
-                value: U256::from_str("0xb1a2bc2ec50000").unwrap(),
-                access_list: Default::default(),
-                input: Bytes::from_str("0x9871efa4000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b1a2bc2ec50000000000000000000000000000000000000000000000000009f7051a01fa559ee400000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000001b0000000000000003b6d0340cab7ab9f1a9add91380a0e8fae700b65f320e667").unwrap(),
-            }),
-        };
-
```
tcoratger, line 355: I'm not in favor of removing this part because it's a mock block with a transaction in it, taken directly from mainnet, which allows us to test the ExEx for all our implementations without having to turn anything on just with the unit test to save a lot of time.

crates/exex/src/exex.rs
```diff
@@ -381,7 +215,7 @@ mod tests {
 
         // Create a sealed block with a single transaction
         let block = SealedBlockWithSenders {
-            block: SealedBlock { header: header.seal_slow(), body: vec![tx], ..Default::default() },
+            block: SealedBlock { header: header.seal_slow(), body: vec![], ..Default::default() },
```
tcoratger, line 218: Likewise, I am in favor of keeping a transaction in the block so that we can test everything in real conditions. The ideal would even be to create a follow up issue to have an input JSON with a dozen real blocks in order to always have concrete, simple and easy to turn on unit tests to test that everything passes in Mainnet conditions.

crates/exex/src/exex.rs
```diff
@@ -416,25 +250,6 @@ mod tests {
...
-        let sender_account_info =
-            db.account(address!("6a3cA5811d2c185E6e441cEFa771824fb355f9Ec"))?;
-        assert_eq!(
-            sender_account_info.unwrap().balance,
-            // Initial balance - value - gas used
-            U256::from(ETH_TO_WEI) -
-                U256::from_str("0xb1a2bc2ec50000").unwrap() -
-                U256::from_str("0xe826f9395cd0").unwrap()
-        );
-
```
tcoratger, line 437: I am not in favor of removing this. It is a succession of assertions that guarantee that our ExEx and the re-execution of the blocks work correctly. Without it we are not sure of what we are doing. The goal is to verify for example that at the output, after the transaction, the balance of the sender is well updated.

crates/exex/src/hints.rs
```diff
@@ -110,85 +94,6 @@ impl Hint {
...
-                        block.base_fee_per_gas.unwrap_or_default(),
-                    )),
-                ],
-            )
-            .map_err(HintError::Memory)?;
-
-            Ok(())
-        },
-    )
-}
```
tcoratger, line 190: We don't need this hint anymore? It was linked to a very recent issue here no? https://github.com/kkrt-labs/keth/issues/23 (the last one I did if I remember well)

crates/exex/src/hints.rs
```diff
@@ -110,85 +94,6 @@ impl Hint {
...
-            if let Some(block) = latest_block {
-                for tx in &block.body {
-                    println!("Block: {}, transaction hash: {}", block.number, tx.hash());
-                }
-            }
-
-            Ok(())
-        },
-    )
-}
```
tcoratger, line 148: If possible I am rather in favor of creating a follow up issue to remove this hint so as not to mix it up with all the rest of this implementation which is already substantial and thus not to inadvertently remove something which could be useful to us.

crates/exex/src/exex.rs
```diff
@@ -197,73 +132,6 @@ impl<Node: FullNodeComponents> KakarotRollup<Node> {
...
-                chain_tip,     // The current chain tip.
-                trace,         // The extracted execution trace.
-                memory,        // The relocated memory.
-                public_input,  // The public input for proving.
-                private_input, // The private input for proving.
-            )?;
-        }
-
-        Ok(())
-    }
```
tcoratger, line 266: We don't need that anymore either? Seems like this was also linked to a very recent issue here https://github.com/kkrt-labs/keth/issues/23 no?

crates/exex/src/exex.rs
```diff
@@ -70,116 +68,53 @@ impl<Node: FullNodeComponents> KakarotRollup<Node> {
...
+
+                // Commit the execution trace to the database
+                self.commit_cairo_execution_traces(
+                    committed_chain.tip().number,
+                    trace,
+                    memory,
+                    public_input,
+                    private_input,
+                )?;
             }
```
tcoratger, line 118: This function was very important, allowing to re-execute our blocks via revm as long as we do not have the run via our own evm. It allows us to run the process as a whole: ExEx + re-execution + hint in any program. It seems early to remove it, right? For me it takes away a test environment and I think we could keep it for now.

crates/exex/src/exex.rs
```diff
@@ -70,116 +68,53 @@ impl<Node: FullNodeComponents> KakarotRollup<Node> {
...
-                        trace,
-                        memory,
-                        public_input,
-                        private_input,
-                    )?;
-                }
-            }
-        }
+                // Load the cairo program from the file
+                let program = std::fs::read(PathBuf::from("../../cairo/programs/os.json"))?;
```
tcoratger, line 82: If we put this program it means that from now on we only do tests on our OS and nothing else correct, like no small random programs? I had the impression that the approach with the Paths vector was perhaps more modular for the moment in the sense that we could enter whatever we wanted in the `start` to test lots of different programs (even if it meant only entering a vector of an element with the OS path if we only want to test that). But hardcoding a path in the function seems a bit restrictive to me, no?

ClementWalter: the idea is that from the exex pov you don't really want to test cairo programs in the context of the exex but hints. these hints should be tested but probable in standalone cairo_run, no need to have the whole exex setup

on the other hand, on the exex we want to make sure that the real program (yet under construction) is working well

crates/exex/src/exex.rs
ClementWalter: I may have misunderstood some part, is it required to re-run with revm to have the initial partial state?

crates/exex/src/exex.rs
ClementWalter: agree that I went to fast with the removal of the tx, fixed

crates/exex/src/exex.rs
ClementWalter: I don't get what the db does tbh, and I understood from slack that it's meant to be removed. I'd like to have the real workflow up (with a in-progress cairo program). Everything else should be removed

crates/exex/src/exex.rs
ClementWalter: block_info was a initial attemps but since then I've better scoped and understood the problem. The actual os.cairo reflects what's required, everything else so far was like warmup

crates/exex/src/hints.rs
ClementWalter: no, cf previous answer. you can recylce some part I guess, but the actual hint is not this one (see corresponding block hint in the PR)

crates/exex/src/exex.rs
tcoratger: No, not really, in fact the process so far was as follows:
- We receive a block with transactions via the ExEx that we mock for our tests (that is to say that we voluntarily inject a real block from the mainnet with a real transaction in it).
- We take this block and we re-execute here the transactions that are inside via the revm (which will be done in the future by our OS that will just have to be injected here when it is ready), which allows us to do all kinds of tests or executions in parallel.

I know that one day we will have to remove this function because our goal is obviously not to reuse revm but it seems a bit early to remove it, wdyt?

crates/exex/src/hints.rs
ClementWalter: transactions are not part of the `model.Block` struct

crates/exex/src/exex.rs
tcoratger: Here in fact the db is just a way to temporarily store things but this test seems important to me because it is the complete flow:
- We mock a block with transactions.
- We send it to the ExEx to simulate a real block that has just been executed.
- We do things on it (in the future we re-execute via our OS)
- We check the results (which are currently stored in a small db but can be stored in a structure or any other means).

crates/exex/src/exex.rs
tcoratger: Ok then let us remove and I'll reuse some parts

crates/exex/src/hints.rs
tcoratger: Ok will recycle and modify this in the follow up Pr after we merge this one

tcoratger (approved):

PR #42, opened 2024-09-30
add block hint

tcoratger:


cairo/programs/os.json
ClementWalter, line 1: I think that you recompiled and didn't `fmt` afterwards. We can add this step to the `compiled_cairo.py` utils to make this easier (happened to me few times also)

crates/exex/src/hints.rs
```diff
@@ -112,3 +115,152 @@ pub fn add_segment_hint() -> Hint {
...
+        String::from("block"),
+        move |vm: &mut VirtualMachine,
+              _exec_scopes: &mut ExecutionScopes,
+              ids_data: &HashMap<String, HintReference>,
+              ap_tracking: &ApTracking,
+              _constants: &HashMap<String, Felt252>|
+              -> HintExecutionResult {
+            // We retrieve the `env` pointer from the `ids_data` hashmap.
+            // This pointer is used to store the block-related values in the VM.
+            let env_ptr = get_ptr_from_var_name("block", vm, ids_data, ap_tracking)?;
```
ClementWalter, line 133:
```suggestion
            // We retrieve the `model.Block*` pointer from the `ids_data` hashmap.
            // This pointer is used to store the block-related values in the VM.
            let block_ptr = get_ptr_from_var_name("block", vm, ids_data, ap_tracking)?;
```

crates/exex/src/hints.rs
```diff
@@ -112,3 +115,152 @@ pub fn add_segment_hint() -> Hint {
...
+                &header.uncle_hash.0[0..16],
+            )));
+
+            block_header.push(MaybeRelocatable::from(Felt252::from_bytes_be_slice(
+                &header.withdrawals_root.unwrap().0[16..],
+            )));
+            block_header.push(MaybeRelocatable::from(Felt252::from_bytes_be_slice(
+                &header.withdrawals_root.unwrap().0[0..16],
+            )));
+
```
ClementWalter, line 232: I'd like us to work on types, so I'd like to have a. `block.into()` here in the hints and all required conversion in a dedicated trait

ClementWalter: we have a `trait KethBlockHeader`, `KethBlock` that implements `from` for various input types as you mentioned these ones and the ef tests ones

crates/exex/testdata/block.json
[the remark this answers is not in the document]
tcoratger: Yes just removed this and used instead

```rust
let block = random_block(
    &mut rng,
    10,
    BlockParams {
        parent: Some(parent),
        tx_count,
        withdrawals_count,
        ommers_count,
        requests_count,
    },
);
```

This generates a primitive `SealedBlock` type.

crates/exex/src/hints.rs
tcoratger: Just added the `pub trait KethBlock` trait to handle the three conversions:
- From reth header to a vec of MaybeRelocatable that can be understood by the cairo vm
- From reth block body to a vec of MaybeRelocatable that can be understood by the cairo vm
- Full conversion that includes Header + Body

cairo/programs/os.json
tcoratger: Just fixed the fmt, yes let us do that in a follow up dedicated pr, maybe a trunk fmt at the end of the script or something like that, wdyt?

cairo/programs/block.cairo
ClementWalter, line 1: be careful, this is not the same as the hint in the `os.cairo` where the segment allocation is done in the hint directly

cairo/programs/block.cairo
```diff
@@ -0,0 +1,17 @@
...
+
+from starkware.cairo.common.memcpy import memcpy
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+
+from src.model import model
+
+func get_block() -> model.Block* {
+    tempvar block = cast(nondet %{ segments.add() %}, model.Block*);
+    %{ block %}
+    return block;
```
ClementWalter, line 11:
```suggestion
    tempvar block: model.Block*;
    %{ block %}
    return block;
```

cairo/programs/block.cairo
ClementWalter: no no, I mean the `test_os.cairo` in `cairo/tests/programs/test_os.cairo`, and you call `test_block` entrypoint, the vm runner should have a way to tell the entrypoint name (`main` is only the default normally)

ClementWalter: the ordre of the builtins can be changed no pb

crates/exex/src/hints.rs
ClementWalter, line 1: I think that we'd like instead to implement `From<SealedBlock>` for a `KethBlock`, and such a `KethBlock` would fit well with the already existing https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/vm_memory/memory_segments.rs#L132-L146 so that it's easier is we update the models

crates/exex/src/hints.rs
```diff
@@ -112,3 +119,188 @@ pub fn add_segment_hint() -> Hint {
...
+              _exec_scopes: &mut ExecutionScopes,
+              ids_data: &HashMap<String, HintReference>,
+              ap_tracking: &ApTracking,
+              _constants: &HashMap<String, Felt252>|
+              -> HintExecutionResult {
+            // We retrieve the `model.Block*` pointer from the `ids_data` hashmap.
+            // This pointer is used to store the block-related values in the VM.
+            let block_ptr = get_ptr_from_var_name("block", vm, ids_data, ap_tracking)?;
+
+            vm.load_data(block_ptr, &block.to_cairo_vm_block()).map_err(HintError::Memory)?;
```
ClementWalter, line 250: here I'd like to use a higher level method consuming a properly defined data type https://github.com/lambdaclass/cairo-vm/blob/main/vm/src/vm/vm_memory/memory_segments.rs#L132-L146

crates/exex/src/hints.rs
```diff
@@ -112,3 +119,188 @@ pub fn add_segment_hint() -> Hint {
...
+                    requests_count,
+                },
+            );
+
+            // Create a new cairo run configuration with the all_cairo layout
+            let config = CairoRunConfig { layout: LayoutName::all_cairo, ..Default::default() };
+
+            // Create a new hint processor with the block hint
+            let mut hint_processor =
+                KakarotHintProcessor::default().with_block(block.clone()).build();
```
ClementWalter, line 295: noob question but why do we need to `.clone()` since we don't reuse the `block` afterwards?

tcoratger: Yes useless here, can be deleted, done.

crates/exex/src/hints.rs
tcoratger: Ok so I used simply:

```rust
// This call will first add a new memory segment to the VM (the base)
// Then we load the block into the VM starting from the base
vm.gen_arg(&block.to_cairo_vm_block())?;
```

where the argument will be downcasted automatically to a vec of MaybeRelocatable, it includes the generation of the new segment and the load of the data.

cairo/programs/block.cairo
tcoratger: I tried to use the `test_block` entrypoint but I've an error like: no stop pointer for the pedersen builtin... Seems like another topic maybe better to keep the very simple test for this PR and do a followup to unify everything inside test_os.cairo

crates/exex/src/hints.rs
tcoratger: KethBlock would be a Vec<MaybeRelocatable>?

cairo/programs/block.cairo
ClementWalter, line 1: It'd have been better to import the `test_block` function from `test_os.cairo` but since #43 is supposed to remove this, I won't block the PR for this

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+        String::from("block"),
+        move |vm: &mut VirtualMachine,
+              _exec_scopes: &mut ExecutionScopes,
+              _ids_data: &HashMap<String, HintReference>,
+              _ap_tracking: &ApTracking,
+              _constants: &HashMap<String, Felt252>|
+              -> HintExecutionResult {
+            // This call will first add a new memory segment to the VM (the base)
+            // Then we load the block into the VM starting from the base
+            vm.gen_arg(&Into::<KethPayload>::into(block.clone()).0)?;
```
ClementWalter, line 361: where do you actually assign the base to the `block` cairo variable here?

ClementWalter: rust question: why do we need to clone?

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+    fn from(value: u64) -> Self {
+        MaybeRelocatable::from(Felt252::from(value)).into()
+    }
+}
+
+impl From<Option<u64>> for KethPayload {
+    fn from(value: Option<u64>) -> Self {
+        MaybeRelocatable::from(Felt252::from(value.unwrap_or_default())).into()
+    }
+}
```
ClementWalter, line 212: not sure where we have `Option<u64>` in blocks, but this would be converted into a `model.Option` in cairo, meaning a struct with 2 values `{is_some: 0, value: 0}`

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+    fn from(value: usize) -> Self {
+        MaybeRelocatable::from(Felt252::from(value)).into()
+    }
+}
+
+impl From<Bloom> for KethPayload {
+    fn from(value: Bloom) -> Self {
+        vec![
+            MaybeRelocatable::from(Felt252::from(value.len())),
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value.0 .0)),
```
ClementWalter, line 224: bloom is a  256 long bytes array, so a vec of felt

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+}
+
+impl From<U256> for KethPayload {
+    fn from(value: U256) -> Self {
+        MaybeRelocatable::from(Felt252::from_bytes_be_slice(
+            &value.to_be_bytes::<{ U256::BYTES }>(),
+        ))
+        .into()
+    }
+}
```
ClementWalter, line 243: the `u256` is not trivial as it can be either `Uint256*` (so a pointer to a new segment with two values) or directly inlined into the struct, and consequently split in `_low` and `_high` parts

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+
+impl From<Bytes> for KethPayload {
+    fn from(value: Bytes) -> Self {
+        vec![
+            MaybeRelocatable::from(Felt252::from(value.len())),
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value.0)),
+        ]
+        .into()
+    }
+}
```
ClementWalter, line 253: bytes are an array of felt, not a single felt

crates/exex/src/hints.rs
tcoratger: - For the base, this is done here no? https://github.com/lambdaclass/cairo-vm/blob/7d19956781901fb05c76f7c02fa82bd907f0f9a2/vm/src/vm/vm_memory/memory_segments.rs#L136 we add a new segment which is automatically set as the base for the block, is there anything else to set?
- For the clone, this is a closure question, for more explicitness, the exact error without cloning is `cannot move out of `block`, a captured variable in an `Fn` closure
move occurs because `block` has type `SealedBlock`, which does not implement the `Copy` trait`

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+
+impl From<Vec<u8>> for KethPayload {
+    fn from(value: Vec<u8>) -> Self {
+        vec![
+            MaybeRelocatable::from(Felt252::from(value.len())),
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value)),
```
ClementWalter, line 259: same, needs to be an array

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+
+impl From<B256> for KethPayload {
+    fn from(value: B256) -> Self {
+        vec![
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value.0[16..])),
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value.0[0..16])),
+        ]
+        .into()
+    }
+}
```
ClementWalter, line 273: see comment, it depends on the target model

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+
+impl From<Option<B256>> for KethPayload {
+    fn from(value: Option<B256>) -> Self {
+        value.unwrap_or_default().into()
+    }
+}
```
ClementWalter, line 279: see above for `Option`

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+
+impl From<Signature> for KethPayload {
+    fn from(value: Signature) -> Self {
+        vec![
+            // The length of the signature payload
+            MaybeRelocatable::from(Felt252::from(value.payload_len())),
+            // The signature payload
+            MaybeRelocatable::from(Felt252::from_bytes_be_slice(&value.to_bytes())),
```
ClementWalter, line 287: see above, is this a `vec`?

crates/exex/src/hints.rs
```diff
@@ -112,3 +123,294 @@ pub fn add_segment_hint() -> Hint {
...
+            // Create a new hint processor with the block hint
+            let mut hint_processor = KakarotHintProcessor::default().with_block(block).build();
+
+            // Load the cairo program from the file
+            let program = std::fs::read(PathBuf::from("../../cairo/programs/block.json")).unwrap();
+
+            // Run the cairo program with the hint processor and the block
+            //
+            // We unwrap the result to ensure that the program ran successfully
+            let _ = cairo_run(&program, &config, &mut hint_processor).unwrap();
```
ClementWalter, line 413: because the program does nothing, there is no chance to have this run failing. (you could remove the hint it'd be the same)

crates/exex/src/hints.rs
tcoratger: Here I put the case where this is always inlined into the struct

PR #45, opened 2024-10-02
model: impl `KethMaybeRelocatable` and `KethOption`

tcoratger:


.tool-versions
```diff
@@ -0,0 +1 @@
+scarb nightly-2024-09-28
```
tcoratger, line 1:
```suggestion
```

crates/exex/src/model.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+impl From<u64> for KethMaybeRelocatable {
+    fn from(value: u64) -> Self {
+        Self(Felt252::from(value).into())
+    }
+}
+
+/// [`KethOption`] is a custom representation of a Rust [`Option<u64>`] type, specifically tailored
```
ClementWalter, line 70: can't we make this `Option<T: Into<KethMaybeRelocatable>>` ?

ClementWalter (approved):
left a comment to be more generic but ok as is

crates/exex/src/model.rs
tcoratger: Yes good idea, just fixed

tcoratger:
> left a comment to be more generic but ok as is

Just fixed the comment, this is nicer

ClementWalter (approved):

PR #46, opened 2024-10-02
model: implement `KethU256`

tcoratger:

### DIFF
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -1,4 +1,5 @@
 use cairo_vm::{types::relocatable::MaybeRelocatable, Felt252};
+use reth_primitives::B256;
 use serde::{Deserialize, Serialize};
 
 /// A custom wrapper around [`MaybeRelocatable`] for the Keth execution environment.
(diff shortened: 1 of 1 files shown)


crates/exex/src/model.rs
```diff
@@ -53,6 +54,31 @@ impl KethMaybeRelocatable {
...
+    ///
+    /// ```
+    /// use kakarot_exex::model::KethMaybeRelocatable;
+    ///
+    /// let bytes: &[u8] = &[0x00, 0x01, 0x02, 0x03]; // Example big-endian byte array
+    /// let keth_value = KethMaybeRelocatable::from_bytes_be_slice(bytes);
+    /// ```
+    pub fn from_bytes_be_slice(bytes: &[u8]) -> Self {
+        Felt252::from_bytes_be_slice(bytes).into()
+    }
```
ClementWalter, line 81: I'm not sure that we actually need this, this would mean that we have data in blocks/state that is a `[u8]` and that we want to data to be loaded in Cairo as a `felt`. I can't see any use case for this (apart from using this together with `u64 to bytes_be for example, but in this latter context it may be better to directly implement such a method

crates/exex/src/model.rs
```diff
@@ -68,37 +94,43 @@ impl From<u64> for KethMaybeRelocatable {
...
-/// - `value`: The actual value of the option, represented as a [`KethMaybeRelocatable`].
+/// - `value`: The actual value of the option, which can be of any type `T`.
 ///
-/// When the option is [`None`], the `value` field is assigned a default zero value. This struct
+/// When the option is [`None`], the `value` field is assigned a default value. This struct
 /// provides a way to handle optional values in a manner consistent with the Keth model's memory
 /// representation.
 #[derive(Debug, Eq, Ord, Hash, PartialEq, PartialOrd, Clone, Serialize, Deserialize)]
-pub struct KethOption {
+pub struct KethOption<T> {
```
ClementWalter, line 109: as per our discussion, `is_some` could be a `Felt252` directly

crates/exex/src/model.rs
tcoratger: This is directly used by the `From<B256> for KethU256` conversion. Indeed, here we take bytes in big endian order and transform them into `MaybeRelocatable` via this helper, that is to say `KethMaybeRelocatable::from_bytes_be_slice(&value.0[16..])`. This is just an helper.

ClementWalter (approved):

crates/exex/src/model.rs
tcoratger: But in the end to fit into the `gen_arg`, `is_some`, despite being a `Felt` will have to be placed as a `MaybeRelocatable`. So maybe it's better to leave everything like that so as not to mix up things no? That way we can tell that everything has the same type.

PR #47, opened 2024-10-02
Add a first loop over block txs

ClementWalter:
### Why

The os.cairo should run the transactions and return the result of the execution.

### What

Implement an initial loop over all the transactions in the block ; just initial validation.


cairo/programs/os.cairo
```diff
@@ -1,19 +1,113 @@
...
+        assert [range_check_ptr] = tx.max_fee_per_gas;
+        let range_check_ptr = range_check_ptr + 1;
+    }
+
+    with_attr error_message("Transaction gas_limit > Block gas_limit") {
+        assert_nn(header.gas_limit - tx.gas_limit);
+    }
+
+    with_attr error_message("Max fee per gas too low") {
+        assert_nn(tx.max_fee_per_gas - header.base_fee_per_gas);
```
obatirou, line 103: Doesn't header.base_fee_per_gas needs a range check ?
From my understanding block is supplied by hint and could lead overflow

obatirou (approved):

PR #49, opened 2024-10-02
model: impl `From<U256>` for `KethU256`

tcoratger:

### DIFF
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -1,5 +1,5 @@
 use cairo_vm::{types::relocatable::MaybeRelocatable, Felt252};
-use reth_primitives::{Address, B256};
+use reth_primitives::{Address, B256, U256};
 use serde::{Deserialize, Serialize};
 
 /// A custom wrapper around [`MaybeRelocatable`] for the Keth execution environment.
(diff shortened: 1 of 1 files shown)


crates/exex/src/model.rs
ClementWalter, line 1: can't we just do one single impl for

```
impl<T> From<Option<T>> for KethOption<KethMaybeRelocatable>
where
    T: Into<KethMaybeRelocatable>,
{
    fn from(value: Option<T>) -> Self {
        match value {
            Some(value) => KethOption { is_some: KethMaybeRelocatable::one(), value: value.into() },
            None => KethOption::default(),
        }
    }
}
impl<T> From<Option<T>> for KethOption<KethU256>
where
    T: Into<KethU256>,
{
    fn from(value: Option<T>) -> Self {
        match value {
            Some(value) => KethOption { is_some: KethMaybeRelocatable::one(), value: value.into() },
            None => KethOption::default(),
        }
    }
```

like 
```
impl<T> From<Option<T>> for KethOption<U>
where
    T: Into<U>,
{
    fn from(value: Option<T>) -> Self {
        match value {
            Some(value) => KethOption { is_some: KethMaybeRelocatable::one(), value: value.into() },
            None => KethOption::default(),
        }
    }
}
```

ClementWalter (approved):
left a comment but looks ok for this PR purpose

ClementWalter (approved):

PR #50, opened 2024-10-03
model: impl `From<Bloom>` for `KethMaybeRelocatables`

tcoratger:

### DIFF
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -429,6 +429,10 @@ mod tests {
 
             // Assert roundtrip conversion is equal to original value
             prop_assert_eq!(roundtrip_bloom, bloom);
+            prop_assert_eq!(
+                bloom.len(),
+                keth_maybe_relocatables.0[0].0.get_int().unwrap().to_string().parse::<usize>().unwrap()
+            );
         }
     }
 
@@ -528,6 +532,10 @@ mod tests {
 
         // Verify that converting back gives the original Bloom filter
         assert_eq!(KethMaybeRelocatables::to_bloom(&keth_maybe_relocatables), bloom);
+        assert_eq!(
+            bloom.len(),
+            keth_maybe_relocatables.0[0].0.get_int().unwrap().to_string().parse::<usize>().unwrap()
+        );
     }
 
     #[test]
@@ -539,5 +547,9 @@ mod tests {
 
         // Verify that converting back gives the original Bloom filter
         assert_eq!(KethMaybeRelocatables::to_bloom(&keth_maybe_relocatables), bloom);
+        assert_eq!(
+            bloom.len(),
+            keth_maybe_relocatables.0[0].0.get_int().unwrap().to_string().parse::<usize>().unwrap()
+        );
     }
 }
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -1,5 +1,5 @@
 use cairo_vm::{types::relocatable::MaybeRelocatable, Felt252};
-use reth_primitives::{Address, B256, U256};
+use reth_primitives::{Address, Bloom, B256, U256};
 use serde::{Deserialize, Serialize};
 
 /// A custom wrapper around [`MaybeRelocatable`] for the Keth execution environment.
@@ -87,8 +87,20 @@ impl From<Felt252> for KethMaybeRelocatable {
     }
 }
 
+impl From<u8> for KethMaybeRelocatable {
+    fn from(value: u8) -> Self {
+        Self(Felt252::from(value).into())
+    }
+}
+
 impl From<u64> for KethMaybeRelocatable {
     fn from(value: u64) -> Self {
+        Self(Felt252::from(value).into())
+    }
+}
+
+impl From<usize> for KethMaybeRelocatable {
+    fn from(value: usize) -> Self {
         Self(Felt252::from(value).into())
     }
 }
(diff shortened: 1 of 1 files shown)


crates/exex/src/model.rs
```diff
@@ -360,6 +418,22 @@ mod tests {
...
+            // Convert back to Bloom
+            let roundtrip_bloom = KethMaybeRelocatables::to_bloom(&keth_maybe_relocatables);
+
+            // Assert roundtrip conversion is equal to original value
+            prop_assert_eq!(roundtrip_bloom, bloom);
+            prop_assert_eq!(
+                bloom.len(),
+                keth_maybe_relocatables.0[0].0.get_int().unwrap().to_string().parse::<usize>().unwrap()
+            );
+        }
```
tcoratger, line 436: @ClementWalter This test is quite heavy (few seconds to run) because the fuzzing on this 256 bits elements is quite huge to generate. To cover a little more I put below an additional test with `Bloom::random()` so we could "in theory" remove the proptest but I hesitate because the fuzzing is quite important on this tricky conversion, what do you think?

crates/exex/src/model.rs
```diff
@@ -360,6 +418,22 @@ mod tests {
...
+            let keth_maybe_relocatables = KethMaybeRelocatables::from(bloom);
+
+            // Convert back to Bloom
+            let roundtrip_bloom = KethMaybeRelocatables::to_bloom(&keth_maybe_relocatables);
+
+            // Assert roundtrip conversion is equal to original value
+            prop_assert_eq!(roundtrip_bloom, bloom);
+            prop_assert_eq!(
+                bloom.len(),
+                keth_maybe_relocatables.0[0].0.get_int().unwrap().to_string().parse::<usize>().unwrap()
```
ClementWalter, line 434: isn't there any easier way to go from a felt to a `usize`?

crates/exex/src/model.rs
ClementWalter: So just confirming that a `Bloom` is 256 bytes long, ie 32bytes * 8 words
So in cairo this will be a Uint256*, and in terms of segment actually 16 felts

crates/exex/src/model.rs
```diff
@@ -87,18 +87,64 @@ impl From<Felt252> for KethMaybeRelocatable {
...
+
+impl From<Bloom> for KethMaybeRelocatables {
+    /// Converts a [`Bloom`] filter into a [[`KethMaybeRelocatables`]] structure.
+    ///
+    /// The [`Bloom`] filter is represented as a 256-byte array in big-endian order. However,
+    /// CairoVM's [`Felt252`] type can only handle values up to 252 bits. As a result, we need to
+    /// convert the 256-byte array into smaller pieces that fit within this limit.
+    ///
+    /// Here's how the conversion works:
+    /// - The first element of the `Vec<KethMaybeRelocatable>` is the length of the Bloom filter.
```
ClementWalter, line 135: I don't think we need the length here since it's always 8 Uint256

crates/exex/src/model.rs
```diff
@@ -87,18 +87,64 @@ impl From<Felt252> for KethMaybeRelocatable {
...
+    /// - The subsequent elements are the bytes of the Bloom filter, reversed to represent the data
```
ClementWalter, line 136: I think that you should iter into a KethU256

crates/exex/src/model.rs
ClementWalter: in terms of testing, can't we just ask for few examples and not a lot ?

crates/exex/src/model.rs
tcoratger: Not from what I know but we could add some helpers in this repo https://github.com/starknet-io/types-rs/blob/main/crates/starknet-types-core/src/felt/mod.rs

crates/exex/src/model.rs
tcoratger: In my new structure that I just pushed, I put 2 fields (len and data) so that we can convert any data and we could use it generically. If we don't want to use it for bloom we will just have to ignore it.

crates/exex/src/model.rs
tcoratger: Just used KethU256 for the data storage

crates/exex/src/model.rs
tcoratger: @ClementWalter For testing I've just removed the proptest for this in favor of a loop over 100 random in a classical test which is way faster and similar in terms of security

ClementWalter: initial bloom is 16 times bytes16 (not sure why this format and not just `[u8]`). I don't know if the best representation for keth is `[u8]` (ie `felt*` with one byte per felt) or `[u128]` (ie `Uint256*` with `low, high, low,high, etc.)

let me check where it's used in the EVM (not used in kakarot currently)

crates/exex/src/model.rs
ClementWalter: I don't see them neither, but I think indeed that it's worth implementing the `TryInto<uX> for Felt`

tcoratger: Yes will do that

crates/exex/src/model.rs
ClementWalter: ok looking at https://github.com/ethereum/execution-specs/blob/9b95554a88d2a8485f8180254d0f6a493a593fda/src/ethereum/cancun/fork.py#L689

it seems like a list of u128 is fine. note sure to understand you reversed order manipulation though, let's say we have  `bloom: felt*`, we'll have

```
bloom[0] = int.from_bytes(bytes([149, 59, 225, 248, 7, 245, 195, 147, 171, 156, 97, 130, 30, 73, 17, 104]))
...
bloom[15] = int.from_bytes(bytes([155, 64, 120, 34, 76, 21, 10, 203, 23, 114, 21, 101, 59, 7, 111, 68]))
```

crates/exex/src/model.rs
```diff
@@ -203,6 +215,48 @@ impl From<U256> for KethU256 {
...
+/// It efficiently stores 256-bit values and provides an interface for converting data structures
+/// into a compatible format.
+#[derive(Debug, Default, Eq, Ord, Hash, PartialEq, PartialOrd, Clone, Serialize, Deserialize)]
+pub struct KethData {
+    /// The length of the data to be stored.
+    len: KethMaybeRelocatable,
+
+    /// A vector holding the main data, represented as [`KethU256`] values.
+    data: Vec<KethU256>,
+}
```
ClementWalter, line 232:
```suggestion
pub struct KethPointer {
    /// The length of the data to be stored.
    len: KethMaybeRelocatable,

    /// A vector holding the main data, represented as [`KethU256`] values.
    data: Vec<KethMaybeRelocatable>,
    
    /// The size of the underlying Cairo struct (see `.SIZE`)
    type_size: usize
}
```

ClementWalter: suggestion to make this closer to cairo, the default being `type_size = 1`

crates/exex/src/model.rs
```diff
@@ -203,6 +215,48 @@ impl From<U256> for KethU256 {
...
+    /// The [`Bloom`] filter is represented as a 256-byte array in big-endian order. Since
+    /// CairoVM's [`Felt252`] can only handle values up to 252 bits, we need to break the
+    /// 256-byte array into smaller chunks that fit within this limit.
+    ///
+    /// The conversion process works as follows:
+    /// - The first field (`len`) holds the length of the original data, represented as a
+    ///   [`KethMaybeRelocatable`].
+    /// - The `data` field stores the remaining elements as chunks of 16 bytes each from the Bloom
+    ///   filter, with each chunk converted into a [`KethU256`]. The chunks are reversed to align
+    ///   with CairoVM's expected little-endian order.
```
ClementWalter, line 246:
```suggestion
    /// - The `data` field stores the remaining elements as chunks of 16 bytes each from the Bloom
    ///   filter, with each chunk converted into a [`KethMaybeRelocatable`].
```

ClementWalter: not sure why you say that the VM expects little endianess

crates/exex/src/model.rs
tcoratger: Done

ClementWalter (approved):

ClementWalter (approved):

PR #51, opened 2024-10-03
Fix model.BlockHeader fields

ClementWalter:
### Why

Following the implementation of Rust hints, we figure out that the initial model based on a sample of the ef tests data was not accurate.

### What

Make it compliant with https://github.com/ethereum/execution-specs/blob/9b95554a88d2a8485f8180254d0f6a493a593fda/src/ethereum/cancun/blocks.py#L38-L62


cairo/src/model.cairo
```diff
@@ -213,29 +213,28 @@ namespace model {
...
+        ommers_hash: Uint256,
         coinbase: felt,
-        difficulty: felt,
-        excess_blob_gas: felt,
-        extra_data_len: felt,
-        extra_data: felt*,
+        state_root: Uint256,
+        transactions_root: Uint256,
+        receipt_root: Uint256,
+        withdrawals_root: Uint256,
```
tcoratger, line 222: Should be an Option no? as per https://github.com/alloy-rs/alloy/blob/8132d691762f69d4385050335978654c5c37510a/crates/consensus/src/header.rs#L49

tcoratger (approved):

cairo/src/model.cairo
```diff
@@ -213,29 +213,28 @@ namespace model {
...
-        transactions_trie: Uint256,
-        uncle_hash: Uint256,
-        withdrawals_root: Uint256,
+        base_fee_per_gas: Option,
+        blob_gas_used: Option,
+        excess_blob_gas: Option,
+        parent_beacon_block_root: Option,
+        requests_root: Option,
+        extra_data_len: felt,
+        extra_data: felt*,
```
obatirou, line 237: No `prev_randao: Bytes32` ?

tcoratger: prev_randao is mixed hash here as per https://github.com/paradigmxyz/reth/blob/96f1406a14b4e9d9ef4ed22879accd34c89430ec/crates/rpc/rpc-eth-types/src/simulate.rs#L292

obatirou: EIP confirms that is it only a suggested renaming https://eips.ethereum.org/EIPS/eip-4399
It seems clients kept `mixHash` and did not do the renaming
Same in geth https://github.com/ethereum/go-ethereum/blob/65e5ca7d8126f7a8c708f8affb64f16c22cc63c0/core/types/block.go#L74

ClementWalter: I think that it's because the name "randao" is misleading as some people have believed that it was a safe random number

PR #60, opened 2024-10-08
model: impl `From<Transaction>` for `KethPointer`

tcoratger:


ClementWalter (approved):
I double checked that `.encode` is the signing data

```
    /// This encodes the transaction _without_ the signature, and is only suitable for creating a
    /// hash intended for signing.
    fn encode(&self, out: &mut dyn bytes::BufMut) {
```

PR #61, opened 2024-10-08
model: `TryFrom<TransactionSigned>` for `KethTransactionEncoded`

tcoratger:


crates/exex/src/model.rs
```diff
@@ -958,4 +1024,75 @@ mod tests {
             assert_eq!(keth_rlp.type_size, 1);
         }
     }
+
+    #[test]
+    fn test_signed_transaction_to_transaction_encoded() {
+        for _ in 0..50 {
```
ClementWalter, line 1030: it's generally not a good idea to do fuzzing like this because you loose the fact that you have one test per test case
- if it fails at some points, everything fails and test stops
- you can't reproduce the failing example

ClementWalter (dismissed):

crates/exex/src/model.rs
tcoratger: fixed

ClementWalter (approved):

PR #63, opened 2024-10-11
Use actualy dict for dict_squash and dict_copy from segment

ClementWalter:


cairo/src/state.cairo
```diff
@@ -417,6 +414,30 @@ namespace Internals {
...
+        accounts_start: DictAccess*, accounts_end: DictAccess*
+    ) {
+        if (accounts_start == accounts_end) {
+            return ();
+        }
+
+        if (accounts_start.new_value == 0) {
+            // If we do a dict_read on an unexisting account, `prev_value` and `new_value` are set to 0.
+            // However we expected pointers to model.Account, and casting 0 to model.Account* will
+            // cause a "Memory address must be relocatable" error.
```
obatirou, line 430: Comment needs fixing

cairo/src/utils/dict.cairo
```diff
@@ -1,134 +1,42 @@
...
-    static_assert len == [ap - 2];
-    static_assert dict == [ap - 1];
-
-    jmp loop if len != 0;
-
-    return (values_len, values);
+    let (local new_start: DictAccess*) = alloc();
+    let new_end = new_start + (dict_end - dict_start);
+    memcpy(new_start, dict_start, dict_end - dict_start);
+    %{ dict_copy %}
```
obatirou, line 16: Why this hint ?

cairo/src/utils/dict.cairo
```diff
@@ -1,134 +1,42 @@
...
-
-    static_assert squashed_start == [ap - 3];
-    static_assert dict_len == [ap - 2];
-    static_assert new_ptr == [ap - 1];
-
-    jmp loop if dict_len != 0;
 
-    return (new_start, new_ptr);
+    %{ dict_squash %}
+    ap += 1;
```
obatirou, line 27: From my understanding this si due to the hint doing  `memory[ap] = base`.
What happen if the wrong value is filled ?
There is no contraints on this memory cell unless I miss something

cairo/src/state.cairo
ClementWalter: ? what is unclear (same comment as in kakarot btw)

cairo/src/utils/dict.cairo
ClementWalter: the hint is for registering the dict into the underlying `DictManager`.

cairo/src/state.cairo
obatirou: mmh it seems I hallucinated an other comment. Sorry for that. It's clear

cairo/src/utils/dict.cairo
obatirou: I did miss something 
Thanks for the clarification

obatirou (approved):

PR #65, opened 2024-10-16
Implement and diff-test validate_header

ClementWalter:
### Why

In order to make sure that keth closely follows the ethereum spec, we need to make sure that the fork rules are correctly implemented.

One of the best way to do this is to work with diff testing.

### What

Refactor the codebase to be closer to the execution spec and diff test every function.

### How

Create a file structure that mimics the one in the execution specs and implement the required functions in order.

This PR starts with the `validate_header` function.

### DIFF
--- a/cairo/programs/fork.cairo
+++ b/cairo/programs/fork.cairo
@@ -8,6 +8,7 @@ from starkware.cairo.common.bool import FALSE
 from src.model import model
 
 using Uint128 = felt;
+using Uint64 = felt;
 
 const ELASTICITY_MULTIPLIER = 2;
 const GAS_LIMIT_ADJUSTMENT_FACTOR = 1024;
@@ -18,7 +19,7 @@ const EMPTY_OMMER_HASH_HIGH = 0x1dcc4de8dec75d7aab85b567b6ccd41a;
 
 // @notice See https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/fork.py#L1118-L1154
 // @dev We use the Uint128 alias to strenghten the fact that these felts should have been range_checked before
-func check_gas_limit{range_check_ptr}(gas_limit: Uint128, parent_gas_limit: Uint128) {
+func check_gas_limit{range_check_ptr}(gas_limit: Uint64, parent_gas_limit: Uint64) {
     let (max_adjustment_delta, _) = unsigned_div_rem(parent_gas_limit, GAS_LIMIT_ADJUSTMENT_FACTOR);
 
     with_attr error_message("InvalidBlock") {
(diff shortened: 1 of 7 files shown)


cairo/programs/fork.cairo
```diff
@@ -0,0 +1,203 @@
...
+    assert [range_check_ptr] = header.transactions_root.high;
+    let range_check_ptr = range_check_ptr + 1;
+    // receipt_root
+    assert [range_check_ptr] = header.receipt_root.low;
+    let range_check_ptr = range_check_ptr + 1;
+    assert [range_check_ptr] = header.receipt_root.high;
+    let range_check_ptr = range_check_ptr + 1;
+    // withdrawals_root
+    assert header.withdrawals_root.is_some * (1 - header.withdrawals_root.is_some) = 0;
+    let withdrawals_root = cast(header.withdrawals_root.value, Uint256*);
```
obatirou, line 114: Why a Uint256* ?
The withdrawals_root is an Option and Option is 
```
struct Option {
        is_some: felt,
        value: felt,
    }
```
https://github.com/kkrt-labs/keth/blob/e8e1da9291a2508d36d7f340d9df9339d8bba389/cairo/src/model.cairo#L222

obatirou: I might miss something as the same pattern is done for `parent_beacon_block_root` but not `base_fee_per_gas`

ClementWalter: `Option` is `Option<type>`, in this case it's a pointer to a `Uint256` (because hashes are uint256)

ClementWalter: you can check the types in https://github.com/alloy-rs/alloy/blob/main/crates/consensus/src/header.rs, some are `Option<u64>` so `felt` is enough, and some are `Option<B256>` so we need to have a `Uint256`, and so a `Uint256*`

cairo/programs/fork.cairo
```diff
@@ -0,0 +1,203 @@
...
+        let range_check_ptr = range_check_ptr + 1;
+        let expected_base_fee_per_gas = calculate_base_fee_per_gas(
+            header.gas_limit,
+            parent_header.gas_limit,
+            parent_header.gas_used,
+            parent_header.base_fee_per_gas.value,
+        );
+
+        assert expected_base_fee_per_gas = header.base_fee_per_gas.value;
+        assert [range_check_ptr] = header.timestamp - parent_header.timestamp - 1;
```
obatirou, line 203: how the parent_header will be filled ?

ClementWalter: it'll be filled with a hint as well. the currently commented hash computation will ensure that the whole parent header is consistent with the current block

PR #67, opened 2024-10-16
rm rust toolchain

tcoratger:
Should close https://github.com/kkrt-labs/keth/issues/66


.github/workflows/trunk_check.yml
```diff
@@ -27,6 +34,7 @@ jobs:
         with:
           python-version: 3.10.14
           cache: pip
+
```
tcoratger, line 37:
```suggestion
```

PR #68, opened 2024-10-18
KakarotSerde: implement `get_identifier`

tcoratger:


crates/exex/src/serde.rs
ClementWalter, line 1: split `test_program_identifier_erroneous_identifier` to have one test per case
we can use `rstest` for factorizing the shared init code

also please include the source code of valid_program as it's otherwise hard to follow

tcoratger: - The program comes from here https://github.com/lambdaclass/cairo-vm/blob/main/cairo_programs/manually_compiled/valid_program_b.json not sure there is a source code for this as it is manually compiled. But I think that this is a valid source of truth for our tests because this is also used to end to end test the cairo vm itself.
- For the tests I avoided doing this so as not to deserialize the program several times, knowing that each assertion will give us the necessary message if it fails and we will know which assertion panicked even if it is in the same test.

tcoratger: 1. Ok will do that but I think that the OS program will be constantly modified (and therefore the JSON will either be obsolete or will have to be constantly updated). For me, this JSON must be stored in the Rust part to really only serve as test data for the small Rust test utilities and not mix with the rest. In my opinion, we could simply keep this small program that is validated by the Rust Cairo-VM and that is quite small and succinct. But ok no problem I will replace in my testdata by the kakarot OS program. But just a quick question, let's imagine that in 2 weeks we completely change the OS program, how to do it?
     - Do we update the JSON in test data at the risk of blowing up all the Rust tests?
     - Or do we leave the old JSON at the risk of it being obsolete compared to the updated Kakarot OS code?

2. I don't like rtest too much because:
     - It adds a dependency and therefore compilation time for tests
     - The manipulation of elements is sometimes not intuitive, I don't like it too much. 
 But for the sake of it, I'm going to make a small setup function that I'll call at the beginning of my tests. For this simple case it's perfectly sufficient.

tcoratger: @ClementWalter Comment is fixed, should be ready to merge

crates/exex/src/serde.rs
```diff
@@ -0,0 +1,226 @@
...
+            kakarot_serde.get_identifier("__temp0", Some("reference".to_string())).unwrap(),
+            Identifier {
+                pc: None,
+                type_: Some("reference".to_string()),
+                value: None,
+                full_name: Some("starkware.cairo.common.memcpy.memcpy.__temp0".to_string()),
+                members: None,
+                cairo_type: Some("felt".to_string())
+            }
+        );
```
ClementWalter, line 149: I would do this test because it doesn't mean anything for a cairo dev, it's auto generated naming

ClementWalter (approved):

crates/exex/src/serde.rs
tcoratger: This is just to assert a complex case and to be sure all the fields are well captured, do you want me to remove?

PR #71, opened 2024-10-21
Set rust version in cargo.toml

ClementWalter:


Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [workspace.package]
 version = "0.1.0"
 edition = "2021"
+rust-version = "1.81"
```
tcoratger, line 4: you can do 1.82 as per https://github.com/paradigmxyz/reth/blob/51594c9a68e437075f0aa0bfc13491dd82176d7c/Cargo.toml#L4

tcoratger (approved):

PR #83, opened 2024-10-31
EELS scaffolding

ClementWalter:


cairo/tests/ethereum/utils/test_numeric.py
```diff
@@ -0,0 +1,20 @@
...
+class TestNumeric:
+    @given(a=uint128, b=uint128)
+    def test_min(self, cairo_run, a, b):
+        assert min(a, b) == cairo_run("test_min", a=a, b=b)
+
+    @given(
+        value=uint128,
+        div=st.integers(min_value=1, max_value=PRIME // (2**128 - 1) - 1),
+    )
+    def test_divmod(self, cairo_run, value, div):
```
enitrat, line 17: ik and have seen that the 0 case is handled correctly but should it be covered by a test

cairo/tests/test_serde.cairo
```diff
@@ -1,60 +1,28 @@
...
-        State.update_account(0x000f3df6d732807ef1319fb7b8bb8522d0beac02, account);
-
-        let account = State.get_account(0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b);
-        tempvar key = new Uint256(0, 0);
-        let (account, value) = Account.read_storage(account, key);
-        State.update_account(0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b, account);
-    }
-    return state;
-}
+from ethereum.base_types import (
```
enitrat, line 1: empty besides imports?

cairo/ethereum/cancun/blocks.cairo
```diff
@@ -0,0 +1,110 @@
...
+    timestamp: U256,
+    extra_data: Bytes,
+    prev_randao: Bytes32,
+    nonce: Bytes8,
+    base_fee_per_gas: Uint,
+    withdrawals_root: Root,
+    blob_gas_used: U64,
+    excess_blob_gas: U64,
+    parent_beacon_block_root: Root,
+}
```
tcoratger, line 56: A question on this; as now we pull all these types (like the Header here for example) to the Ethereum Execution Layer spec, how are we going to match "exactly" the reth types on the other side for the interaction with the ExEx? 

A few weeks ago, I remember that you made a Header Cairo struct that exactly matches the alloy type line by line and type by type, so that means here in Cairo that we will now have a Header Execution Spec struct and a Header Alloy struct or are we going to do the matching afterwards or via other means?

cairo/tests/test_serde.cairo
ClementWalter: actually the `test_serde.py` doesn't even use the cairo file because it does the roundtrip without the cairo run. I could actually update the tests to also use the roundtrip with the cairo run step

cairo/tests/ethereum/utils/test_numeric.py
ClementWalter: `divmod(x, 0)` raises in python with `ZeroDivisionError` and in cairo with `AssertionError: div=0x0 is out of the valid range.`

cairo/ethereum/cancun/blocks.cairo
ClementWalter: good question, so for now I've not updated the main flow, this is like a new line of research (as we spoke) but I wanted to clean the wip and push it to merge. The `os.cairo` still uses the previously defined `Headers` and related types.

That being said, I think that the previous approach was less maintainable as we defined new types in cairo and then new types in python/rust to match them, so doing a eels/alloy to python/rust equivalent of cairo conversion, then this python/rust type to cairo. I think that it's better to skip these new types and do the job to "use directly" the eels ones. If you implement the same `gen_arg` in rust as in python, because eels and alloy types are supposed to be the same, you should be good, and it requires less per final type logic

PR #86, opened 2024-11-03
serde: start `serialize_inner` implementation

tcoratger:


crates/exex/src/serde.rs
```diff
@@ -936,4 +996,106 @@ mod tests {
...
+
+    #[test]
+    fn test_serialize_inner_felt() {
+        // Setup the KakarotSerde instance
+        let mut kakarot_serde = setup_kakarot_serde();
+
+        // Setup
+        let output_ptr = Relocatable { segment_index: 10, offset: 11 };
+        let range_check_ptr = Felt252::ZERO;
+        let bitwise_ptr = Felt252::from(55);
```
ClementWalter, line 1008: naming is misleading as one expects to have pointers, not felt

ClementWalter (approved):

crates/exex/src/serde.rs
tcoratger: just fixed that simply with a and b so that this is really generic and not prone to misinterpretation.

ClementWalter (approved):

PR #89, opened 2024-11-05
serde: implement `serialize_option`

tcoratger:


crates/exex/src/serde.rs
```diff
@@ -344,6 +344,27 @@ impl KakarotSerde {
...
+        &self,
+        ptr: Relocatable,
+    ) -> Result<Option<MaybeRelocatable>, KakarotSerdeError> {
+        // Retrieve the serialized "Option" struct as a map of field names to values.
+        let raw = self.serialize_pointers("model.Option", ptr)?;
+
+        // Check if "is_some" field is present; if not, return None.
+        if !raw.contains_key("is_some") {
+            return Ok(None);
+        }
```
ClementWalter, line 362: how could this field not be present? it's supposed to be present but 0 or 1

crates/exex/src/serde.rs
```diff
@@ -1095,4 +1136,100 @@ mod tests {
...
         assert_eq!(result, Ok(SerializedScope::U256(x)));
     }
+
+    #[test]
+    fn test_serialize_option_some_value() {
+        // Setup KakarotSerde instance
+        let mut kakarot_serde = setup_kakarot_serde(TestProgram::ModelOption);
+
+        // Setup
+        let is_some_ptr = kakarot_serde.runner.vm.add_memory_segment();
```
ClementWalter, line 1146: `is_some` is not a pointer but a felt

crates/exex/src/serde.rs
```diff
@@ -1095,4 +1136,100 @@ mod tests {
...
+
+    #[test]
+    fn test_serialize_option_none_value() {
+        // Setup KakarotSerde instance
+        let kakarot_serde = setup_kakarot_serde(TestProgram::ModelOption);
+
+        // Serialize the Option struct without any memory segment.
+        let result = kakarot_serde
+            .serialize_option(Relocatable::default())
+            .expect("failed to serialize model.Option");
```
ClementWalter, line 1177: I don't understand this test, as there is no `model.Option` definition

crates/exex/src/serde.rs
```diff
@@ -1095,4 +1136,100 @@ mod tests {
...
+            .get_relocatable()
+            .unwrap();
+
+        // Serialize the Option struct with `is_some` set to `false`.
+        let result =
+            kakarot_serde.serialize_option(base).expect("failed to serialize model.Option");
+
+        // Assert that the result is None since `is_some` is `false`.
+        assert!(result.is_none());
+    }
```
ClementWalter, line 1234: the only test we want is this one, plus another with a `Felt252::ONE`.

crates/exex/src/serde.rs
tcoratger: It's to validate that in this case it simply return a None. No model.Option means a None no? I don't think throwing an error is good practice in this case because in Rust a common practice is to return None is these cases

crates/exex/src/serde.rs

crates/exex/src/serde.rs

ClementWalter (approved):

PR #92, opened 2024-11-06
serde: implement `serialize_dict`

tcoratger:


crates/exex/src/serde.rs
```diff
@@ -1254,4 +1345,209 @@ mod tests {
...
+        kakarot_serde
+            .runner
+            .vm
+            .insert_value((dict_ptr + 6usize).unwrap(), key3.clone())
+            .expect("failed to insert key2");
+        kakarot_serde
+            .runner
+            .vm
+            .insert_value((dict_ptr + 8usize).unwrap(), value3.clone())
+            .expect("failed to insert value2");
```
ClementWalter, line 1418: insert directly a `Vec`, no need to do this 6 times

crates/exex/src/serde.rs
```diff
@@ -1254,4 +1345,209 @@ mod tests {
...
+        kakarot_serde
+            .runner
+            .vm
+            .insert_value((dict_ptr + 0usize).unwrap(), key.clone())
+            .expect("failed to insert key");
+        kakarot_serde
+            .runner
+            .vm
+            .insert_value((dict_ptr + 2usize).unwrap(), null_value)
+            .expect("failed to insert null value");
```
ClementWalter, line 1451: see above

crates/exex/src/serde.rs
tcoratger: ah yes forgot this trick again, fixed

crates/exex/src/serde.rs

ClementWalter (approved):

PR #94, opened 2024-11-07
serde: implement `serialize_memory`

tcoratger:


crates/exex/src/serde.rs
```diff
@@ -1520,4 +1633,244 @@ mod tests {
...
+
+        // Key-value pairs to be inserted into the dictionary
+        let key1 = MaybeRelocatable::Int(Felt252::from(0));
+        let key2 = MaybeRelocatable::Int(Felt252::from(1));
+        let key3 = MaybeRelocatable::Int(Felt252::from(2));
+        let key4 = MaybeRelocatable::Int(Felt252::from(3));
+        let value1 = MaybeRelocatable::Int(Felt252::from(10));
+        let value2 = MaybeRelocatable::Int(Felt252::from(20));
+        let value3 = MaybeRelocatable::Int(Felt252::from(30));
+        let value4 = MaybeRelocatable::Int(Felt252::from(40));
```
ClementWalter, line 1682: could use holes in keys, and not in order as well, to make sure serialization gets it right

tcoratger: just added a special test for that

PR #97, opened 2024-11-12
Add rlp methods

ClementWalter:
See https://github.com/ethereum/execution-specs/blob/master/src/ethereum/rlp.py


cairo/ethereum/crypto/hash.cairo
```diff
@@ -1,3 +1,20 @@
...
+from starkware.cairo.common.alloc import alloc
 
 using Hash32 = Bytes32;
+
+func keccak256{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: KeccakBuiltin*}(
+    buffer: Bytes
+) -> Hash32 {
+    alloc_locals;
+    let (local dst: felt*) = alloc();
+    bytes_to_bytes8_little_endian(dst, buffer.value.len, buffer.value.data);
```
enitrat, line 14: q: in that case no need to return `dst_len` because the value we're passing in other functions is the `buffer.value.len` right?

cairo/ethereum/crypto/hash.cairo
```diff
@@ -1,3 +1,20 @@
...
+
+    let (code_hash) = keccak_bigend(dst, buffer.value.len);
+    tempvar value = new Uint256(low=code_hash.low, high=code_hash.high);
```
enitrat, line 17: can use `&` instead of a new Uint256 ?

enitrat (approved):

cairo/ethereum/crypto/hash.cairo
ClementWalter: not possible, even with `__fp__`. The only other option is to `get_ap()` then `cast(ap, Uint256*)`. 

`get_ap` does two function calls, while `new` only one (`alloc`), at the end i think that there is a significant different difference (given that the whole function consumes much more in any case)

PR #98, opened 2024-11-13
serde: implement `serialize_struct`

tcoratger:


crates/exex/src/serde.rs
```diff
@@ -2326,4 +2419,107 @@ mod tests {
...
+            .gen_arg(&vec![
+                member_value1.clone(),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                MaybeRelocatable::Int(Felt252::ZERO),
+                member_value2.clone(),
```
ClementWalter, line 2505: this is not a valid KeccackBuiltin struct, which is defined as

```
struct KeccakBuiltin {
    input: KeccakBuiltinState,
    output: KeccakBuiltinState,
}
```

with

```
struct KeccakBuiltinState {
    s0: felt,
    s1: felt,
    s2: felt,
    s3: felt,
    s4: felt,
    s5: felt,
    s6: felt,
    s7: felt,
}
```

meaning that `KeccakBuiltin` is 2 x 8 = 16 keys long

crates/exex/src/serde.rs
```diff
@@ -135,6 +140,15 @@ impl CairoType {
     ) -> Self {
         Self::Tuple { members, has_trailing_comma, location }
     }
+
+    /// Returns the name of the type, if it has one.
```
ClementWalter, line 144: wrong doc string

crates/exex/src/serde.rs
tcoratger: Yes you mean that my member_value1 and member_value2 should be Relocatable pointing to my two KeccakBuiltinState respectively right?

But here this is not supposed to be realistic, just wanted to test with the minimal setup to ensure that we fetch properly the values. So for this test and only for this test (and not in real life):

```rust
 struct KeccakBuiltin {
    input: felt,
    output: felt,
}
```

tcoratger: And even if I wanted to, right now in my `serialize_inner` I don't have the proper tools needed to serialize it completely (only have felt stuffs)

ClementWalter: no no, not relocatable, in cairo zero, this definition

```
struct KeccakBuiltin {
    input: KeccakBuiltinState,
    output: KeccakBuiltinState,
}
```

actually inlines the `KeccakBuiltinState`, and so lead to the exact same memory layout as (leaving the conflicting name for the sake of clarity)

```
struct KeccakBuiltin {
    s0: felt,
    s1: felt,
    s2: felt,
    s3: felt,
    s4: felt,
    s5: felt,
    s6: felt,
    s7: felt,
    s0: felt,
    s1: felt,
    s2: felt,
    s3: felt,
    s4: felt,
    s5: felt,
    s6: felt,
    s7: felt,
}
```

crates/exex/src/serde.rs
```diff
@@ -2326,4 +2393,113 @@ mod tests {
...
+        let kakarot_serde = setup_kakarot_serde(&TestProgram::KeccakAddUint256);
+
+        // Test with a struct name that doesn't exist in the program
+        let ptr = Some(Relocatable { segment_index: 0, offset: 0 });
+        let result = kakarot_serde.serialize_struct("NonExistentStruct", ptr);
+
+        // Assert that the error indicates the struct was not found
+        assert_eq!(
+            result,
+            Err(KakarotSerdeError::StructNotFound("NonExistentStruct".to_string(), 0))
```
ClementWalter, line 2453:
```suggestion
            Err(KakarotSerdeError::StructNotFound("NonExistingStruct".to_string(), 0))
```

crates/exex/src/serde.rs
```diff
@@ -2326,4 +2393,113 @@ mod tests {
...
+        fn create_tmp(value: u64) -> SerializedResult {
+            SerializedResult::Tmp(Some(MaybeRelocatable::Int(Felt252::from(value))))
+        }
+
+        // Helper function to create a `HashMap` for struct serialization
+        fn create_struct_data(start: u64) -> SerializedResult {
+            SerializedResult::Struct(
+                (0..8).map(|i| (format!("s{i}"), Some(create_tmp(start + i)))).collect(),
+            )
+        }
```
ClementWalter, line 2474: no helper function in test. a test function should remain as small and focused on the sut as possible

ClementWalter (approved):
we can merge and then plan a refacto of the module

PR #100, opened 2024-11-14
add `SharpSdk` scaffolding

tcoratger:


crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub trace_generation: Url,
+    /// Submit an Atlantic trace generation and proof generation query
+    pub trace_generation_proof_generation: Url,
+    /// Submit a L2 Atlantic query
+    pub atlantic_query: Url,
+    /// Submit a L2 Atlantic proof generation and proof verification query
+    pub proof_generation_verification: Url,
+    /// Submit a L2 Atlantic proof verification query
+    pub proof_verification: Url,
+}
```
ClementWalter, line 19: this should probably be an enum

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+/// The Sharp queries endpoints.
+#[derive(Debug, Clone)]
+pub struct SharpQueriesEndpoints {
+    /// Get the list of Atlantic queries submitted by the user
+    pub get_queries: Url,
+    /// Get the details of a specific Atlantic query
+    pub get_query: Url,
+    /// Get the list of jobs for a specific Atlantic query
+    pub get_query_jobs: Url,
+}
```
ClementWalter, line 45: also an enum here. And I think that the `get_` prefix is superfluous

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_l2_endpoints_invalid_base_url() {
+        // Test with an invalid base URL
+        let base_url = Url::parse("invalid-url");
+        assert!(base_url.is_err(), "Expected base URL parsing to fail");
+    }
```
ClementWalter, line 92: this is testing the `url` module and not our code, so useless (unless you don't trust `url`, but then do we want to use it?)

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    #[test]
+    fn test_l2_endpoints_valid_base_url_trailing_slash() {
+        let base_url = Url::parse("https://example.com/")
+            .expect("Failed to parse base URL with trailing slash");
```
ClementWalter, line 97: again here, external deps are supposed to work as expected, we can just `.?` or `unwrap`

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+            Url::parse("https://example.com/l2/atlantic-query/proof-verification").unwrap()
+        );
+    }
+
+    #[test]
+    fn test_sharp_queries_endpoints_invalid_base_url() {
+        // Test with an invalid base URL
+        let base_url = Url::parse("invalid-url");
+        assert!(base_url.is_err(), "Expected base URL parsing to fail");
+    }
```
ClementWalter, line 166: is this the exact same test as `test_l2_endpoints_invalid_base_url` ?

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,266 @@
...
+
+    #[test]
+    fn test_sharp_queries_endpoints_valid_base_url_trailing_slash() {
+        let base_url = Url::parse("https://example.com/")
+            .expect("Failed to parse base URL with trailing slash");
```
ClementWalter, line 171: see above

crates/exex/src/atlantic/endpoints.rs
ClementWalter, line 1: I think that all endpoints should be Enum. Also it's not clear from here how many enums are required (you defined 4 structs, but maybe they could be joined?)

crates/exex/src/atlantic/endpoints.rs
tcoratger: I would keep it as a structure because when we initialize the Sharp SDK, we initialize all the endpoints at once so that we can then call them as desired in different functions. 

If we made it an enum, when we do `SharpSdk::new`, we would be forced to specify the L2 endpoint which is not very practical. Here especially since it is only URLs so it is not very heavy.

crates/exex/src/atlantic/endpoints.rs
tcoratger: - Same comment for the enum
- Just fixed for the get prefixes

crates/exex/src/atlantic/model.rs
ClementWalter, line 1: use `#[serde(rename_all = "camelCase")]` for changing the case, instead of renaming each field

crates/exex/src/atlantic/endpoints.rs
tcoratger: here this is exactly the same as unwrap in rust just with a nice message in case of failure but the behaviour is exactly the same

crates/exex/src/atlantic/endpoints.rs
tcoratger: same, this is unwrap with a nice message

crates/exex/src/atlantic/model.rs
tcoratger: nice point, fixed

crates/exex/src/atlantic/endpoints.rs
ClementWalter: but it creates more words to read and gives the impression that we expect this to possibly fail, which we don't want to bother at this place

tcoratger: fixed

crates/exex/src/atlantic/model.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        let serialized = serde_json::to_string(&context).expect("Failed to serialize Context");
+        assert_eq!(
+            serialized, r#"{"proofPath":"path/to/proof"}"#,
+            "Serialization of Context is incorrect"
+        );
+
+        let deserialized: Context =
+            serde_json::from_str(&serialized).expect("Failed to deserialize Context");
+        assert_eq!(deserialized, context);
```
ClementWalter, line 59: this part of the test is a bit odd because you test that `serde` is consistent with itself, precisely you do `serde_json::from_str(serde_json::to_string(content)) == content`

crates/exex/src/atlantic/model.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        let serialized = serde_json::to_string(&context).expect("Failed to serialize Context");
+        assert_eq!(
+            serialized, r#"{"proofPath":null}"#,
+            "Serialization of Context with None is incorrect"
+        );
+
+        let deserialized: Context =
+            serde_json::from_str(&serialized).expect("Failed to deserialize Context");
+        assert_eq!(deserialized, context);
```
ClementWalter, line 74: same

crates/exex/src/atlantic/model.rs
```diff
@@ -0,0 +1,160 @@
...
+        let serialized = serde_json::to_string(&job).expect("Failed to serialize Job");
+        assert_eq!(
+            serialized,
+            r#"{"id":"job123","status":"completed","context":{"proofPath":"path/to/proof"}}"#,
+            "Serialization of Job is incorrect"
+        );
+
+        let deserialized: Job =
+            serde_json::from_str(&serialized).expect("Failed to deserialize Job");
+        assert_eq!(deserialized, job);
```
ClementWalter, line 94: same

crates/exex/src/atlantic/model.rs
ClementWalter, line 1: you don't need to test the deserialization as it's testing serde_json, not our code

crates/exex/src/atlantic/endpoints.rs
```diff
@@ -0,0 +1,228 @@
...
+    /// Submit an Atlantic trace generation query
+    pub trace_generation: Url,
+    /// Submit an Atlantic trace generation and proof generation query
+    pub trace_generation_proof_generation: Url,
+    /// Submit a L2 Atlantic query
+    pub atlantic_query: Url,
+    /// Submit a L2 Atlantic proof generation and proof verification query
+    pub proof_generation_verification: Url,
+    /// Submit a L2 Atlantic proof verification query
+    pub proof_verification: Url,
```
ClementWalter, line 18: not all these are "l2" as per the doc

crates/exex/src/atlantic/model.rs
tcoratger: Here is my take:
- When I run the first Sharp job, I would like to fill with a round of real data here to really make sure that the serde, renames, variable names, format... match what is expected, hence keeping these tests and adding real data to them when available.

Please confirm that you still want to remove them?

crates/exex/src/atlantic/sharp.rs
```diff
@@ -0,0 +1,66 @@
...
+
+        // Create a new HTTP client
+        let client = Client::new();
+
+        // Construct the full URL including the API key
+        let url = format!("{}?apiKey={}", self.l2.proof_generation, self.api_key);
+
+        // Make a POST request to the proof generation endpoint
+        let response =
+            client.post(&url).multipart(form).send().await?.json::<QueryResponse>().await?;
```
ClementWalter, line 62: don't use `format` but add `.query(&[("apiKey", &self.api_key)])`

crates/exex/src/atlantic/sharp.rs
```diff
@@ -0,0 +1,66 @@
...
+                    .mime_str("application/zip")?,
+            )
+            .text("layout", layout.to_string())
+            .text("prover", prover.to_string());
+
+        // Create a new HTTP client
+        let client = Client::new();
+
+        // Construct the full URL including the API key
+        let url = format!("{}?apiKey={}", self.l2.proof_generation, self.api_key);
```
ClementWalter, line 58: note that `proof_generation` is not part of the `l2` namespace.

crates/exex/src/atlantic/endpoints.rs
tcoratger: I've seen the same, but as per

https://github.com/chudkowsky/herodotus_sharp_playground/blob/ed6a31b34f854fefacee4312fba6dff4d57bfe6d/src/models.rs#L102-L110

everything is in L2Endpoints. Since this guy has already run jobs on it, I guess he knows what it's about. Also, the documentation is currently not complete on herodotus so in reality we don't really know precisely what each endpoint does.

I also had to rename some endpoints that didn't seem to me to be in accordance with the docs compared to the client. Since we're not sure, I'll leave it here for now as per the example client we have and we can change it in the future.

crates/exex/src/atlantic/sharp.rs
tcoratger: fixed

crates/exex/src/atlantic/endpoints.rs
tcoratger: I see your approach but adding a new strum deps or doing some macro for that is really overkilled for now in my opinion. Here we have something clear and readable so that I know that in my Sharp sdk I've all my urls and I can query them as I want. We have less than 10 urls in total, this should be fine to have them like that. The debate would be different with 100 I think.

crates/exex/src/atlantic/endpoints.rs
ClementWalter: if some things are unclear, we should ask, but **never** copy code just because we don't know

crates/exex/src/atlantic/model.rs
ClementWalter: then we may add `deserialization` tests, but as it is today, this deserialization of a serialized data is useless

crates/exex/src/atlantic/endpoints.rs
ClementWalter: it's just that a struct is not meant to store a set of possible values

tcoratger: Final conclusion? you want me to use https://github.com/kkrt-labs/keth/pull/100#pullrequestreview-2438192033?

ClementWalter: I'm open to other suggestions, including not defining an enum and inlining directly the `str` in each function

ClementWalter: We could also leverage the trait names to do something like

```
use function_name::named;

impl SharpSdk {
    #[named]
    pub async fn proof_generation(
        &self,
        pie_file: Vec<u8>,
        layout: &str,
        prover: ProverVersion,
    ) -> Result<QueryResponse, SharpSdkError> {
        let endpoint = function_name!().to_string();  // Will return "proof_generation"
        // ... rest of the code
    }
}
```

crates/exex/src/atlantic/model.rs
ClementWalter: cursor ai my best friend

crates/exex/src/atlantic/endpoints.rs
tcoratger: fixed

ClementWalter (approved):

PR #102, opened 2024-11-15
Add rlp.decode methods and enum type

ClementWalter:


cairo/tests/utils/serde.py
```diff
@@ -61,10 +62,16 @@ def get_struct_definition(program, struct_name):
...
 class Serde:
-    def __init__(self, segments, program):
+    def __init__(self, segments, program, cairo_file):
         self.segments = segments
         self.memory = segments.memory
         self.program = program
+        self.cairo_file = cairo_file
+
+    @property
+    def main_part(self):
```
enitrat, line 72: ik this is only to go fast etc but it's a bit hard to follow the purpose of these functions without a one-liner comment

cairo/tests/utils/serde.py
```diff
@@ -371,6 +378,29 @@ def serialize_tuple(self, ptr, item_scope):
...
+        sequence_struct_ptr = self.serialize_pointers(f"Sequence{item_scope}", ptr)[
+            "value"
+        ]
+        raw = self.serialize_pointers(
+            f"Sequence{item_scope}Struct", sequence_struct_ptr
+        )
+        return list(self.serialize_list(raw["value"], item_scope, list_len=raw["len"]))
+
+    def serialize_enum(self, ptr, item_scope):
+        enum_struct_ptr = self.serialize_pointers(f"Enum{item_scope}", ptr)["value"]
```
enitrat, line 391: so this is now the generic way of modelling an Enum in cairo ?

enitrat (approved):

cairo/tests/utils/serde.py
ClementWalter: the "generic" way I've come up with, which comes with this kind of definition

```
struct EnumMyEnum {
    value: EnumMyEnumStruct*,
}

struct EnumMyEnumStruct {
    sequence: Sequence,
    bytes: Bytes,
    // other options
}
```

PR #107, opened 2024-11-18
Gas and transactions (part.)

ClementWalter:


cairo/ethereum/cancun/blocks.cairo
ClementWalter, line 1: `block` was inaccurate from previous attempts, just removed it first, will re-introduce when transactions are fixed

cairo/ethereum/cancun/fork.cairo
ClementWalter, line 1: access list deserialization is still wip, as the current `gen_arg` is not able to properly handle composite types not based on dataclasses (like `Tuple[Tuple[Address, Tuple[Bytes32, ...]], ...]` = access list)

Issue #110, opened 2024-11-19
rust model: transform `KethBlockHeader` into a `KethPayload`


tcoratger:
At the moment, we are able to transform a reth `Header` to a `KethBlockHeader` via

https://github.com/kkrt-labs/keth/blob/bd14d973f19314bb5cb246aaff6ceab3e4d8f65f/crates/exex/src/model.rs#L474-L501

But to be able to generate values ​​that will be injected into the segments of the Cairo VM for the `block` hint, we must transform this `KethBlockHeader` into a vector of `MaybeRelocatable` (format that the Cairo VM understands) and this in the right order, that is to say that we must respect the order provided by the Cairo structure `model.BlockHeader` (transactions will be handled separately in a follow up)

https://github.com/kkrt-labs/keth/blob/bd14d973f19314bb5cb246aaff6ceab3e4d8f65f/cairo/src/model.cairo#L214-L238. To do this, it is proposed, in Rust, to perform the following steps:

**- Create a `KethPayload` structure that will be used generally to create any Cairo VM payload via**

```rust
/// [`KethPayload`] is a wrapper around a [`Vec<MaybeRelocatable>`], designed to represent
/// a collection of `[MaybeRelocatable`] values in the Keth environment.
///
/// This struct is primarily used to encapsulate data that can injected into the Cairo VM for hint
/// manipulation.
#[derive(Debug, Eq, Ord, Hash, PartialEq, PartialOrd, Clone, Serialize, Deserialize)]
pub struct KethPayload(pub Vec<MaybeRelocatable>);

impl Deref for KethPayload {
    type Target = Vec<MaybeRelocatable>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

**- Create a `From` implementation that transform a `KethBlockHeader` to a `KethPayload` via**

```rust
impl From<KethBlockHeader> for KethPayload {
    /// Converts the [`KethBlockHeader`] into a [`KethPayload`] for use in the Cairo VM.
    ///
    /// This method serializes the [`KethBlockHeader`] into a format suitable for
    /// use in the Keth execution environment. The resulting [`KethPayload`] represents
    /// the block header data, making it compatible with the Cairo VM memory model.
    fn from(header: KethBlockHeader) -> Self {
        // Initialize the payload with fixed fields
        let mut payload = vec![
            header.parent_hash.low.0.clone(),
            header.parent_hash.high.0.clone(),
            header.ommers_hash.low.0.clone(),
[...]

PR #112, opened 2024-11-20
model: implement `From<KethBlockHeader>` for `KethPayload`

tcoratger:
Should close https://github.com/kkrt-labs/keth/issues/110

### DIFF
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -1170,6 +1170,7 @@ mod tests {
     }
 
     #[test]
+    #[allow(clippy::too_many_lines, clippy::cognitive_complexity)]
     fn test_keth_block_header_to_payload() {
         // Create realistic non-default values for the fields
         let parent_hash = B256::from([0x11u8; 32]);
--- a/crates/exex/src/model.rs
+++ b/crates/exex/src/model.rs
@@ -21,8 +21,8 @@ pub enum ConversionError {
 
 /// A custom trait for encoding types into a vector of [`MaybeRelocatable`] values.
 pub trait KethEncodable {
-    /// Encodes the type into a vector of [`MaybeRelocatable`] values.
-    fn encode(&self) -> Vec<MaybeRelocatable>;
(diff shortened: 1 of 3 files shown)


crates/exex/src/model.rs
```diff
@@ -266,6 +297,14 @@ impl Default for KethPointer {
     }
 }
 
+impl KethEncodable for KethPointer {
+    fn encode(&self) -> Vec<MaybeRelocatable> {
+        let mut encoded = vec![self.len.0.clone()];
+        encoded.extend(self.data.iter().map(|item| item.0.clone()));
+        encoded
+    }
```
ClementWalter, line 305: not sure about the underlying type you target, I understand that this creates a single segment with `[len, ...data]` while it should probably be just `[len, data*]`

crates/exex/src/model.rs
```diff
@@ -500,6 +539,59 @@ impl From<Header> for KethBlockHeader {
...
+}
+
+impl From<KethBlockHeader> for KethPayload {
+    fn from(header: KethBlockHeader) -> Self {
+        let mut payload = Vec::new();
+
+        // Dynamically encode each field using a trait
+        let fields = [
+            &header.parent_hash as &dyn KethEncodable,
```
ClementWalter, line 564: why do we need `dyn` actually and cannot know at compile time that all fields types implement `KethEncodable`?

crates/exex/src/model.rs
tcoratger: This is because in my keth pointer I've:
- len: The length of the data to be stored
- data: a vector of MaybeRelocatable holding the main data

So here I simply group them all into a single vector of MaybeRelocatable to have something that I'll inject into gen_arg at the end

crates/exex/src/model.rs
tcoratger: Unfortunately this is impossible.

While all these types implement the `KethEncodable` trait, their underlying types are different, and Rust needs to know the exact type of each element at compile time when creating the array.

ClementWalter (approved):

PR #120, opened 2024-11-22
Add oracle hint

ClementWalter:


cairo/ethereum/cancun/trie.cairo
ClementWalter, line 1: the commented boilerplate code is codegen

cairo/tests/ethereum/cancun/test_trie.py
ClementWalter, line 1: the file is generate with the scripts. Just uncommented the passing one. It's not enough to add `@pytest.mark.skip` now because the codegen script is not yet perfect so some types are broken

cairo/ethereum/cancun/trie.cairo
```diff
@@ -0,0 +1,314 @@
...
+//             // compact.append(16 * x[i] + x[i + 1])
+//     // else:
+//         // compact.append(16 * (2 * is_leaf + 1) + x[0])
+//         // for i in range(1, len(x), 2):
+//         // compact.append(16 * x[i] + x[i + 1])
+//             // compact.append(16 * x[i] + x[i + 1])
+//     // return Bytes(compact)
+// }
+
+func bytes_to_nibble_list{bitwise_ptr: BitwiseBuiltin*}(bytes_: Bytes) -> Bytes {
```
enitrat, line 171: 1 bitwise = 4 range check, price wise. so i think it's equivalent (nothing to do, just noting)

enitrat (approved):
💯

PR #124, opened 2024-11-25
Run txs in os

ClementWalter:


cairo/src/gas.cairo
```diff
@@ -177,7 +177,7 @@ namespace Gas {
         return expansion;
 
         no_expansion:
-        let range_check_ptr = [fp - 8];
+        let range_check_ptr = [fp - 9];
```
obatirou, line 180: Why ?
There is no local variable assignation before this line, there are 5 arguments (words_len, offset_1, offset_2, size_1, size_2) and one implicit that we want to retrieve range_check_ptr
By taking into account that a call instruction push next pc and current fp that are at fp-1 and fp-2, from my understanding fp - 8 is ok
What do I miss here ?

solidity_contracts/README.md
```diff
@@ -0,0 +1,66 @@
+## Foundry
```
obatirou, line 1: no need for the generic readme.
It needs to be removed or updated

solidity_contracts/src/ERC721.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.8.0;
+
+import {ERC721} from "solmate/tokens/ERC721.sol";
+
+contract KethToken is ERC721 {
```
obatirou, line 6: Not used ?

cairo/src/account.cairo
```diff
@@ -151,23 +150,15 @@ namespace Account {
...
-        let (local storage_addr) = Internals._storage_addr(key);
-
-        let (pointer) = dict_read{dict_ptr=storage}(key=storage_addr);
-        local value_ptr: Uint256*;
-        if (pointer != 0) {
-            assert value_ptr = cast(pointer, Uint256*);
-        } else {
-            assert value_ptr = new Uint256(0, 0);
-        }
+        tempvar value = Uint256(0, 0);
```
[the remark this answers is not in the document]
ClementWalter: ofc it's incorrect, just skipping now because it requires to update the way we fetch previous storage

solidity_contracts/src/ERC721.sol
ClementWalter: indeed, I wanted to add it in the test block but forget after all trials and error for ERC20

obatirou (dismissed):

PR #125, opened 2024-11-25
submit sharp job

### LINKED ISSUE #17: feat: send Cairo program traces to provers
### Why

Upon execution of the exex, the trace needs to be proven.

### What

- [ ] create a dummy `programs/os_mock.cairo` that does nothing (beware of builtins)
- [ ] investigate the following proving options (see below), create dedicated issues
- [ ] where is the proof verify ?

Possible provers are

- SHARP
- Gevulot
- local Stone

Possible verifiers

- SHARP => eth mainnet ?
- hylé
- aligned
- gevulot ?

tcoratger:
Resolves #17


crates/exex/src/exex.rs
```diff
@@ -86,11 +95,31 @@ impl<Node: FullNodeComponents> KakarotRollup<Node> {
...
+
+                // Path to a temporary file for the CairoPie.
+                let temp_file_path = Path::new("temp_pie.zip");
+
+                // Write the CairoPie data into a temporary file.
+                cairo_pie.write_zip_file(temp_file_path)?;
+
+                // Read the temporary file into a Vec<u8>.
+                let mut file = fs::File::open(temp_file_path)?;
+                let mut pie_bytes = Vec::new();
```
ClementWalter, line 113: can't we get the pie bytes from the `cairo_pie` to avoid this I/O op?

tcoratger: I've looked at the Cairo vm implementation and unfortunately I didn't see any function of this kind here https://github.com/lambdaclass/cairo-vm/blob/22be18094244c45e352dd6ca624a8d32f4cf8857/vm/src/vm/runners/cairo_pie.rs#L227

I wanted to investigate that precisely in the future with a full working example from Herodotus. There is probably a trick to encode the cairo pie in a friendly format that is well suited for Sharp.

I've seen also some manipulations of this kind in snos https://github.com/Moonsong-Labs/snos/blob/6376bb61f08d25dfbe68d670be542d5d3dcee029/src/sharp/pie.rs

But I would say that the best option is to ask directly to Herodotus a full working example (from a toy program to a proof generation via Atlantic) to go straight to the point.

ClementWalter (approved):

PR #142, opened 2024-11-28
chore: add cairo-lang dependency

obatirou:


obatirou:
Not necessary with uv workspace

PR #172, opened 2024-11-29
implement full block hint

tcoratger:
Should close https://github.com/kkrt-labs/keth/issues/138


ClementWalter (approved):

crates/exex/src/hints.rs
```diff
@@ -125,18 +135,183 @@ pub fn add_segment_hint() -> Hint {
...
+         _exec_scopes: &mut ExecutionScopes,
+         _ids_data: &HashMap<String, HintReference>,
+         _ap_tracking: &ApTracking,
+         _constants: &HashMap<String, Felt252>|
+         -> HintExecutionResult {
+            // Placeholder logic for account
+            Ok(())
+        },
+    )
+}
```
ClementWalter, line 194: remove, see https://github.com/kkrt-labs/keth/pull/181/files (my bad)
```suggestion
```

crates/exex/src/hints.rs
```diff
@@ -125,18 +135,183 @@ pub fn add_segment_hint() -> Hint {
...
+            layout: LayoutName::all_cairo,
+            trace_enabled: true,
+            relocate_mem: true,
+            proof_mode: true,
+            entrypoint: "test_block_hint",
+            ..Default::default()
+        };
+
+        // Prepare a random byte array for testing
+        let raw_bytes = [0u8; 1500];
```
ClementWalter, line 303: side question: where does the size (1500) come from?

tcoratger: The `arbitrary` library uses a random data stream from `raw_bytes`. The structure of `SealedBlock` influences the minimum amount of data needed in `raw_bytes`. If the array is too small, arbitrary might run out of data to generate a valid object.

So the 1500 provides this size and allows to have a full block with enough data and multiple transactions. For example, it is very likely that if the data sample was not enough and that few or no transactions are generated and also often the last data in the block are default data because the unstructured data source is not big enough.

In my experience on reth, for a block, between 1000 and 1500 is good to do complete randomness. So I chose 1500 here.

Issue #197, opened 2024-12-03
feat: remove transfers from state


ClementWalter:
## Why

In Kakarot, we needed to maintain a list of transfers to be performed at the end of the transaction because the balances were actually stored in a native Starknet ERC20. This is not the case in keth, and consequently these struct are useless.

## What

Remove the `transfers_len` and `transfers` from the `model.State` and everything related to it.

## How

Search for `state.transfers` everywhere and remove.

Issue #195, opened 2024-12-02
dev: remove compilation artifact


obatirou:
https://github.com/kkrt-labs/keth/tree/main/cairo/programs

Those compilation artifact seems to up to date. They need to be removed
- remove the compiled fibonacci json
- add a steps in the CI to generate it before the rust tests

PR #200, opened 2024-12-03
feat: remove transfers from state

tcoratger:
Should close https://github.com/kkrt-labs/keth/issues/197


cairo/programs/os.json
ClementWalter, line 1: let's stop committing this file, see also #195

tcoratger: makes sense, just deleted it

cairo/src/state.cairo
```diff
@@ -305,13 +278,11 @@ namespace State {
...
             events_len=state.events_len + 1,
             events=state.events,
-            transfers_len=state.transfers_len,
-            transfers=state.transfers,
         );
         return ();
     }
 
-    // @notice Add a transfer to the Transfer* array
+    // @notice Processes a transfer (we don't need to include the it in the state)
```
ClementWalter, line 285:
```suggestion
    // @notice Process a transfer
    // @dev Check and update balances
```

the comment `(we don't need to include ...` is really based on a previous non relevant context (being on sn)

ClementWalter (approved):

PR #204, opened 2024-12-03
fix: returndata of create txs

### LINKED ISSUE #143: fix: returndata of create txs
https://github.com/kkrt-labs/kakarot/pull/1557

obatirou:
Closes #143

Return the contract code for create txs

Note to reviewer: this is a fix imported from C4 mitigation, ensure the fix was correctly ported by looking at the corresponding issue and PR.

### DIFF
--- a/cairo/tests/programs/test_os.py
+++ b/cairo/tests/programs/test_os.py
@@ -150,3 +150,40 @@ class TestOs:
                 block=block_to_pass,
                 state=State.model_validate(initial_state),
             )
+
+    def test_create_tx_returndata(self, cairo_run):
+        initial_state = {
+            OWNER: {
+                "code": [],
+                "storage": {},
+                "balance": int(1e18),
+                "nonce": 0,
+            },
+            COINBASE: {
+                "code": [],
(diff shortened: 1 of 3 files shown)


cairo/tests/programs/test_os.py
```diff
@@ -121,3 +121,40 @@ def test_block_hint(self, cairo_run):
...
+                "nonce": 0,
+            },
+        }
+        transaction = {
+            "to": "",
+            "data": "0x604580600e600039806000f350fe7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3",
+            "value": 0,
+            "signer": OWNER,
+        }
+        state = cairo_run(
```
obatirou, line 152: This test only ensures the state is correct but do not verify the return data.
As the execution data is not needed in the context of keth, it seems difficult to test. Modifying the os to get it does not make sense either: adds code and complexity that will not be used

ClementWalter (dismissed):

PR #210, opened 2024-12-04
Trie lib (part.)

ClementWalter:


cairo/tests/fixtures/compiler.py
```diff
@@ -40,20 +41,25 @@ def cairo_file(request):
...
     return parts[1:] if parts[0] == "cairo" else parts
+
+
+@pytest.fixture(scope="module")
+def cairo_program(cairo_file, main_path):
+    start = perf_counter()
+    program = cairo_compile(cairo_file)
+    program.hints = implement_hints(program)
+    all_identifiers = list(program.identifiers.dict.items())
+    for k, v in all_identifiers:
```
ClementWalter, line 59: when running the tests, the main file is the test file and the compiler is not able to find struct defined therein in a breakpoint, looking for, e.g. `ethereum.cancun.trie.InternalNode` while only `__main__.InternalNode` exists. There is probably a better way to solve this at the IdentifierManager level, but it's not part of this repo and I'm not sure how fast can PR be merged on cairo-lang. Consequently, I've just decided to add them "manually"

cairo/ethereum/cancun/trie.cairo
```diff
@@ -1,83 +1,156 @@
...
+    tempvar sequence = SequenceExtended(new SequenceExtendedStruct(sequence_data, 2));
+    let unencoded_ = ExtendedImpl.sequence(sequence);
+    assert unencoded = unencoded_;
+    jmp common;
+
+    branch_node:
+    assert [
+        node.value.branch_node.value.subnodes.value.value +
+        node.value.branch_node.value.subnodes.value.len
+    ] = node.value.branch_node.value.value;
```
enitrat, line 129: c'est pas un probleme de reutiliser le meme segment de memoire ? `node` ne sera plus jamais utilisée ?

cairo/ethereum/cancun/trie.cairo
```diff
@@ -134,45 +207,87 @@ from ethereum.cancun.transactions import LegacyTransaction
...
+        assert [compact] = 16 * (2 * is_leaf.value);
+        tempvar result = Bytes(new BytesStruct(compact, 1));
+        return result;
+    }
+
+    local remainder = nondet %{ ids.x.value.len % 2 %};
+    assert remainder * (1 - remainder) = 0;
+    assert [compact] = 16 * (2 * is_leaf.value + remainder) + x.value.data[0] * remainder;
+
+    if (x.value.len == remainder) {
```
enitrat, line 258:
```suggestion
    if (x.value.len == 1) {
```

I would hardcode 1 here because the zero case has been handled explicitly before

cairo/ethereum/cancun/trie.cairo
```diff
@@ -134,45 +207,87 @@ from ethereum.cancun.transactions import LegacyTransaction
...
+    let (local compact) = alloc();
+
+    if (x.value.len == 0) {
+        assert [compact] = 16 * (2 * is_leaf.value);
+        tempvar result = Bytes(new BytesStruct(compact, 1));
+        return result;
+    }
+
+    local remainder = nondet %{ ids.x.value.len % 2 %};
+    assert remainder * (1 - remainder) = 0;
```
enitrat, line 255: what constrains remainder to be the correct value here ?

couldn't a malicious prover invert the `remainder` value returned ?

enitrat: perhaps we should do something like:
```
local remainder
local quotient
%{memory[fp], memory[fp+1] = oracle...()%}
assert quotient * 2 + remainder == ids.x.value.len ?
```

ClementWalter: it doesn't change anything, the below assertion is valid for both 0 and 1

See the python implementation

```
    if len(x) % 2 == 0:  # ie even length
        compact.append(16 * (2 * is_leaf))
    else:
        compact.append(16 * ((2 * is_leaf) + 1) + x[0])
```

cairo/ethereum/cancun/trie.cairo
ClementWalter: avoir avoir bien regardé l'implémentation, je pense que c'est ok mais c'est vrai que je ne suis pas 100% sûr donc peut-être qu'un memcpy ne ferait pas de mal

enitrat: add as todo comment maybe

cairo/tests/fixtures/compiler.py
enitrat: Add as comment for postériorité

enitrat (dismissed):

cairo/ethereum/cancun/trie.cairo
enitrat:
```
E           starkware.cairo.lang.vm.vm_exceptions.VmException: :272:45: Error at pc=0:1777:
E           Unknown value for memory cell at address 7:2.
E               assert [compact] = 16 * x_ptr.data[i] + x_ptr.data[i + 1];
E                                                       ^***************^
E           Cairo traceback (most recent call last):
E           :115:19: (pc=0:1569)
E               let compact = nibble_list_to_compact(node.value.extension_node.value.key_segment, bool(0));
```


```
    @given(node=...)
    @settings(deadline=None, max_examples=5)
    def test_encode_internal_node_should_fail(
        self, cairo_program, cairo_run, node: Optional[InternalNode]
    ):
        with (
            patch_hint(
                cairo_program,
                'ids.remainder = ids.x.value.len % 2',
                "ids.remainder = 1",
            )
        ):
            assert sequence_equal(
                encode_internal_node(node), cairo_run("encode_internal_node", node)
            )
 ```
 
 Can't you fill the memory with whatever you want here ?

Issue #212, opened 2024-12-04
fix: add assertion on remainder value in `nibble_list_to_compact`


enitrat:
see comment: https://github.com/kkrt-labs/keth/pull/210#discussion_r1869445417

```
    local remainder = nondet %{ ids.x.value.len % 2 %};
    assert remainder * (1 - remainder) = 0;
```

When we rely on a prover hint to compute `remainder = x_len % 2` we expose ourselves to a malicious prover that could fill any value, e.g. `remainder = 1`.

Currently, we assert that $remainder \in [0, 1[$.

However, if the malicious prover decides to replace the value with a wrong one, we end up underflowing to `P-1` on the operation
```
tempvar cond = x_ptr.len - i - 2;
```

which makes us loop forever, and would fail at generating a proof. 

Task:
Add an assertion to constrain the provided value. One idea is:
```
local remainder
local quotient
%{memory[fp], memory[fp+1] = div_rem(x, 2)%}
assert quotient * 2 + remainder == ids.x.value.len
```

PR #213, opened 2024-12-04
fix: add constraint on nibbles

enitrat:
Closes https://github.com/kkrt-labs/keth/issues/212
Closes https://github.com/kkrt-labs/keth/issues/214

- Adds a constraint that `(x.value.len - remainder) / 2;` is in range [0, range_check[; if it doesn't, the prover has provided a wrong value.

- Adds support to patch `nondet` hints.

### DIFF
--- a/cairo/ethereum/cancun/trie.cairo
+++ b/cairo/ethereum/cancun/trie.cairo
@@ -84,8 +84,8 @@ func encode_internal_node{
     range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: KeccakBuiltin*
 }(node: InternalNode) -> Extended {
     alloc_locals;
-
     local unencoded: Extended;
+    local range_check_ptr_end;
 
     tempvar is_leaf = cast(node.value.leaf_node.value, felt);
     jmp leaf_node if is_leaf != 0;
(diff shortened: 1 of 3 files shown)


cairo/ethereum/cancun/trie.cairo
```diff
@@ -238,18 +242,26 @@ func common_prefix_length(a: Bytes, b: Bytes) -> felt {
...
 
     if (x.value.len == 0) {
         assert [compact] = 16 * (2 * is_leaf.value);
         tempvar result = Bytes(new BytesStruct(compact, 1));
         return result;
     }
 
-    local remainder = nondet %{ ids.x.value.len % 2 %};
-    assert remainder * (1 - remainder) = 0;
+    // TODO: add a way to patch `nondet %{ ids.x.value.len % 2 %};` in test runner
```
ClementWalter, line 256: looking at the compiled program, you'd see a hint like

```
'memory[fp + 1] = to_felt_or_relocatable(ids.x.value.len % 2)'
```

so you don't need to rewrite the hint, just use this string in the test

enitrat: done with https://github.com/kkrt-labs/keth/issues/214

ClementWalter (approved):

PR #216, opened 2024-12-06
ci: add nightly fuzzing job

### LINKED ISSUE #169: test: run a nightly CI with increased fuzzing
## Why

More and more tests purely rely on fuzzing and diff testing against the reference EELS implementation. While we don't want to slow down the dev cycle, we want to make sure that we don't let cracks in the code.

## What

Add a nightly job like similar to https://github.com/kkrt-labs/kakarot/blob/main/.github/workflows/cairo-zero-nightly-fuzzing.yml

enitrat:
Closes #169

### DIFF
--- a/.github/workflows/nighly_fuzzing.yml
+++ b/.github/workflows/nighly_fuzzing.yml
@@ -1,0 +1,75 @@
+name: keth-NIGHTLY-FUZZING
+
+on:
+  schedule:
+    - cron: 20 0 * * *
+  workflow_dispatch: {}
+
+permissions: read-all
+
+jobs:
+  tests-unit:
+    runs-on: ubuntu-latest-16-cores
+    timeout-minutes: 360
+    env:
+      PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION: python
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: true
+          cache-dependency-glob: uv.lock
+      - uses: actions/setup-python@v5
+        with:
+          python-version-file: ./cairo/.python-version
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/nighly_fuzzing.yml
```diff
@@ -0,0 +1,56 @@
...
+          version: nightly
+      - name: Run tests
+        id: tests
+        continue-on-error: true # To send a notification if the tests fail
+        run: |
+          echo "HYPOTHESIS_PROFILE=nightly" >> .env
+          cd cairo
+          uv run compile
+          forge build
+          uv run pytest "tests" -m "not NoCI" -n logical --seed 42 --junitxml=junit.xml -o junit_family=legacy
```
obatirou, line 37: I don't know if we want/need to have a seed 
Does it influence hypothesis input generation ?

.github/workflows/nighly_fuzzing.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version-file: ./cairo/.python-version
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+      - name: Run tests
+        id: tests
+        continue-on-error: true # To send a notification if the tests fail
```
obatirou, line 31: Seems to make the step green even if failed

enitrat: yes, hence the

```
      - name: Fail workflow if tests failed
        if: steps.tests.outcome == 'failure'
        run: exit 1
```

at the end

.github/workflows/nighly_fuzzing.yml
enitrat: I think it only determines the seed for the random number generators, not the test themselves. you're right that it's probably best without.

ClementWalter (approved):

PR #217, opened 2024-12-06
fix: set code_address of the deployment code to the created address

### LINKED ISSUE #149: fix: set code_address of the deployment code to the created address
https://github.com/kkrt-labs/kakarot/pull/1587

obatirou:
Closes #149 

Set code_address of the deployment code to the created address

Note to reviewer: this is a fix imported from C4 mitigation, ensure the fix was correctly ported by looking at the corresponding issue and PR.

### DIFF
--- a/cairo/tests/programs/test_os.py
+++ b/cairo/tests/programs/test_os.py
@@ -229,7 +229,6 @@ class TestOs:
             state=State.model_validate(initial_state),
         )
 
-
     def test_create_opcode(self, cairo_run):
         plain_opcodes = get_contract("PlainOpcodes", "PlainOpcodes")
 
--- a/cairo/tests/programs/test_os.py
+++ b/cairo/tests/programs/test_os.py
@@ -187,3 +187,84 @@ class TestOs:
         bytes.fromhex(
             "7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3"
         ) == state["accounts"]["0x32dCAB0EF3FB2De2fce1D2E0799D36239671F04A"]["code"]
+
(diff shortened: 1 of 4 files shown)


cairo/tests/programs/test_os.py
```diff
@@ -187,3 +187,83 @@ def test_create_tx_returndata(self, cairo_run):
         bytes.fromhex(
             "7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3"
         ) == state["accounts"]["0x32dCAB0EF3FB2De2fce1D2E0799D36239671F04A"]["code"]
+
+    def test_jump_creation_code_deploy_tx_should_succeed(self, cairo_run):
```
obatirou, line 191: Those were not failling test in the first place as the behavior in keth is different:
The valid jumpdest fo the bytecode to deploy are passed in to the child message 
https://github.com/kkrt-labs/keth/blob/a37235b7f96123e2061a25a30f748976e03997ee/cairo/src/instructions/system_operations.cairo#L151

And the jump instruction uses those ones directly 
https://github.com/kkrt-labs/keth/blob/a37235b7f96123e2061a25a30f748976e03997ee/cairo/src/evm.cairo#L269

Hence the original issue raised in C4 is not present anymore.

Should those tests be kept to ensure there are no regression ?
@ClementWalter @enitrat

obatirou: After discussion IRL
The execution specs put `None` as `code_address`
https://github.com/ethereum/execution-specs/blob/717677f35ab3337b1c0dff459caa7df8d0beae21/src/ethereum/cancun/vm/instructions/system.py#L122

It seems the original issue was a problem in how jumpdest were handled.
This is now fix. A tracking issue will be created to include `None` into the message

Issue #242, opened 2024-12-10
bug: `state` hint is not generating account storages


enitrat:
The `state` hint defined here https://github.com/kkrt-labs/keth/blob/5b18a0e8ce48a3d5dba753bb823fdcb00d5721eb/cairo/tests/utils/hints.py#L119-L124

does not put the account's storage into cairo memory.

repro:

```
        def test_should_return_account_when_account_in_state(self, cairo_run):
            initial_state = {
                OWNER: {
                    "code": [],
                    "storage": {0x1: 0x2},
                    "balance": int(1e18),
                    "nonce": 0,
                }
            }
            account = cairo_run(
                "test__get_account",
                evm_address=int(OWNER, 16),
                state=State.model_validate(initial_state),
            )
            expected = {
                "code": bytes(initial_state[OWNER]["code"]),
                "code_hash": int.from_bytes(
                    keccak256(bytes(initial_state[OWNER]["code"])), "big"
                ),
                "balance": initial_state[OWNER]["balance"],
                "nonce": initial_state[OWNER]["nonce"],
                "storage": initial_state[OWNER]["storage"],
                "transient_storage": {},
                "valid_jumpdests": {},
                "selfdestruct": 0,
                "created": 0,
            }
            assert account == expected
```

output:

```
{'storage': {}} != {'storage': {1: 2}}
````

memory inspection in `get_account`: 

```
(Pdb) ids.account.storage_start.prev_value
*** starkware.cairo.lang.vm.memory_dict.UnknownMemoryError: Unknown value for memory cell at address 7:1.
(Pdb) ids.account.storage_start.new_value
*** starkware.cairo.lang.vm.memory_dict.UnknownMemoryError: Unknown value for memory cell at address 7:2.
(Pdb) ids.account.storage_start.key
*** starkware.cairo.lang.vm.memory_dict.UnknownMemoryError: Unknown value for memory cell at address 7:0.
````

ClementWalter:
I don't see which cairo code you are running but the "problem" comes from the fact that the dict values are never read, not that the dict manager doesn't have them. If you were accessing the storage slot you'll see the segment populated

enitrat:
I noticed this while doing https://github.com/kkrt-labs/keth/pull/244

ClementWalter:
you need to do like https://github.com/kkrt-labs/keth/blob/main/cairo/tests/utils/args_gen.py#L227-L231
in `gen_arg_pydantic`

PR #244, opened 2024-12-10
feat: create new, empty account when not in initial state

### LINKED ISSUE #243: feat: `get_account` should return a new, empty account if not in the initial state
If an account is first interacted with during the transaction that applies the state transition `STF(S, S')` and an account part of S' is not in the input S, then calling `get_account` will cast 0 to a `model.Account*` which is incorrect.

This scenario could happen when the first interaction with an account happens in the transaction. ~Because the account is not part of state S before the transaction, it's not required to have it as part of the sparse MPT given as input to compute the new root.~

enitrat:
Closes #243

### DIFF
--- a/cairo/src/account.cairo
+++ b/cairo/src/account.cairo
@@ -22,9 +22,24 @@ from src.model import model
 from src.utils.dict import dict_copy, dict_squash
 from src.utils.utils import Helpers
 from src.utils.bytes import keccak
+from src.constants import Constants
 
 namespace Account {
-    // @notice Create a new account
+    // @notice Creates a new, empty account
+    // @dev This is used to initialize a new, empty account in the state
+    // @dev when an account is part of a transaction, but was never interacted with before.
+    // @return The new account
+    func new_empty() -> model.Account* {
+        let (code) = alloc();
+        tempvar empty_code_hash = new Uint256(
+            low=Constants.EMPTY_CODE_HASH_LOW, high=Constants.EMPTY_CODE_HASH_HIGH
+        );
+        tempvar empty_balance = new Uint256(low=0, high=0);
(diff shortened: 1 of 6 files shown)


cairo/src/account.cairo
```diff
@@ -22,9 +22,24 @@ from src.model import model
...
 from src.utils.bytes import keccak
+from src.constants import Constants
 
 namespace Account {
-    // @notice Create a new account
+    // @notice Creates a new, empty account
+    // @dev This is used to initialize a new, empty account in the state
+    // @dev when an account is part of a transaction, but was never interacted with before.
+    // @return The new account
+    func new_empty() -> model.Account* {
```
ClementWalter, line 32: can we call this `default` instead? I like the rust default way

cairo/tests/src/test_state.py
```diff
@@ -10,3 +16,55 @@ def test_should_return_true_when_account_in_state(self, cairo_run):
...
+                state=State.model_validate(initial_state),
+            )
+            expected = {
+                "code": bytes(initial_state[OWNER]["code"]),
+                "code_hash": int.from_bytes(
+                    keccak256(bytes(initial_state[OWNER]["code"])), "big"
+                ),
+                "balance": initial_state[OWNER]["balance"],
+                "nonce": initial_state[OWNER]["nonce"],
+                "storage": initial_state[OWNER]["storage"],
```
ClementWalter, line 42: storage keys are hashed so you cannot just do this

enitrat:
blocked by: https://github.com/kkrt-labs/keth/issues/242 (storage dict issue for test)

ClementWalter (approved):

PR #245, opened 2024-12-10
dev: add pre-push trunk hook

enitrat:
Adds the pre-push trunk hook back, to avoid cases where CI fails because of a lint check

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -88,7 +88,7 @@ lint:
 actions:
   disabled:
     - trunk-announce
-    - trunk-check-pre-push
     - trunk-fmt-pre-commit
   enabled:
+    - trunk-check-pre-push
     - trunk-upgrade-available


.trunk/trunk.yaml
```diff
@@ -88,7 +88,7 @@ lint:
 actions:
   disabled:
     - trunk-announce
-    - trunk-check-pre-push
     - trunk-fmt-pre-commit
```
obatirou, line 91: fmt pre-commit could also be interesting no ?

enitrat: no, i find it really annoying in the dev flow , as you commit a lot more than you push, which interrupts the flow

obatirou: I see 
then a fmt pre push would be better

obatirou (approved):

.trunk/trunk.yaml
enitrat: no because for a fmt pre-push you need an automated commit system, which is annoying, and you can't always fix the check issues automatically - this just replicates what we had on the kakarot, which I think was good

PR #267, opened 2024-12-12
Add caching for tests

ClementWalter:
## Why

To speed up dev and CI, we want to skip tests that have not changed.

## What

Skip tests if nothing has changed.

## How

After compilation of the target cairo program, we compute the hash of the program and the test file.
We use pytest-cache to store this hash.

If the hash has not changed, we skip the test.

The flag `--skip-cached-tests` is used to skip the test, with default value `True`.
The flag `--no-skip-cached-tests` is used to force running the tests, and update the cache key.

Tests are not skipped when running with the nightly profile.

Solves: #280


.github/workflows/python_tests.yml
```diff
@@ -44,21 +45,27 @@ jobs:
...
         run: forge build
 
+      - name: Load pytest cache
+        uses: actions/cache@v4
+        with:
+          path: cairo/.pytest_cache
+          key: ${{ runner.os }}-pytest
+
       - name: Run tests with src
         if: ${{ needs.paths-filter.outputs.src == 'true' }}
```
obatirou, line 55: Should the path filterer be removed ?

cairo/tests/fixtures/runner.py
```diff
@@ -272,3 +302,9 @@ def _factory(entrypoint, *args, **kwargs):
         return final_output
 
     return _factory
+
+
+def pytest_sessionfinish(session, exitstatus):
```
obatirou, line 307: Why not in conf test and no `@pytest.hookimpl()` ?
How is it discovered and called ?
It is not well documented in pytest it seems

cairo/tests/utils/caching.py
```diff
@@ -0,0 +1,35 @@
...
+import xxhash
+
+
+def testfile_hash(file_path: str | Path) -> bytes:
+    """
+    A simple, fast, hash of the test file.
+    """
+    file_path = Path(file_path)
+    with open(file_path, "rb") as f:
+        file_hash = xxhash.xxh64(f.read()).digest()
```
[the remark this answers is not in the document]
ClementWalter: cursor told me to use this 🤷

obatirou (dismissed):

.github/workflows/python_tests.yml
```diff
@@ -48,24 +33,16 @@ jobs:
...
-          cd cairo
-          uv run compile_os
-          uv run pytest -n logical --ignore=tests/src --ignore=tests/programs --junitxml=junit.xml -o junit_family=legacy --durations=0 -v
-
-      - uses: codecov/codecov-action@v5.0.5
-        with:
-          directory: ./cairo/coverage/
-      - uses: codecov/test-results-action@v1
-        if: ${{ !cancelled() }}
+          uv run pytest -n logical --durations=0 -v
```
obatirou, line 48: nit: for nightly tests, there is "tests" specified 
`uv run pytest "tests" -n logical --durations=0 --no-skip-cached-tests`

cairo/tests/src/utils/test_signature.py
```diff
@@ -66,12 +66,10 @@ def test_should_raise_with_invalid_y_parity(
...
-                st.just(U256(0)),
-                st.integers(
-                    min_value=int(SECP256K1N), max_value=int(U256.MAX_VALUE)
-                ).map(U256),
-            ),
-            s=st.integers(min_value=1, max_value=int(SECP256K1N) - 1).map(U256),
+                st.just(0),
+                st.integers(min_value=int(SECP256K1N), max_value=int(U256.MAX_VALUE)),
+            ).map(U256),
+            s=...,
```
obatirou, line 72: this should be 
```suggestion
            s=st.integers(min_value=1, max_value=int(SECP256K1N) - 1).map(U256),
```

obatirou (approved):

PR #272, opened 2024-12-13
feat: use the same compiler in cairo_compile and in tests

### LINKED ISSUE #211: feat: use the same compiler in cairo_compile and in tests
## Why

The test runner currently uses the internal compile function from SW while the CLI does some more checks. Especially these fixtures implements the hints.

## What

In `compile_cairo.py`, use the same logic as in `cairo/tests/fixtures/compiler.py` and dump the final program as json.

```
def cairo_compile(path):
    module_reader = get_module_reader(cairo_path=[str(Path(__file__).parents[2])])

    pass_manager = default_pass_manager(
        prime=DEFAULT_PRIME, read_module=module_reader.read
    )

    return compile_cairo(
        Path(path).read_text(),
        pass_manager=pass_manager,
        debug_info=True,
    )
```

dump the program with 

```
json.dump(
    assembled_program.Schema().dump(assembled_program), out, indent=4, sort_keys=True
)
```

obatirou:
Close #211

### DIFF
--- a/cairo/src/utils/compile_cairo.py
+++ b/cairo/src/utils/compile_cairo.py
@@ -1,10 +1,9 @@
+import argparse
 import json
 import logging
-import sys
 from pathlib import Path
 
-from tests.fixtures.compiler import cairo_compile
-from tests.utils.hints import implement_hints
+from src.utils.compiler import cairo_compile, implement_hints
 
 # Configure the logger
 logging.basicConfig(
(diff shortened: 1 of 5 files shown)


cairo/src/utils/compile_cairo.py
```diff
@@ -13,30 +16,11 @@
 def compile_cairo(file_name):
     input_path = Path(file_name)
     output_path = input_path.with_suffix(".json")
+    program = cairo_compile(input_path, debug_info=False, proof_mode=True)
+    program.hints = implement_hints(program)
```
ClementWalter, line 20: add a CLI arg to make this optional otherwise it won't wokr on the rust vm

cairo/src/utils/compile_cairo.py
```diff
@@ -1,8 +1,11 @@
+import json
 import logging
-import subprocess
 import sys
 from pathlib import Path
 
+from tests.fixtures.compiler import cairo_compile
+from tests.utils.hints import implement_hints
```
ClementWalter, line 7: import in test and not here

obatirou:
Test failling are the same ones than with https://github.com/kkrt-labs/keth/pull/269, the latter can be merged as the tests are taking too long to finish for now and are reported are failure

ClementWalter (approved):

PR #281, opened 2024-12-17
feat: error types

### LINKED ISSUE #257: feat: design the error type
## How

Exceptions are implemented with

```
struct Exception {
    value: ExceptionStruct*,
}

struct ExceptionStruct {
    value: Bytes,
    len: felt,
}
```

Custom exception can be created with, e.g.

```
struct StackUnderflow {
    value: ExceptionStruct*,
}
```

when raising an exception, the exception is a valid pointer, otherwise it's 0:

- this is an error with an empty message
```
tempvar error = StackUnderflow(new ExceptionStruct(cast(0, felt*), 0));
```

- this is not an error
```
tempvar no_error = StackUnderflow(cast(0, ExceptionStruct*));
```

- [ ] for each implemented exception, add in the mapping as any type
- [ ] in `serde.py`, add the case where python_cls is an Exception

```
        if ptr == 0 and isinstance(path, Exception):
            return NO_ERROR_FLAG

        if isinstance(python_cls, Exception):
            raise python_cls

NO_ERROR_FLAG = object()  # Creates a unique sentinel object

        if isinstance(cairo_type, TypeTuple):
            raw = [
                self._serialize(m.typ, ptr + i)
                for i, m in enumerate(cairo_type.members)
            ]
            return [x for x in raw if x is not NO_ERROR_FLAG]
```

- [ ] add in `test_serde` a new test function that should raises to test these error types

```
with pytest.raises(Exception):
```

enitrat:
Closes #257

### DIFF
--- a/cairo/ethereum/exceptions.cairo
+++ b/cairo/ethereum/exceptions.cairo
@@ -1,0 +1,25 @@
+// Error types common across all Ethereum forks.
+//
+// When raising an exception, the exception is a valid pointer. Otherwise, the pointer is `0`.  When
+// checking for an exceptino, a simple cast(maybe_exception, felt) != 0 is enough to check if the
+// function raised.
+//
+// Example:
+// This is an error:
+// ```
+// tempvar inner_error = new BytesStruct(cast(0, felt*), 0);
+// let error = EthereumException(inner_error);
+// ```
+//
+// This is not an error:
+// ```
+// tempvar no_error = EthereumException(cast(0, BytesStruct*));
+// ```
+
+from ethereum_types.bytes import BytesStruct
+
+// @notice Base type for all exceptions _expected_ to be thrown during normal
+// operation.
+struct EthereumException {
+    value: BytesStruct*,
+}
--- a/cairo/ethereum/exceptions.cairo
+++ b/cairo/ethereum/exceptions.cairo
@@ -1,6 +1,6 @@
 // Error types common across all Ethereum forks.
 //
-// When raising an exception, the exception is a valid pointer. Otherwise, the pointer is `0`.  When
+// When raising an exception, the exception is a valid pointer. Otherwise, the pointer is `0`. When
 // checking for an exceptino, a simple cast(maybe_exception, felt) != 0 is enough to check if the
 // function raised.
 //
--- a/cairo/tests/test_serde.py
+++ b/cairo/tests/test_serde.py
@@ -176,15 +176,18 @@ class TestSerde:
         result = serde.serialize(to_cairo_type(type_), base, shift=0)
         assert result == b
 
-    def test_exceptions(self, to_cairo_type, segments, serde, gen_arg):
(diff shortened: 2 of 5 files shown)


cairo/ethereum/exceptions.cairo
```diff
@@ -0,0 +1,25 @@
+// Error types common across all Ethereum forks.
+//
+// When raising an exception, the exception is a valid pointer. Otherwise, the pointer is `0`.  When
```
obatirou, line 3:
```suggestion
// When raising an exception, the exception is a valid pointer. Otherwise, the pointer is `0`. When
```

cairo/tests/test_serde.py
```diff
@@ -173,3 +174,16 @@ def test_type(
...
+        base = gen_arg(EthereumException, None)
+        result = serde.serialize(to_cairo_type(EthereumException), base, shift=0)
+        assert result is None
+
+        # Test error case - cast(base, felt) != 0
+        expected_error = EthereumException("Some Error")
+        base = segments.gen_arg([gen_arg(EthereumException, expected_error)])
+        with pytest.raises(EthereumException) as exception:
+            serde.serialize(to_cairo_type(EthereumException), base, shift=0)
+        assert str(exception.value) == str(expected_error)
```
ClementWalter, line 189: plz do unit testing, no two tests in the same test

cairo/tests/utils/args_gen.py
```diff
@@ -274,6 +276,18 @@ def _gen_arg(
             else int.from_bytes(arg, "little")
         )
 
+    if issubclass(arg_type, Exception):
+        # For exceptions, we either return 0 (no error) or create an error with a message
+        if arg is None:
+            return 0
+
+        error_bytes = bytes(str(arg), "utf-8")
```
ClementWalter, line 284: fyi

```suggestion
        error_bytes = str(arg).encode()
```

cairo/tests/utils/serde.py
```diff
@@ -195,6 +209,18 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
                 return bytes(data).decode()
             return python_cls(data)
 
+        if python_cls and issubclass(python_cls, Exception):
+            tuple_struct_ptr = self.serialize_pointers(path, ptr)["value"]
+            if not tuple_struct_ptr:
+                return NO_ERROR_FLAG
+            # All exceptions share the same inner struct "BytesStruct"
+            struct_name = "BytesStruct"
```
ClementWalter, line 217: avoid using hard coded struct name in serde

cairo/tests/test_serde.py
ClementWalter: can't we have hypothesis to generate errors ? I'd have liked to have the same as `test_type` with a list of Exceptions as we add some more

ClementWalter (approved):

PR #304, opened 2024-12-19
docs: code architecture docs & cursorrules

enitrat:


.cursorrules
```diff
@@ -0,0 +1,489 @@
...
+
+1. Simple types (direct value):
+
+```cairo
+struct U64 {
+    value: felt
+}
+```
+
+A `null` U64 is represented by the outer U64\* with a value of 0.
```
obatirou, line 158: Not sure about the formulation here. But not sure about the proposition also
```suggestion
A `null` U64 is represented by a U64 pointer with a value of 0.
```
or 
```suggestion
A `null` U64 is represented by a U64* with a value of 0.
```
or just 
```suggestion
A `null` U64 is represented by a outer U64\* with a value of 0.
```

.cursorrules
```diff
@@ -0,0 +1,489 @@
...
+
+2. Complex types (pointer-based):
+
+```cairo
+struct U256 {
+    value: U256Struct*,  // null when pointer is 0
```
obatirou, line 164: Not sure how it will be interpreted by cursor: a U256 value is _considered_ null if the U256Struct* is 0. To understand this comment you need to put it in context.

```suggestion
    value: U256Struct*,  // [value is] considered null when pointer is 0
```

Maybe I am nitpicking

.cursorrules
```diff
@@ -0,0 +1,489 @@
...
+    value: MappingBytesBytesStruct*,
+}
+```
+
+Key implementation notes:
+
+- Keys stored as pointers, thus querying a key is a pointer equality check, not
+  a value equality check.
+- In the future, we might consider using key hashes for content-based comparison
+  rather than pointer comparison.
```
obatirou, line 284:
```suggestion
- Keys are stored as pointers, thus querying a key is a pointer equality check, not
  a value equality check.
- In the future, we consider using key hashes for content-based comparison
  rather than pointer comparison.
```

docs/code_architecture.md
```diff
@@ -0,0 +1,508 @@
+# Keth codebase architecture
```
obatirou, line 1: Any changes in the cursorrules would need to be ported here

cairo/_typos.toml
ClementWalter, line 1: add it instead of cspell in trunk then

enitrat: there's no typos plugin for trunk :/

enitrat: will do in later PR

docs/code_architecture.md
```diff
@@ -0,0 +1,467 @@
...
+    def test_operation(self, cairo_run, ...):
+        # Test success case
+        result_cairo = cairo_run("operation", ...)
+        result_py = operation(...)
+        assert result_cairo == result_py
+
+        # Test error case
+        with pytest.raises(ErrorType):
+            cairo_run("operation", ...)
+```
```
ClementWalter, line 98: split in two tests

PR #305, opened 2024-12-19
dev: add SetAddress cairo type to prepare for EVM object

### LINKED ISSUE #293: dev: implement Set[Address] type
to prepare for [EVM struct](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/__init__.py#L79), implement the `Set[Address]` type for key [`account_to_delete`](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/vm/__init__.py#L94)

```[tasklist]
### Tasks
- [x] create SetAddress struct in `fork_types.cairo` next to `Address`
- [x] create case in args_gen.py `("ethereum", "cancun", "fork_types", "SetAddress"): Set[Address],`
- [x] create test case in `test_serde.py`
```

Eikix:
resolves #293 

implement SetAddress as `EVM` object has `address_to_delete` field that is a set of addresses

### DIFF
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -2,18 +2,18 @@
 # To learn more about the format of this file, see https://docs.trunk.io/reference/trunk-yaml
 version: 0.1
 cli:
-  version: 1.22.7
+  version: 1.22.8
 # Trunk provides extensibility via plugins. (https://docs.trunk.io/plugins)
 plugins:
   sources:
     - id: trunk
-      ref: v1.6.4
+      ref: v1.6.6
       uri: https://github.com/trunk-io/plugins
 # Many linters and tools depend on runtimes - configure them here. (https://docs.trunk.io/runtimes)
 runtimes:
   enabled:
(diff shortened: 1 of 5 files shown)


Eikix:
Is part of https://github.com/kkrt-labs/keth/issues/303

.trunk/trunk.yaml
```diff
@@ -48,7 +48,7 @@ lint:
         - output: rewrite
           success_codes: [0]
           formatter: true
-          run: cairo-format ${target}
+          run: .venv/bin/cairo-format ${target}
```
obatirou, line 51: My bad for this. 
You should have cairo-format in your path in local in fact (symbolic link or installed in your system) and not modify this line.
Other solution to avoid failure,  is to modify the CI to create the .venv (hence uv/.venv would need to be setup) in the workflow trunk_check.yml as for now the cairo-lang is installed with pip.

ClementWalter (approved):

Issue #141, opened 2024-11-28
feat: use poseidon instead of pedersen builtins


ClementWalter:
## Why

According to Starkware's latest numbers, poseidon is 10x cheaper than pedersen.

## What

In the current codebase, we use hashing only to compute storage keys in the [account.cairo](../blob/main/cairo/src/account.cairo#L467-L470)

## How

Replace all the `pedersen` instances with a `poseidon` one

Thirumurugan7:
I would like to work on this issue

PR #333, opened 2024-12-31
feat: replace pedersen with poseidon

varun-doshi:
Ref https://github.com/kkrt-labs/keth/issues/141


ClementWalter:
hey @varun-doshi thanks for taking the time to look into this. This is a more global issue than just few pointers in the account file. But we're also completly removing all these files eventually so it's not necessary to spend much time on this. once the codebase stabilizes in the new `ethereum` folder, we'll open new good first issues

PR #365, opened 2025-01-06
feat: memory opcodes

### LINKED ISSUE #351: memory opcodes
https://github.com/ethereum/execution-specs/blob/4d9cc42b0e89cac652f54811e0683ffb896975c5/src/ethereum/cancun/vm/instructions/memory.py

enitrat:
Resolves #351


cairo/ethereum/cancun/vm/instructions/memory_instructions.cairo
```diff
@@ -0,0 +1,259 @@
...
+from src.utils.utils import Helpers
+
+// @notice Stores a word to memory
+func mstore{range_check_ptr, evm: Evm}() -> ExceptionalHalt* {
+    alloc_locals;
+    // STACK
+    let stack = evm.value.stack;
+    with stack {
+        let (start_position, err1) = pop();
+        if (cast(err1, felt) != 0) {
```
Eikix, line 35: use `err` everywhere imo and shadow it as code goes

cairo/ethereum/cancun/vm/instructions/memory_instructions.cairo
```diff
@@ -0,0 +1,259 @@
...
+    }
+
+    // GAS
+    tempvar mem_access_tuple = new TupleU256U256(
+        new TupleU256U256Struct(start_position, U256(new U256Struct(32, 0)))
+    );
+    tempvar mem_access_list = ListTupleU256U256(new ListTupleU256U256Struct(mem_access_tuple, 1));
+    let extend_memory = calculate_gas_extend_memory(evm.value.memory, mem_access_list);
+
+    // assumed that cost < 2**110 (see calculate_memory_gas_cost)
```
Eikix, line 143: what does it mean that it is assumed?
what happens if cost > 2**110?

obatirou: calculate_gas_extend_memory will saturate if size < 2**64
If not mistaken the cost will be < 2**110 
If those are not respected, it might not be sound

cairo/ethereum/cancun/vm/instructions/memory_instructions.cairo
```diff
@@ -0,0 +1,259 @@
...
+        destination, U256(new U256Struct(length.value.low, 0))
+    );
+    tempvar mem_access_list = ListTupleU256U256(
+        new ListTupleU256U256Struct(new TupleU256U256(mem_access_tuples), 2)
+    );
+    let extend_memory = calculate_gas_extend_memory(evm.value.memory, mem_access_list);
+
+    // copy_gas_cost in [0, 3 * 2**120)
+    // extend_memory.value.cost.value in [0, 2**110)
+    // -> sum < felt_size, no overflow
```
Eikix, line 240: should we add invariants? such as `assert sum < PRIME - 1` (this doesnt work since it may overflow and if sum wraps around, then obv it'll be less than prime but still, maybe use overflowing math)

c.f. [primeagen](https://www.youtube.com/watch?v=jc_6ZkkOUpQ) and [tiger beetle](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md)

plus it'll help us when we switch primes

cairo/tests/ethereum/cancun/vm/instructions/test_memory_instructions.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_mcopy(self, cairo_run, evm: Evm):
+        try:
+            cairo_result = cairo_run("mcopy", evm)
+        except ExceptionalHalt as cairo_error:
+            with pytest.raises(type(cairo_error)):
+                mcopy(evm)
+            return
+
+        mcopy(evm)
+        assert evm == cairo_result
```
Eikix, line 69: noob question but is this equality `==` a deep equality of the evm objects? since the memory is nested

cairo/ethereum/cancun/vm/memory.cairo
```diff
@@ -56,6 +66,12 @@ func memory_write{range_check_ptr, memory: Memory}(start_position: U256, value:
 func memory_read_bytes{memory: Memory}(start_position: U256, size: U256) -> Bytes {
     alloc_locals;
 
+    // Early return if nothing to read
+    if (size.value.low == 0) {
```
obatirou, line 70: should this be after the check on `size.value.high = 0` ?
As here in the case size = (0, 1) it should revert instead of returning empty

obatirou: same for buffer_read

cairo/ethereum/cancun/vm/instructions/memory_instructions.cairo
enitrat: > what happens if cost > 2**110?

it won't happen because the cost is saturated

cairo/ethereum/cancun/vm/memory.cairo
enitrat: yes !

ClementWalter (approved):

PR #377, opened 2025-01-08
dev: replace bytes strategies by integer of fixed size

### LINKED ISSUE #376: dev(hypothesis): optimize bytes-related strategies
```[tasklist]
### Tasks
- [x] try changing bytes strategies from random byte per byte generation to integer generation followed by type conversion to bytes
```

Eikix:
- in strategies.py, when an array of bytes is of fixed size, i.e. in `st.binary` `min_size == max_size`, then replace it by an integer of the correct size and map it to its bytes repr in little endian

closes #376


ClementWalter:
Can we have a screenshot of profiling before and after?

ClementWalter (dismissed):

Eikix:
> Can we have a screenshot of profiling before and after?

Ok working on it

enitrat (changes requested):
cache issue but fails locally

Eikix:
# BEFORE

image


# AFTER
image

enitrat (approved):

PR #378, opened 2025-01-09
feat: trie get and hashed dicts

enitrat:


cairo/ethereum/cancun/trie.cairo
```diff
@@ -554,6 +612,67 @@ func _search_common_prefix_length{
     return _search_common_prefix_length(obj + BytesBytesDictAccess.SIZE, current_length);
 }
 
+func _get_branche_for_nibble_at_level_inner{poseidon_ptr: PoseidonBuiltin*}(
```
obatirou, line 615:
```suggestion
func _get_branch_for_nibble_at_level_inner{poseidon_ptr: PoseidonBuiltin*}(
```

cairo/tests/utils/serde.py
```diff
@@ -274,25 +277,85 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
...
+                        hashed_key = poseidon_hash_many(key)
+                        preimage = sum(
+                            felt * 2 ** (128 * i) for i, felt in enumerate(key)
+                        )
+                        serialized_dict[preimage] = dict_data[hashed_key]
+
+                    elif python_key_type == Bytes:
+                        hashed_key = poseidon_hash_many(key)
+                        preimage = bytes(list(key))
+                        serialized_dict[preimage] = dict_data[hashed_key]
```
obatirou, line 327: add a default case in case no match ?

cairo/tests/utils/serde.py
```diff
@@ -274,25 +277,85 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
...
+            dict_data = {
+                self._serialize(cairo_key_type, dict_ptr + i): self._serialize(
                     value_type, dict_ptr + i + 2
                 )
                 for i in range(0, segment_size, 3)
             }
 
+            serialized_dict = {}
+            tracker_data = self.dict_manager.trackers[dict_ptr.segment_index].data
+            if isinstance(cairo_key_type, TypeFelt):
```
obatirou, line 306: A set of felt would match this case no ?

enitrat: no, because the cairo_key_type is a Uint

cairo/tests/utils/serde.py
obatirou: how would it panic ?
I was suggesting to add a default case (else) with a raise as here it can go through the loop with no match and continue execution

enitrat: I thought you meant `dict_data[hashed_key]`

For the default case yes we could add a raise. Will do it in next PR if CI passes this one

PR #395, opened 2025-01-10
feat: improve rust bindings compat

enitrat:
improve rust bindings compat, enabling syntaxes like
- segments.memory.get(idx)
or 
- dict_trackers[idx] = X

based on a custom fork of the cairo vm in which I unbotherly created a special `DictKey` type which is an 
enum

```rust
#[derive(
    Eq, Ord, Hash, PartialEq, PartialOrd, Clone, Debug,
)]
pub enum DictKey {
    Simple(MaybeRelocatable),
    Compound(Vec<MaybeRelocatable>),
}
```

Which allows a Vec of Felts / Relocatable to be used as dict keys. For us this is important because we need to track preimages that might not fit in a felt inside the dict tracker. 

This was implemented in a "dumb" way that is unconsidering of external side effects to the Cairo-VM - once synced with LC we will want to revisit the approach. see https://github.com/kkrt-labs/cairo-vm/tree/kkrt-compat

### DIFF
--- a/crates/cairo-addons/src/vm/dict_manager.rs
+++ b/crates/cairo-addons/src/vm/dict_manager.rs
@@ -1,13 +1,83 @@
 use cairo_vm::{
     hint_processor::builtin_hint_processor::dict_manager::{
-        DictManager as RustDictManager, DictTracker,
+        DictKey as RustDictKey, DictManager as RustDictManager, DictTracker,
     },
     types::relocatable::MaybeRelocatable,
 };
-use pyo3::prelude::*;
+use pyo3::{prelude::*, types::PyTuple};
 use std::{cell::RefCell, collections::HashMap, rc::Rc};
 
 use super::{maybe_relocatable::PyMaybeRelocatable, relocatable::PyRelocatable};
+
+#[derive(FromPyObject, Eq, PartialEq, Hash)]
+pub enum PyDictKey {
+    #[pyo3(transparent)]
+    Simple(PyMaybeRelocatable),
+    #[pyo3(transparent)]
+    Compound(Vec<PyMaybeRelocatable>),
+}
+
+impl IntoPy<PyObject> for PyDictKey {
+    fn into_py(self, py: Python<'_>) -> PyObject {
+        match self {
+            PyDictKey::Simple(val) => val.into_py(py),
+            PyDictKey::Compound(vals) => {
+                // Convert Vec to tuple
+                let elements: Vec<PyObject> = vals.into_iter().map(|v| v.into_py(py)).collect();
+                PyTuple::new_bound(py, elements).into()
+            }
+        }
+    }
+}
+
+impl From<PyDictKey> for RustDictKey {
+    fn from(value: PyDictKey) -> Self {
+        match value {
+            PyDictKey::Simple(val) => RustDictKey::Simple(val.into()),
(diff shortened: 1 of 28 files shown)


crates/cairo-addons/src/vm/memory_segments.rs
```diff
@@ -10,8 +10,26 @@ pub struct PyMemorySegmentManager {
...
+/// Enables syntax `segments.memory.<op>`
+#[pyclass(name = "MemoryWrapper", unsendable)]
+pub struct PyMemoryWrapper {
+    pub(crate) runner: *mut RustCairoRunner,
+}
+
+#[pymethods]
+impl PyMemoryWrapper {
+    fn get(&self, key: PyRelocatable) -> Option<PyMaybeRelocatable> {
+        unsafe { (*self.runner).vm.get_maybe(&key.inner).map(PyMaybeRelocatable::from) }
```
Eikix, line 22: why unsafe?

enitrat: that's how it was. I'm not modifying this

crates/cairo-addons/src/vm/dict_manager.rs
```diff
@@ -47,24 +123,22 @@ impl PyDictManager {
...
 }
 
 #[pymethods]
 impl PyDictTracker {
-    // Note: This is a temporary implementation, need to understand why HashMap<PyMaybeRelocatable,
-    // PyMaybeRelocatable> is not working
     #[new]
-    #[pyo3(signature = (keys, values, current_ptr, default_value=None))]
+    #[pyo3(signature = (data, current_ptr, default_value=None))]
     fn new(
```
Eikix, line 135: this look like a from method rather than a new?

crates/cairo-addons/src/vm/dict_manager.rs
```diff
@@ -75,4 +149,33 @@ impl PyDictTracker {
...
+    fn current_ptr(&self) -> PyRelocatable {
+        PyRelocatable { inner: self.inner.current_ptr }
+    }
+
+    #[getter]
+    fn data(&self) -> HashMap<PyDictKey, PyMaybeRelocatable> {
+        self.inner
+            .get_dictionary_ref()
+            .iter()
+            .map(|(k, v)| (PyDictKey::from(k.clone()), PyMaybeRelocatable::from(v.clone())))
```
Eikix, line 163: why are we cloning entire dictionaries when getting them?

crates/cairo-addons/src/vm/dict_manager.rs
enitrat: what?

Eikix: idiomatically, it looks like this function could be called `from_py_dict` or smth that suggests the hashmap data exists, with keys and values of python types

enitrat: no, it's a pyo3 constructor

crates/cairo-addons/src/vm/dict_manager.rs
enitrat: the values are behind shared references and need to be moved to the python objects.

Eikix (approved):

PR #412, opened 2025-01-13
feat: rust vm hints

enitrat:
Adds most common hints to run the tests in the Rust VM.

### DIFF
--- a/cairo/ethereum/cancun/trie.cairo
+++ b/cairo/ethereum/cancun/trie.cairo
@@ -10,7 +10,7 @@ from starkware.cairo.common.cairo_builtins import KeccakBuiltin
 from starkware.cairo.common.memcpy import memcpy
 
 from src.utils.bytes import uint256_to_bytes32_little
-from src.utils.dict import hashdict_read, hashdict_write
+from src.utils.dict import hashdict_read, hashdict_write, dict_new_empty
 from ethereum.crypto.hash import keccak256
 from ethereum.utils.numeric import min, is_zero
 from ethereum.rlp import encode, _encode_bytes, _encode
@@ -814,8 +814,7 @@ func _get_branch_for_nibble_at_level{poseidon_ptr: PoseidonBuiltin*}(
     alloc_locals;
     // Allocate a segment for the branch and register an associated tracker
     // dict_new expectes an initial_dict hint argument.
-    %{ initial_dict = {} %}
-    let (branch_start_: DictAccess*) = dict_new();
+    let (branch_start_: DictAccess*) = dict_new_empty();
     let branch_start = cast(branch_start_, BytesBytesDictAccess*);
     let dict_ptr_stop = obj.value.dict_ptr;
 
--- a/cairo/ethereum/cancun/trie.cairo
+++ b/cairo/ethereum/cancun/trie.cairo
@@ -715,7 +715,7 @@ func _search_common_prefix_length{
         return current_length;
     }
 
-    let preimage = _get_preimage_for_key(obj, dict_ptr_stop);
+    let preimage = _get_preimage_for_key(obj.key.value, dict_ptr_stop);
     tempvar sliced_key = Bytes(
         new BytesStruct(preimage.value.data + level.value, preimage.value.len - level.value)
     );
@@ -741,7 +741,7 @@ func _get_branch_for_nibble_at_level_inner{poseidon_ptr: PoseidonBuiltin*}(
         return (branch_ptr, value);
     }
 
-    let preimage = _get_preimage_for_key(dict_ptr, dict_ptr_stop);
+    let preimage = _get_preimage_for_key(dict_ptr.key.value, dict_ptr_stop);
 
     // Check cases
     let is_value_case = is_zero(preimage.value.len - level);
@@ -764,14 +764,11 @@ func _get_branch_for_nibble_at_level_inner{poseidon_ptr: PoseidonBuiltin*}(
         assert [branch_ptr].prev_value = dict_ptr.prev_value;
         assert [branch_ptr].new_value = dict_ptr.new_value;
 
-        // Add an entry to the dict_tracker
-        %{
-            obj_tracker = __dict_manager.get_tracker(ids.dict_ptr_stop.address_)
-            dict_tracker = __dict_manager.get_tracker(ids.branch_ptr.address_)
-            dict_tracker.current_ptr += ids.DictAccess.SIZE
-            preimage = next(key for key in obj_tracker.data.keys() if poseidon_hash_many(key) == ids.dict_ptr.key.value)
-            dict_tracker.data[preimage] = obj_tracker.data[preimage]
-        %}
+        // Copy the entry from the dict_ptr's tracker to the branch_ptr's tracker
+        let source_key = dict_ptr.key.value;
+        let source_ptr_stop = dict_ptr_stop;
+        let dest_ptr = branch_ptr;
+        %{ copy_hashdict_tracker_entry %}
 
         return _get_branch_for_nibble_at_level_inner(
             dict_ptr + BytesBytesDictAccess.SIZE,
@@ -954,7 +951,7 @@ func _get_branches{poseidon_ptr: PoseidonBuiltin*}(obj: MappingBytesBytes, level
         assert value = value_15;
         assert value_set = 1;
     }
-    %{ ids.value_set = memory.get(fp + 2) or 0 %}
+    %{ value_set_or_zero %}
     if (value_set != 1) {
         let (data: felt*) = alloc();
         tempvar empty_bytes = Bytes(new BytesStruct(data, 0));
@@ -967,28 +964,22 @@ func _get_branches{poseidon_ptr: PoseidonBuiltin*}(obj: MappingBytesBytes, level
 
 // @notice Given a key (inside `dict_ptr`), returns the preimage of the key registered in the tracker.
 // The preimage is validated to be correctly provided by the prover by hashing it and comparing it to the key.
+// @param key - The key to get the preimage for. Either a hashed or non-hashed key - but it must be a felt.
+// @param dict_ptr_stop - The pointer to the end of the dict segment, the one registered in the tracker.
 func _get_preimage_for_key{poseidon_ptr: PoseidonBuiltin*}(
-    dict_ptr: BytesBytesDictAccess*, dict_ptr_stop: BytesBytesDictAccess*
+    key: felt, dict_ptr_stop: BytesBytesDictAccess*
 ) -> Bytes {
     alloc_locals;
 
     // Get preimage data
     let (local preimage_data: felt*) = alloc();
     local preimage_len;
-    %{
-        from starkware.cairo.lang.vm.crypto import poseidon_hash_many
-        hashed_value = ids.dict_ptr.key.value
-        dict_tracker = __dict_manager.get_tracker(ids.dict_ptr_stop)
-        # Get the key in the dict that matches the hashed value
-        preimage = bytes(next(key for key in dict_tracker.data.keys() if poseidon_hash_many(key) == hashed_value))
-        segments.write_arg(ids.preimage_data, preimage)
-        ids.preimage_len = len(preimage)
-    %}
+    %{ get_preimage_for_key %}
 
     // Verify preimage
     let (preimage_hash) = poseidon_hash_many(preimage_len, preimage_data);
     with_attr error_message("preimage_hash != key") {
-        assert preimage_hash = dict_ptr.key.value;
+        assert preimage_hash = key;
     }
 
     tempvar res = Bytes(new BytesStruct(preimage_data, preimage_len));
@@ -1013,7 +1004,8 @@ func patricialize{
     }
 
     let arbitrary_value = obj.value.dict_ptr_start.new_value;
-    let preimage = _get_preimage_for_key(obj.value.dict_ptr_start, obj.value.dict_ptr);
+    let current_key = obj.value.dict_ptr_start.key.value;
+    let preimage = _get_preimage_for_key(current_key, obj.value.dict_ptr);
 
     // if leaf node
     if (len == 1) {
--- a/cairo/ethereum/cancun/trie.cairo
+++ b/cairo/ethereum/cancun/trie.cairo
@@ -351,11 +351,7 @@ func copy_TrieAddressOptionalAccount{range_check_ptr, trie: TrieAddressOptionalA
 
     local new_dict_ptr: AddressAccountDictAccess*;
     tempvar original_mapping = trie.value._data.value;
-    %{
-        dict_tracker = __dict_manager.get_tracker(ids.original_mapping.dict_ptr)
-        copied_data = dict_tracker.data
-        ids.new_dict_ptr = __dict_manager.new_dict(segments, copied_data)
-    %}
+    %{ copy_dict_segment %}
 
     tempvar res = TrieAddressOptionalAccount(
         new TrieAddressOptionalAccountStruct(
@@ -375,13 +371,7 @@ func copy_trieBytes32U256{range_check_ptr, trie: TrieBytes32U256}() -> TrieBytes
 
     local new_dict_ptr: Bytes32U256DictAccess*;
     tempvar original_mapping = trie.value._data.value;
-    %{
-        from starkware.cairo.lang.vm.crypto import poseidon_hash_many
-
-        dict_tracker = __dict_manager.get_tracker(ids.original_mapping.dict_ptr)
-        copied_data = dict_tracker.data
-        ids.new_dict_ptr = __dict_manager.new_dict(segments, copied_data)
-    %}
+    %{ copy_dict_segment %}
 
     tempvar res = TrieBytes32U256(
         new TrieBytes32U256Struct(
--- a/cairo/ethereum/cancun/trie.cairo
+++ b/cairo/ethereum/cancun/trie.cairo
@@ -951,7 +951,7 @@ func _get_branches{poseidon_ptr: PoseidonBuiltin*}(obj: MappingBytesBytes, level
         assert value = value_15;
         assert value_set = 1;
     }
-    %{ value_set_or_zero %}
+    %{ fp_plus_2_or_0 %}
     if (value_set != 1) {
         let (data: felt*) = alloc();
         tempvar empty_bytes = Bytes(new BytesStruct(data, 0));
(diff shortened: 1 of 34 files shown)


cairo/ethereum/cancun/trie.cairo
obatirou, line 1: A few hints are missing: 
* %{ initial_dict = {} %}
* %{ memory[fp] = oracle(ids, reference="ethereum.cancun.trie.common_prefix_length") %}
* local remainder = nondet %{ ids.x.value.len % 2 %};
* %{ memory[ap - 1] = oracle(ids, reference="ethereum.cancun.trie.bytes_to_nibble_list") %}
* %{
            obj_tracker = __dict_manager.get_tracker(ids.dict_ptr_stop.address_)
            dict_tracker = __dict_manager.get_tracker(ids.branch_ptr.address_)
            dict_tracker.current_ptr += ids.DictAccess.SIZE
            preimage = next(key for key in obj_tracker.data.keys() if poseidon_hash_many(key) == ids.dict_ptr.key.value)
            dict_tracker.data[preimage] = obj_tracker.data[preimage]
        %}
* %{ ids.value_set = memory.get(fp + 2) or 0 %}
* %{
        from starkware.cairo.lang.vm.crypto import poseidon_hash_many
        hashed_value = ids.dict_ptr.key.value
        dict_tracker = __dict_manager.get_tracker(ids.dict_ptr_stop)
        # Get the key in the dict that matches the hashed value
        preimage = bytes(next(key for key in dict_tracker.data.keys() if poseidon_hash_many(key) == hashed_value))
        segments.write_arg(ids.preimage_data, preimage)
        ids.preimage_len = len(preimage)
    %}

cairo/tests/ethereum/cancun/vm/test_gas.py
obatirou, line 1: Same in gas.cairo a few hint were not replaced 
* %{ memory[ap] = 0 if 0 <= (ids.a % PRIME) < range_check_builtin.bound else 1 %}
*  %{ memory[ap] = 0 if 0 <= ((-ids.a - 1) % PRIME) < range_check_builtin.bound else 1 %}

enitrat: whitelisted/default hints don't need to be replaced

crates/cairo-addons/src/vm/hint_definitions/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+        errors::{hint_errors::HintError, memory_errors::MemoryError},
+        vm_core::VirtualMachine,
+    },
+    Felt252,
+};
+
+use crate::vm::hints::Hint;
+
+#[allow(non_snake_case)]
+pub fn bytes__eq__() -> Hint {
```
obatirou, line 22: I think this could be refactored
The same thing is done for self and other and then they are compared and the same behavior is done if there is a diff or not.
Could be factorized with helper closures ?

crates/cairo-addons/src/vm/hint_definitions/utils.rs
```diff
@@ -0,0 +1,145 @@
...
+        String::from("b_le_a"),
+        |vm: &mut VirtualMachine,
+         _exec_scopes: &mut ExecutionScopes,
+         ids_data: &HashMap<String, HintReference>,
+         ap_tracking: &ApTracking,
+         _constants: &HashMap<String, Felt252>|
+         -> Result<(), HintError> {
+            let a = get_maybe_relocatable_from_var_name("a", vm, ids_data, ap_tracking)?;
+            let b = get_maybe_relocatable_from_var_name("b", vm, ids_data, ap_tracking)?;
+            let result = if b <= a { 1 } else { 0 };
```
obatirou, line 134:
```suggestion
            let result = usize::from(b <= a);
```

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
```diff
@@ -0,0 +1,121 @@
...
+            let dict_ptr = get_ptr_from_var_name("dict_ptr", vm, ids_data, ap_tracking)?;
+            let dict_manager_ref = exec_scopes.get_dict_manager()?;
+            let mut dict = dict_manager_ref.borrow_mut();
+            let tracker = dict.get_tracker_mut(dict_ptr)?;
+            tracker.current_ptr.offset += DICT_ACCESS_SIZE;
+
+            // Get key parameters
+            let key = get_maybe_relocatable_from_var_name("key", vm, ids_data, ap_tracking)?
+                .get_relocatable()
+                .ok_or_else(|| HintError::IdentifierNotRelocatable(Box::from("key")))?;
```
obatirou, line 45:
```suggestion
                .ok_or(HintError::IdentifierNotRelocatable(Box::from("key")))?;
```

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
```diff
@@ -0,0 +1,121 @@
...
+
+            let key_len: usize =
+                get_maybe_relocatable_from_var_name("key_len", vm, ids_data, ap_tracking)?
+                    .get_int()
+                    .ok_or_else(|| HintError::IdentifierNotInteger(Box::from("key_len")))?
```
obatirou, line 50:
```suggestion
                    .ok_or(HintError::IdentifierNotInteger(Box::from("key_len")))?
```

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
obatirou: Could also simplify by removing the Box and "key".into() maybe ?

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
```diff
@@ -0,0 +1,121 @@
...
+            let dict_ptr = get_ptr_from_var_name("dict_ptr", vm, ids_data, ap_tracking)?;
+            let dict_manager_ref = exec_scopes.get_dict_manager()?;
+            let mut dict = dict_manager_ref.borrow_mut();
+            let tracker = dict.get_tracker_mut(dict_ptr)?;
+            tracker.current_ptr.offset += DICT_ACCESS_SIZE;
+
+            // Get key parameters
+            let key = get_maybe_relocatable_from_var_name("key", vm, ids_data, ap_tracking)?
+                .get_relocatable()
+                .ok_or_else(|| HintError::IdentifierNotRelocatable(Box::from("key")))?;
```
obatirou, line 82:
```suggestion
                .ok_or(HintError::IdentifierNotRelocatable(Box::from("key")))?;
```

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
```diff
@@ -0,0 +1,121 @@
...
+
+            let key_len: usize =
+                get_maybe_relocatable_from_var_name("key_len", vm, ids_data, ap_tracking)?
+                    .get_int()
+                    .ok_or_else(|| HintError::IdentifierNotInteger(Box::from("key_len")))?
```
obatirou, line 87:
```suggestion
                    .ok_or(HintError::IdentifierNotInteger(Box::from("key_len")))?
```

crates/cairo-addons/src/vm/hint_definitions/utils.rs
enitrat: nice

crates/cairo-addons/src/vm/hint_definitions/utils.rs
enitrat: yes excellent idea!

cairo/ethereum/cancun/trie.cairo
enitrat: migrated all, but not oracle ones. might keep it for a future PR

cairo/ethereum/cancun/trie.cairo
```diff
@@ -963,7 +949,7 @@ func _get_branches{poseidon_ptr: PoseidonBuiltin*}(obj: MappingBytesBytes, level
         assert value = value_15;
         assert value_set = 1;
     }
-    %{ ids.value_set = memory.get(fp + 2) or 0 %}
+    %{ value_set_or_zero %}
```
ClementWalter, line 952: I'd advocate for a more detailed naming like `fp_plus_2_or_0`. We also probably would like to have a generic hint. Thinking that since it's basically just a string, we could encode more data in it to have cairo 1 like hint (with src and dst)

cairo/ethereum/cancun/trie.cairo
```diff
@@ -976,28 +962,22 @@ func _get_branches{poseidon_ptr: PoseidonBuiltin*}(obj: MappingBytesBytes, level
...
-    %{
-        from starkware.cairo.lang.vm.crypto import poseidon_hash_many
-        hashed_value = ids.dict_ptr.key.value
-        dict_tracker = __dict_manager.get_tracker(ids.dict_ptr_stop)
-        # Get the key in the dict that matches the hashed value
-        preimage = bytes(next(key for key in dict_tracker.data.keys() if poseidon_hash_many(key) == hashed_value))
-        segments.write_arg(ids.preimage_data, preimage)
-        ids.preimage_len = len(preimage)
-    %}
+    %{ get_preimage_for_key %}
```
ClementWalter, line 975: I really think that not only this refacto better for the devx, but it's also much cleaner when the hint content is not visible so that one directly catch what's in the program and what's not

image

enitrat: totally agree with that :)

cairo/tests/fixtures/runner.py
```diff
@@ -526,7 +526,9 @@ def _factory_rs(entrypoint, *args, **kwargs):
...
-    if request.node.get_closest_marker("python_vm"):
-        return _factory_py
-    else:
-        return _factory_rs
+    def _run(entrypoint, *args, vm=None, **kwargs):
+        if vm == "python":
+            return _factory_py(entrypoint, *args, **kwargs)
+        return _factory_rs(entrypoint, *args, **kwargs)
+
+    return _run
```
ClementWalter, line 534: issue with this is that the `RustProgram` may not be defined, see above

```
def rust_program(request, cairo_program: Program) -> RustProgram:
    if request.node.get_closest_marker("python_vm"):
        return None
```

if we want a per test flag, we rather just create two fixtures `cairo_run_py` and `cairo_run(_rs)` instead of the factory I introduced (since actually it's already two fixtures in one)

cairo/ethereum/cancun/trie.cairo
enitrat: yes, I was thinking of generic hint like this that would take a dynamic argument from the hint name. 

e.g. `fp_offset_or_0(offset: int)` and then use this as function param 🤔

crates/cairo-addons/Cargo.toml
```diff
@@ -20,6 +20,7 @@ pyo3 = { version = "0.22.4", features = [
 cairo-vm = { workspace = true }
 num-traits = "0.2.18"
 num-bigint = "0.4.6"
+starknet-crypto = "0.7.4"
```
ClementWalter, line 23: what for?

crates/cairo-addons/src/vm/hint_definitions.rs
```diff
@@ -0,0 +1,6 @@
+mod dict;
+pub use dict::HINTS as DICT_HINTS;
+mod hashdict;
+pub use hashdict::HINTS as HASHDICT_HINTS;
+mod utils;
+pub use utils::HINTS as UTILS_HINTS;
```
ClementWalter, line 6: don't know what clippy thinks nor how to enforce, but I prefer when we first have all the `mod`, then all the `(pub) use`

crates/cairo-addons/src/vm/hint_definitions/dict.rs
```diff
@@ -0,0 +1,69 @@
...
+         -> Result<(), HintError> {
+            //Check if there is a dict manager in scope, create it if there isnt one
+            let base = if let Ok(dict_manager) = exec_scopes.get_dict_manager() {
+                dict_manager.borrow_mut().new_dict(vm, Default::default())?
+            } else {
+                let mut dict_manager = DictManager::new();
+                let base = dict_manager.new_dict(vm, Default::default())?;
+                exec_scopes.insert_value("dict_manager", Rc::new(RefCell::new(dict_manager)));
+                base
+            };
```
ClementWalter, line 38: I really don't like this if or create for the dict manager in all dict related hints. I have put a dict manager from the beginning in the runner so I think that we can skip it

crates/cairo-addons/src/vm/hint_definitions/dict.rs
```diff
@@ -0,0 +1,69 @@
...
+    types::exec_scope::ExecutionScopes,
+    vm::{errors::hint_errors::HintError, vm_core::VirtualMachine},
+    Felt252,
+};
+
+use crate::vm::hints::Hint;
+
+pub const HINTS: &[fn() -> Hint] = &[dict_new_empty, copy_dict_segment];
+
+pub fn dict_new_empty() -> Hint {
```
ClementWalter, line 21: we could probably have a macro similar to the decorator in python to manage hint registration, but let's keep it like that for now

enitrat: I spent half-an-hour trying to make a macro, as an attribute `#[register_hint]` macro, but had some issues getting it to work properly. Thought this was an acceptable workaround

crates/cairo-addons/src/vm/hints.rs
```diff
@@ -44,22 +49,33 @@ impl Hint {
...
     }
 
     #[must_use]
-    pub fn with_hint(mut self, hint: &Hint) -> Self {
-        self.inner.add_hint(hint.id.clone(), hint.func.clone());
+    pub fn with_hints(mut self, hints: Vec<fn() -> Hint>) -> Self {
+        for fn_hint in hints {
+            let hint = fn_hint();
+            self.inner.add_hint(
+                self.python_hints.get(&hint.id).unwrap_or(&hint.id).to_string(),
```
ClementWalter, line 66: this is to handle the case where the compiled program has already the hints implemented for python right?

crates/cairo-addons/src/vm/mod.rs
ClementWalter, line 1: maybe time for a little reorg of the module. can be done in a follow up PR

crates/cairo-addons/src/vm/hint_definitions/dict.rs
enitrat: The logic here was to be as little differentiated from the original dict_new hint as possible

python/cairo-addons/src/cairo_addons/hints/dict.py
```diff
@@ -6,6 +6,22 @@
...
+    dict_manager: DictManager,
+    ids: VmConsts,
+    segments: MemorySegmentManager,
+    memory: MemoryDict,
+    ap: RelocatableValue,
+):
+    if "__dict_manager" not in globals():
+        from starkware.cairo.common.dict import DictManager
+
+        __dict_manager = DictManager()
```
ClementWalter, line 20: see comment about the dict manager, I have on purpose removed all these `if`

python/cairo-addons/src/cairo_addons/hints/hashdict.py
```diff
@@ -0,0 +1,81 @@
...
+    dict_manager: DictManager,
+    ids: VmConsts,
+    segments: MemorySegmentManager,
+    memory: MemoryDict,
+    ap: RelocatableValue,
+) -> int:
+    dict_tracker = dict_manager.get_tracker(ids.dict_ptr)
+    dict_tracker.current_ptr += ids.DictAccess.SIZE
+    preimage = tuple([memory[ids.key + i] for i in range(ids.key_len)])
+    # Not using [] here because it will register the value for that key in the tracker.
```
ClementWalter, line 20: I don't understand this comment

enitrat:
```
        # Not using [] here because it will register the value for that key in the tracker.
        ids.value = dict_tracker.data.get(preimage, dict_tracker.data.default_factory())
```

I noticed that doing 

dict_tracker.data[preimage] would either:
1. return the value if it exists
2. write the default value in the dict, and return it, it it doesn not exist.

We want to avoid writing default values to the dict. As such, doing dict.get(key, default_value) prevents this.

crates/cairo-addons/src/vm/hints.rs
enitrat: yes, exactly.

python/cairo-addons/src/cairo_addons/hints/hashdict.py
ClementWalter: ah ok I didn't even understand that `[]` applied to `data`. You can just give more context like "not using `__getitem__` on `data` dict because this would add the key in case `data` is a `defaultdict`

cairo/tests/ethereum/cancun/test_trie.py
```diff
@@ -149,56 +139,63 @@ def test_get_branches(self, cairo_run, obj, level):
...
-    ):
-        cairo_trie = cairo_run("trie_set_TrieBytes32U256", trie, key, value)
-        trie_set(trie, key, value)
-        assert cairo_trie == trie
+    class TestGet:
+        @given(trie=..., key=...)
+        def test_trie_get_TrieAddressAccount(
+            self, cairo_run, trie: Trie[Address, Optional[Account]], key: Address
+        ):
+            [trie_cairo, result_cairo] = cairo_run(
```
ClementWalter, line 152:
```suggestion
            trie_cairo, result_cairo = cairo_run(
```

I guess that you don't need the brackets here and in the following test cases

cairo/tests/ethereum/cancun/test_trie.py
```diff
@@ -103,11 +88,14 @@ def test_nibble_list_to_compact_should_raise_when_wrong_remainder(
             ),
             cairo_error(message="nibble_list_to_compact: invalid remainder"),
         ):
-            cairo_run("nibble_list_to_compact", x, is_leaf)
+            # Always run patch_hint tests with the python VM
```
ClementWalter, line 91: we need to think about a solution for this for ok for now

python/cairo-addons/src/cairo_addons/hints/hashdict.py
ClementWalter, line 1: note: one could put all the imports at the module level and no imports in the hints because the whole module is loaded so they would be available at runtime, but I think that it's also good to keep the module imports at the minimum required to define a hint

ClementWalter (approved):
ok to merge but I'd like to better organize rust VS python hints, probably in their own crates/workspace package

PR #414, opened 2025-01-14
feat: set storage

### LINKED ISSUE #409: set_storage

Eikix:
Setting to draft as tests cannot pass until we skip serialization of nested dictionaries that are filled with default_values in serde.py, as seen in https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/state.py#L318

closes #409

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -161,7 +161,7 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
     let value = trie_get_TrieBytes32U256{poseidon_ptr=poseidon_ptr, trie=storage_trie}(key);
 
     // Rebind the storage trie to the state
-    let new_storage_trie_ptr = cast(storage_trie_ptr, felt);
+    let new_storage_trie_ptr = cast(storage_trie.value, felt);
 
     hashdict_write{poseidon_ptr=poseidon_ptr, dict_ptr=storage_tries_dict_ptr}(
         1, &address.value, new_storage_trie_ptr
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -218,6 +218,8 @@ func set_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
     }(1, &address.value);
 
     if (cast(storage_trie_pointer, felt) == 0) {
+        // dict_new expects an initial_dict hint argument.
+        %{ initial_dict = {} %}
         let (new_mapping_dict_ptr) = dict_new();
         tempvar _storage_trie = new TrieBytes32U256Struct(
             secured=bool(1),
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -203,6 +203,8 @@ func set_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
     // Assert that the account exists
     let account = get_account_optional(address);
     if (cast(account.value, felt) == 0) {
+        // TODO: think about which cases lead to this error and decide on the correct type of exception to raise
+        // perhaps AssertionError
         with_attr error_message("Cannot set storage on non-existent account") {
             assert 0 = 1;
         }
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -25,7 +25,7 @@ from ethereum.cancun.trie import (
     TrieBytes32U256Struct,
 )
 from ethereum_types.bytes import Bytes, Bytes32
-from src.utils.dict import hashdict_read, hashdict_write
+from src.utils.dict import hashdict_read, hashdict_write, hashdict_get
 from ethereum_types.numeric import U256, U256Struct
 
 struct AddressTrieBytes32U256DictAccess {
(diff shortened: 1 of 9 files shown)


cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+    address: Address, key: Bytes32, value: U256
+) {
+    alloc_locals;
+    let fp_and_pc = get_fp_and_pc();
+    local __fp__: felt* = fp_and_pc.fp_val;
+
+    // Assert that the account exists
+    let account = get_account_optional(address);
+    if (cast(account.value, felt) == 0) {
+        with_attr error_message("Cannot set storage on non-existent account") {
```
enitrat, line 202: Not sure if this should be an Exception of type AssertionError instead. I need to think more about it

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+                    dict_ptr=cast(new_mapping_dict_ptr, Bytes32U256DictAccess*),
+                    original_mapping=cast(0, MappingBytes32U256Struct*),
+                ),
+            ),
+        );
+
+        let new_storage_trie_ptr = cast(_storage_trie, felt);
+        hashdict_write{poseidon_ptr=poseidon_ptr, dict_ptr=storage_tries_dict_ptr}(
+            1, &address.value, new_storage_trie_ptr
+        );
```
enitrat, line 230: I personnally prefer 

```suggestion
    with poseidon_ptr, dict_ptr{
        hashdict_write(
            1, &address.value, new_storage_trie_ptr
        );
    }
```

 However it allows you to not define dict_ptr=xxx explicitly. Which is also nice. Perhaps we should stick to one style only.

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+
+        let storage_trie = TrieBytes32U256(_storage_trie);
```
enitrat, line 232: unused

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+        tempvar state = State(
+            new StateStruct(
+                _main_trie=state.value._main_trie,
+                _storage_tries=storage_tries,
+                _snapshots=state.value._snapshots,
+                created_accounts=state.value.created_accounts,
+            ),
+        );
+        return ();
+    } else {
```
enitrat, line 254: no else after return

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+        let trie_struct = cast(storage_trie_pointer, TrieBytes32U256Struct*);
+        let storage_trie = TrieBytes32U256(trie_struct);
+        trie_set_TrieBytes32U256{poseidon_ptr=poseidon_ptr, trie=storage_trie}(key, value);
+
+        // From EELS <https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/state.py#L318>:
+        // if trie._data == {}:
```
enitrat, line 260: a shortcut we can take is that trie._data can only be {} if value == default_value.to keep in mind

cairo/ethereum/cancun/state.cairo
Eikix: should I add todo?

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +188,106 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+    let storage_tries_dict_ptr = cast(state.value._storage_tries.value.dict_ptr, DictAccess*);
+    let (storage_trie_pointer) = hashdict_read{
+        poseidon_ptr=poseidon_ptr, dict_ptr=storage_tries_dict_ptr
+    }(1, &address.value);
+
+    // If the storage trie is not found, create an empty one and store it in the state
+    if (cast(storage_trie_pointer, felt) == 0) {
+        let (new_mapping_dict_ptr) = dict_new();
+        tempvar _storage_trie = new TrieBytes32U256Struct(
+            secured=bool(1),
```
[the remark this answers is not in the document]
Eikix: good question

cairo/ethereum/cancun/state.cairo
Eikix: i prefer the following:
- if it's only one line, then I use implicit argument in the function call
- if i need multiple lines, then using `with` is nice

cairo/ethereum/cancun/state.cairo
enitrat: think about which cases lead to this, yes

obatirou (dismissed):

cairo/ethereum/cancun/state.cairo
```diff
@@ -181,3 +190,109 @@ func get_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+
+    let storage_tries_dict_ptr = cast(storage_tries.value.dict_ptr, DictAccess*);
+    // Use `hashdict_get` instead of `hashdict_read` because `MappingAddressTrieBytes32U256` is not a
+    // `default_dict`. Accessing a key that does not exist in the dict would have panicked for `hashdict_read`.
+    let (storage_trie_pointer) = hashdict_get{
+        poseidon_ptr=poseidon_ptr, dict_ptr=storage_tries_dict_ptr
+    }(1, &address.value);
+
+    if (cast(storage_trie_pointer, felt) == 0) {
+        let (new_mapping_dict_ptr) = dict_new();
```
obatirou, line 221: Does not it needs a %{initial_dict = {} %} ?
https://github.com/starkware-libs/cairo-lang/blob/8e11b8cc65ae1d0959328b1b4a40b92df8b58595/src/starkware/cairo/common/dict.cairo#L7

PR #437, opened 2025-01-15
feat: set_account

Eikix:
tests are broken so far, will investigate


enitrat:
@Eikix should be good now

enitrat (approved):

PR #444, opened 2025-01-16
feat: get_storage_original

### LINKED ISSUE #440: get_storage_original

enitrat:
Closes #440 

Note: introduces a new field in the state that holds the original storage tries. 
Because in sstore, we get the original storage, meaning we need to access the snapshots[0] on each sstore. 
Because a dict_read writes memory data, it would mean we have to override snapshots[0] on each dict_read, which is inconvenient. 
This way, we can only overwrite the `original_storage_tries` field of the State struct.

### DIFF
--- a/cairo/tests/utils/args_gen.py
+++ b/cairo/tests/utils/args_gen.py
@@ -615,6 +615,13 @@ def _gen_arg(
             )
             for f in fields(arg_type_origin)
         ]
+
+        if arg_type_origin is State:
+            # In case of a Trie, we need to put the original_storage_tries (trie._snapshots[0]) in a
+            # special field of the State. We want easy access / overrides to this specific snapshot in
+            # Cairo, as each `sstore` performs an update of this original trie.
+            data.append(data[1])
+
         segments.load_data(struct_ptr, data)
 
         if arg_type_origin is Trie:
--- a/cairo/tests/utils/args_gen.py
+++ b/cairo/tests/utils/args_gen.py
@@ -617,7 +617,7 @@ def _gen_arg(
         ]
 
         if arg_type_origin is State:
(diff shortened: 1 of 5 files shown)


cairo/tests/utils/args_gen.py
```diff
@@ -615,6 +615,13 @@ def _gen_arg(
             )
             for f in fields(arg_type_origin)
         ]
+
+        if arg_type_origin is State:
+            # In case of a Trie, we need to put the original_storage_tries (trie._snapshots[0]) in a
```
Eikix, line 620:
```suggestion
            # In case of a Trie, we need to put the original_storage_tries (state._snapshots[0][1]) in a
```

cairo/tests/utils/strategies.py
```diff
@@ -377,6 +378,24 @@ def tuple_strategy(thing):
...
+            _snapshots=[
+                (
+                    copy_trie(state._main_trie),
+                    {
+                        addr: copy_trie(trie)
+                        for addr, trie in state._storage_tries.items()
+                    },
+                )
+            ],
+            created_accounts=set(),
```
obatirou, line 397: created_accounts will always be empty 
Do we pass the condition at some point ?
```
    if (is_created == 0) {
        tempvar res = U256(new U256Struct(0, 0));
        return res;
    }
```

Eikix: as execution advances, this set will grow

Eikix: alternatively, @enitrat 's test can call `mark_account_created`

enitrat: modified the strategy.

obatirou (approved):

PR #450, opened 2025-01-16
feat: begin transaction

### LINKED ISSUE #415: begin_transaction
https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/state.py#L75
```[tasklist]
### Tasks
- [ ] use trie_copy on _main_trie
- [ ] iterate over the keys of the _storage_tries (mapping Address -> Trie[Bytes32, U256]) and trie_copy on each returned trie
- [ ] store the above two in the snapshots
- [ ] repeat for transient storage
```

obatirou:
Close #415

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -791,9 +791,7 @@ func begin_transaction{
     );
 
     // Update the snapshots list
-    let current_snapshot = state.value._snapshots.value.data + state.value._snapshots.value.len;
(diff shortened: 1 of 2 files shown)


enitrat (dismissed):

enitrat:
otherwise looks good for me, I believe we could keep going with this explicit snapshot system for the time being.
Considering that the the snapshot system is isolated to:
- commit
- rollback
- begin_transaction

doing it the "naive" way allows us to keep working in parallel on the codebase in the immediate future

enitrat (approved):

PR #474, opened 2025-01-17
tests: stricter error assertions

### LINKED ISSUE #438: create custom contextmanager for error assertions
The current error assertion framework is too relaxed, and will consider that if Cairo throws Exception and Python throws a non-EthereumException error, the condition passes.

We should have more strict comparison: e.g.

```
    @given(data=state_and_address_and_optional_key(key_strategy=bytes32), value=...)
    def test_set_storage(self, cairo_run, data, value: U256):
        state, address, key = data
        try:
            state_cairo = cairo_run("set_storage", state, address, key, value)
        except Exception as e:
            with assert_raises_exactly(
                type(e)
            ):
                set_storage(state, address, key, value)
            return

        set_storage(state, address, key, value)
        assert state_cairo == state

from contextlib import contextmanager
from typing import Type

@contextmanager
def assert_raises_exactly(expected_exception: Type[Exception], msg: str = None):
    """
    Context manager that verifies an exception of exactly the specified type is raised.
    Unlike pytest.raises, this doesn't allow subclass exceptions to match.

    Args:
        expected_exception: The exact exception type expected
        msg: Optional message to include in assertion error

    Example:
        with assert_raises_exactly(AssertionError):
            raise AssertionError()  # passes

        with assert_raises_exactly(Exception):
            raise AssertionError()  # fails - more specific exception
    """
    try:
        yield
    except Exception as e:
        if type(e) is not expected_exception:
            raise AssertionError(
                msg or f"Expected exactly {expected_exception.__name__}, but got {type(e).__name__}"
            )
    else:
        raise AssertionError(
            msg or f"Expected {expected_exception.__name__} to be raised, but no exception was raised"
        )

```

For that specific test, we have a `assert X = Y` instruction that could fail in cairo if an account's storage trie is None. This test used to (wrongly) pass because the runner would throw a generic exception.

We should specific the runner to throw AssertionError if it fails on an ASSERT_EQ: 

```
        try:
            runner.run_until_pc(end, run_resources)
        except Exception as e:
            if "An ASSERT_EQ instruction failed" in str(e):
                raise AssertionError(e) from e
            raise Exception(str(e)) from e
```

enitrat:
Closes #438
use a special contextmanager that asserts the error raised is EXACTLY of the expected type.

### DIFF
--- a/cairo/tests/ethereum/cancun/vm/instructions/test_arithmetic.py
+++ b/cairo/tests/ethereum/cancun/vm/instructions/test_arithmetic.py
@@ -1,4 +1,3 @@
-import pytest
 from hypothesis import given
 
 from ethereum.cancun.vm.instructions.arithmetic import (
@@ -15,6 +14,7 @@ from ethereum.cancun.vm.instructions.arithmetic import (
     sub,
 )
 from tests.utils.args_gen import Evm
+from tests.utils.errors import assert_raises_exactly
 from tests.utils.evm_builder import EvmBuilder
 
 arithmetic_tests_strategy = EvmBuilder().with_stack().with_gas_left().build()
@@ -26,7 +26,7 @@ class TestArithmetic:
         try:
             cairo_result = cairo_run("add", evm)
         except Exception as cairo_error:
-            with pytest.raises(type(cairo_error)):
+            with assert_raises_exactly(type(cairo_error)):
                 add(evm)
             return
 
@@ -38,7 +38,7 @@ class TestArithmetic:
         try:
             cairo_result = cairo_run("sub", evm)
         except Exception as cairo_error:
-            with pytest.raises(type(cairo_error)):
+            with assert_raises_exactly(type(cairo_error)):
                 sub(evm)
             return
 
@@ -50,7 +50,7 @@ class TestArithmetic:
         try:
             cairo_result = cairo_run("mul", evm)
         except Exception as cairo_error:
-            with pytest.raises(type(cairo_error)):
+            with assert_raises_exactly(type(cairo_error)):
                 mul(evm)
             return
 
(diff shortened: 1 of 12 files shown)


cairo/tests/ethereum/cancun/vm/instructions/test_arithmetic.py
```diff
@@ -38,7 +38,7 @@ def test_sub(self, cairo_run, evm: Evm):
         try:
             cairo_result = cairo_run("sub", evm)
         except Exception as cairo_error:
-            with pytest.raises(type(cairo_error)):
+            with assert_raises_exactly(type(cairo_error)):
```
Eikix, line 41: why doesn't pytest.raises assert 'strictly' on the `type(cairo_error)`?

enitrat: i dont know what to answer you but approve me

Eikix: maybe there's an easier wrapper? nit at this point but worth a look
image

Eikix: https://docs.pytest.org/en/stable/reference/reference.html#pytest.raises

Eikix: The context manager produces an [ExceptionInfo](https://docs.pytest.org/en/stable/reference/reference.html#pytest.ExceptionInfo) object which can be used to inspect the details of the captured exception:

enitrat: best I can do is a wrapper around pytest's raise, which clearly says

obatirou (dismissed):

Eikix:
conflicts

obatirou (approved):

PR #476, opened 2025-01-17
Feat/move ether

### LINKED ISSUE #461: move_ether

Eikix:
- move ether
- add utils to add and sub with panics for overflow/underflow for U256 type
- add state deep copy util in test_state 

close #461


cairo/ethereum/cancun/state.cairo
```diff
@@ -184,6 +185,30 @@ func set_account{poseidon_ptr: PoseidonBuiltin*, state: State}(
     return ();
 }
 
+func move_ether{range_check_ptr, poseidon_ptr: PoseidonBuiltin*, state: State}(
+    sender_address: Address, recipient_address: Address, amount: U256
+) {
+    alloc_locals;
+    let fp_and_pc = get_fp_and_pc();
+    local __fp__: felt* = fp_and_pc.fp_val;
```
obatirou, line 193: is this necessary ?

cairo/tests/ethereum/cancun/test_state.py
```diff
@@ -116,6 +136,36 @@ def test_set_account(self, cairo_run, data, account: Optional[Account]):
...
+    @given(
+        data=state_and_address_and_optional_key(), recipient_address=address, amount=...
+    )
+    def test_move_ether(
+        self, cairo_run, data, recipient_address: Address, amount: U256
+    ):
+        state, sender_address = data
+        # We need to create a deep copy of the state to avoid mutating the original state
+        # We can refactor it into a deep_copy State util if we ever need to do this again
+        python_state = _state_deep_copy(state)
```
obatirou, line 148: I do not understand why doing this
To avoid having doing a copy you first execute the cairo code, then the python one like other tests, unless I miss something?
Also if there is any need you can use https://docs.python.org/3/library/copy.html to copy objects if not mistaken

cairo/tests/ethereum/cancun/test_state.py
```diff
@@ -116,6 +136,36 @@ def test_set_account(self, cairo_run, data, account: Optional[Account]):
...
+                cairo_run(
+                    "move_ether", state, sender_address, recipient_address, amount
+                )
+            return
+        except OverflowError:
+            with cairo_error("OverflowError"):
+                cairo_run(
+                    "move_ether", state, sender_address, recipient_address, amount
+                )
+            return
```
obatirou, line 163: I think the PR on error assertion from @enitrat will help with this

cairo/tests/ethereum/cancun/test_state.py
enitrat: yes, run the cairo operation before the python one.

enitrat (changes requested):
try with https://github.com/kkrt-labs/keth/pull/474 now

cairo/tests/ethereum/cancun/test_state.py
Eikix: previously if i ran the cairo operation before the python one, then i'm not able to know which error will be thrown

Eikix: > I do not understand why doing this To avoid having doing a copy you first execute the cairo code, then the python one like other tests, unless I miss something? Also if there is any need you can use https://docs.python.org/3/library/copy.html to copy objects if not mistaken

you can't copy state because Account is a frozen type,  at least that's the error i was getting when trying to just use `deepcopy` from `copy` library

Eikix:
> try with #474 now

thanks that's way better

cairo/ethereum/cancun/state.cairo
Eikix: no thx nice catch

cairo/ethereum/utils/numeric.cairo
```diff
@@ -148,3 +150,36 @@ func U256__eq__(a: U256, b: U256) -> bool {
...
+            assert 0 = 1;
+        }
+    }
+    let (result) = uint256_sub(cast([a.value], Uint256), cast([b.value], Uint256));
+    tempvar res = U256(new U256Struct(result.low, result.high));
+    return res;
+}
+
+func U256_le{range_check_ptr}(a: U256, b: U256) -> bool {
+    let (result) = uint256_le(cast([a.value], Uint256), cast([b.value], Uint256));
```
obatirou, line 182:
```suggestion
    let (result) = uint256_le([a.value], [b.value]);
```
You should not have to cast: a.value is a Uint256*

(to apply in all functions above)

Eikix: I thought it'd be clearer that we're using U256Struct as a Uint256 here by casting it, even though U256Struct is just a type alias for Uint256

cairo/ethereum/cancun/state.cairo
```diff
@@ -184,6 +185,27 @@ func set_account{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+    sender_address: Address, recipient_address: Address, amount: U256
+) {
+    alloc_locals;
+    let sender_account = get_account(sender_address);
+    let recipient_account = get_account(recipient_address);
+    let sender_balance = sender_account.value.balance;
+
+    let is_sender_balance_sufficient = U256_le(amount, sender_balance);
+    if (is_sender_balance_sufficient.value == 0) {
+        assert 0 = 1;
```
obatirou, line 197: should raise an AssertionError from exceptions.cairo

cairo/ethereum/utils/numeric.cairo
```diff
@@ -148,3 +150,36 @@ func U256__eq__(a: U256, b: U256) -> bool {
...
+
+// @dev Panics if overflow
+func U256_add{range_check_ptr}(a: U256, b: U256) -> U256 {
+    alloc_locals;
+    let (res, carry) = uint256_add([a.value], [b.value]);
+    if (carry != 0) {
+        with_attr error_message("OverflowError") {
+            assert 0 = 1;
+        }
+    }
```
ClementWalter, line 162:
```suggestion
    with_attr error_message("OverflowError") {
        assert 0 = carry;
    }
```

cairo/ethereum/utils/numeric.cairo
```diff
@@ -148,3 +150,36 @@ func U256__eq__(a: U256, b: U256) -> bool {
...
+
+// @dev Panics if underflow with OverflowError
+func U256_sub{range_check_ptr}(a: U256, b: U256) -> U256 {
+    alloc_locals;
+    let is_within_bounds = U256_le(b, a);
+    if (is_within_bounds.value == 0) {
+        with_attr error_message("OverflowError") {
+            assert 0 = 1;
+        }
+    }
```
ClementWalter, line 175: remove if, cf above

cairo/ethereum/cancun/state.cairo
```diff
@@ -184,6 +185,27 @@ func set_account{poseidon_ptr: PoseidonBuiltin*, state: State}(
...
+    sender_address: Address, recipient_address: Address, amount: U256
+) {
+    alloc_locals;
+    let sender_account = get_account(sender_address);
+    let sender_balance = sender_account.value.balance;
+
+    let is_sender_balance_sufficient = U256_le(amount, sender_balance);
+    if (is_sender_balance_sufficient.value == 0) {
+        assert 0 = 1;
+    }
```
obatirou, line 198: same here

cairo/tests/ethereum/utils/test_numeric.py
```diff
@@ -81,3 +82,30 @@ def test_U256__eq__(self, cairo_run, a: U256, b: U256):
...
+    def test_U256_le(self, cairo_run, a: U256, b: U256):
+        assert (a <= b) == cairo_run("U256_le", a, b)
+
+    @given(a=..., b=...)
+    def test_U256_add(self, cairo_run, a: U256, b: U256):
+        try:
+            a + b
+        except Exception as e:
+            with cairo_error(str(e.__class__.__name__)):
+                cairo_result = cairo_run("U256_add", a, b)
```
enitrat, line 96:
```suggestion
            with strict_raises(type(e)):
                cairo_result = cairo_run("U256_add", a, b)
```

and in runner.py add a mechanism to throw the correct error based on the string: "Error message: OverflowError"

cairo/tests/ethereum/utils/test_numeric.py
```diff
@@ -81,3 +82,30 @@ def test_U256__eq__(self, cairo_run, a: U256, b: U256):
...
+    def test_U256_sub(self, cairo_run, a: U256, b: U256):
+        try:
+            a - b
+        except Exception as e:
+            with cairo_error(str(e.__class__.__name__)):
+                cairo_result = cairo_run("U256_sub", a, b)
+            return
+
+        cairo_result = cairo_run("U256_sub", a, b)
+        assert cairo_result == a - b
```
obatirou, line 111: could you explain why those tests are architectured differently?
Why execute python then cairo and why not using strict_raises

cairo/ethereum/utils/numeric.cairo
```diff
@@ -165,3 +166,34 @@ func U256_from_be_bytes20{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(bytes20
...
+    let (result) = uint256_sub([a.value], [b.value]);
+    tempvar res = U256(new U256Struct(result.low, result.high));
+    return res;
+}
+
+func U256_le{range_check_ptr}(a: U256, b: U256) -> bool {
+    let (result) = uint256_le([a.value], [b.value]);
+    tempvar res = bool(result);
+    return res;
+}
```
obatirou, line 199: nit: could be inlined ?
it seems not necessary to put in a function

Eikix: inlined where? in all the places we'll use it?

PR #489, opened 2025-01-17
feat: blob_hash

### LINKED ISSUE #479: `blob_hash`

obatirou:
Close #479

### DIFF
--- a/cairo/ethereum/cancun/vm/instructions/environment.cairo
+++ b/cairo/ethereum/cancun/vm/instructions/environment.cairo
@@ -2,6 +2,7 @@ from starkware.cairo.common.cairo_builtins import BitwiseBuiltin, PoseidonBuilti
 from starkware.cairo.common.registers import get_fp_and_pc
 from starkware.cairo.common.dict_access import DictAccess
 from starkware.cairo.common.math import split_felt
+from starkware.cairo.common.uint256 import uint256_lt, Uint256
 
 from ethereum_types.bytes import Bytes32, Bytes32Struct
 from ethereum_types.numeric import U256, U256Struct, Uint, UnionUintU256, UnionUintU256Enum
@@ -13,7 +14,7 @@ from ethereum.cancun.vm.stack import Stack, push, pop
 from ethereum.cancun.state import get_account
 from ethereum.cancun.utils.address import to_address
 
-from ethereum.utils.numeric import U256_lt, U256_from_be_bytes, U256_from_be_bytes20
+from ethereum.utils.numeric import U256_from_be_bytes, U256_from_be_bytes20
 
 from src.utils.bytes import felt_to_bytes20_little
 from src.utils.dict import hashdict_read, hashdict_write
@@ -352,8 +353,8 @@ func blob_hash{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, evm: Evm}() -> Exc
     // If index is within bounds, get the hash at that index
     // Otherwise return zero bytes
     let (high, low) = split_felt(blob_hashes.value.len);
-    let in_bounds = U256_lt(index, U256(new U256Struct(low, high)));
-    if (in_bounds.value == 0) {
+    let (in_bounds) = uint256_lt([index.value], Uint256(low, high));
+    if (in_bounds == 0) {
         tempvar blob_hash = Bytes32(new Bytes32Struct(0, 0));
     } else {
         tempvar blob_hash = blob_hashes.value.data[index.value.low];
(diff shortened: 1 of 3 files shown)


cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -0,0 +1,59 @@
...
+    if (cast(err, felt) != 0) {
+        return err;
+    }
+
+    // OPERATION
+    let blob_hashes = evm.value.env.value.blob_versioned_hashes;
+
+    // If index is within bounds, get the hash at that index
+    // Otherwise return zero bytes
+    let (high, low) = split_felt(blob_hashes.value.len);
```
Eikix, line 38: i guess it's fine to do this because we don't want to over analyze the evm

but blob_hashes.value.len right now is between 2 and 5, in the future it'll be at most 200 or maybe 1000

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -0,0 +1,59 @@
...
+    let out_of_bounds = U256_lt(index, U256(new U256Struct(low, high)));
+    if (out_of_bounds.value == 0) {
+        tempvar blob_hash = Bytes32(new Bytes32Struct(0, 0));
+    } else {
+        tempvar blob_hash = blob_hashes.value.data[index.value.low];
```
Eikix, line 43: if we're only using index.value.low, we should probably assert index.value.high == 0, as well as blob_hashes.value.len.high ==0,
then just compare the two low felt values

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -0,0 +1,59 @@
...
+        return err;
+    }
+
+    // OPERATION
+    let blob_hashes = evm.value.env.value.blob_versioned_hashes;
+
+    // If index is within bounds, get the hash at that index
+    // Otherwise return zero bytes
+    let (high, low) = split_felt(blob_hashes.value.len);
+    let out_of_bounds = U256_lt(index, U256(new U256Struct(low, high)));
```
enitrat, line 39: in_bounds ?

```suggestion
    let in_bounds = U256_lt(index, U256(new U256Struct(low, high)));
```

cairo/ethereum/cancun/vm/instructions/environment.cairo
enitrat: > we should probably assert index.value.high == 0,

no, this would be a deviation to spec. MAX_U256 would be a valid blob_hash index query and would return zero. We must not assume anything on the validity range of this input value.

for len(blob_hashes), we can assume that this number is way below 2**128.

so either:
1. just do u256 comparison, which is done now and is correct
2. return zero if input >= 2**128 or if input < len

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -0,0 +1,59 @@
...
+        return err;
+    }
+
+    // OPERATION
+    let blob_hashes = evm.value.env.value.blob_versioned_hashes;
+
+    // If index is within bounds, get the hash at that index
+    // Otherwise return zero bytes
+    let (high, low) = split_felt(blob_hashes.value.len);
+    let in_bounds = U256_lt(index, U256(new U256Struct(low, high)));
```
enitrat, line 39: considering that it's just a wrapper around uint256_lt i believe we could just do like

```suggestion
    let in_bounds = uint256_lt(index, U256Struct(low, high));
```

and avoid useless function calls

obatirou: [549271e](https://github.com/kkrt-labs/keth/pull/489/commits/549271ec0727a269a9fea1908c5381b832b32edf)

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -324,3 +328,49 @@ func base_fee{range_check_ptr, evm: Evm}() -> ExceptionalHalt* {
...
+    if (cast(err, felt) != 0) {
+        return err;
+    }
+
+    let blob_hashes = evm.value.env.value.blob_versioned_hashes;
+
+    // If index is within bounds, get the hash at that index
+    // Otherwise return zero bytes
+    let (high, low) = split_felt(blob_hashes.value.len);
+    let (in_bounds) = uint256_lt([index.value], Uint256(low, high));
```
enitrat, line 356:
```suggestion
    let (in_bounds) = uint256_lt([index.value], U256Struct(low, high));
```

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -1,7 +1,10 @@
 from starkware.cairo.common.cairo_builtins import BitwiseBuiltin, PoseidonBuiltin
 from starkware.cairo.common.registers import get_fp_and_pc
 from starkware.cairo.common.dict_access import DictAccess
+from starkware.cairo.common.math import split_felt
+from starkware.cairo.common.uint256 import uint256_lt, Uint256
```
enitrat, line 5:
```suggestion
from starkware.cairo.common.uint256 import uint256_lt
```

enitrat (approved):

PR #496, opened 2025-01-20
feat: dict_update

### LINKED ISSUE #403: hashdict_finalize
## Why

We need two behaviors
- rollback to a previous state of the dict with soundness (not only memcpy a portion of a segment)
- integrate values otherwise with soundness

The original dict_squash method only verifies that a DictAccess segment is sound.

## What

Add a function `hashdict_finalize`  in `dict.cairo` to handle these two behaviors.

## How

- [ ] take a `merge: felt` argument to define the target behavior
- [ ] if `merge=FALSE`, then `dict_squash` and create a new dict segment of `k, p, p` created from the squashed dict
- [ ] if `merge=TRUE`, then do nothing
- [ ] if `original_mapping` exists, append the above mentioned segment (newly or existing) to the original segment
- [ ] if not, then return the above mentioned segment
- [ ] remove the previous `test_dict.cairo` file
- [ ] returns dict_start, dict
- [ ] for the test, use a hint in `test_dict.cairo` to load the initial data into a `DictAccess*`

```
@given(...)
def test_hashdict_finalize(original_mapping: Dict[int, int], merge: bool):
    cairo_run_py("test_hashdict_finalize", ...)
```

in cairo
- dict_copy
- in cairo hint, append in `current_mapping` values to simulate writes
- call function and return a dict so that it's serde automatically
- assert in python `values = values + int(merge)`

enitrat:
Closes #403


cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
     let dict_ptr = dict_ptr + DictAccess.SIZE;
     return ();
 }
+
+func hashdict_finalize{range_check_ptr}(
+    dict_ptr_start: DictAccess*,
+    dict_ptr: DictAccess*,
+    original_mapping_start: DictAccess*,
+    original_mapping_end: DictAccess*,
+    should_merge: felt,
```
ClementWalter, line 146: I'd put `merge` only

cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
...
+        let (prev_values_start, prev_values_end) = prev_values(
+            squashed_dict_start, squashed_dict_end
+        );
+        tempvar dict_ptr_start = prev_values_start;
+        tempvar dict_ptr = prev_values_end;
+        tempvar range_check_ptr = range_check_ptr;
+    } else {
+        tempvar dict_ptr_start = dict_ptr_start;
+        tempvar dict_ptr = dict_ptr;
+        tempvar range_check_ptr = range_check_ptr;
```
ClementWalter, line 162: reorganize the range_check and args to avoid doing useless `tempvar` when calling `prev_values`

```suggestion
    } else {
        tempvar range_check_ptr = range_check_ptr;
        tempvar dict_ptr_start = dict_ptr_start;
        tempvar dict_ptr = dict_ptr;
```

cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
...
+
+    if (cast(original_mapping_end, felt) == 0) {
+        // No parent mapping, just return the current dict.
+        return (dict_ptr_start, dict_ptr);
+    }
+
+    let segment_to_merge = dict_ptr_start;
+    let len_to_merge = dict_ptr - dict_ptr_start;
+    let dst_ptr = original_mapping_end;
+    memcpy(dst_ptr, segment_to_merge, len_to_merge);
```
ClementWalter, line 176: tbh I find it more explicit than doing all these intermediate namings

```suggestion
    memcpy(original_mapping_end, dict_ptr_start, dict_ptr - dict_ptr_start);
```

ClementWalter: you can `tempvar len = dict_ptr - dict_ptr_start` to reuse it in both the `memcpy` and the `return`

cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
...
+    tempvar dict_ptr = dict_ptr_start;
+
+    loop:
+    let prev_values = cast([ap - 2], DictAccess*);
+    let dict_ptr = cast([ap - 1], DictAccess*);
+
+    let is_done = is_zero(dict_ptr_stop - dict_ptr);
+
+    tempvar prev_values_end = prev_values;
+    jmp end if is_done != 0;
```
ClementWalter, line 203: this will do one useless `tempvar` per loop, do an early return instead if there is nothing to copy, and check the condition at the end. You can also do the control flow in a hint, and just assert outside of the `loop` that `dict_ptr = dict_ptr_stop`

cairo/tests/src/utils/test_dict.cairo
```diff
@@ -1,50 +1,74 @@
 %builtins range_check
 
+from ethereum.cancun.fork_types import (
+    MappingBytes32U256,
+    MappingBytes32U256Struct,
+    Bytes32U256DictAccess,
+)
```
ClementWalter, line 7: unused

cairo/tests/src/utils/test_dict.cairo
```diff
@@ -1,50 +1,74 @@
 %builtins range_check
```
ClementWalter, line 1: can remove this

cairo/tests/src/utils/test_dict.py
```diff
@@ -1,11 +1,32 @@
...
 pytestmark = pytest.mark.python_vm
 
 
-@pytest.mark.parametrize(
-    "test_case",
-    ["test__dict_copy__should_return_copied_dict"],
-)
-def test_dict(cairo_run, test_case):
-    cairo_run(test_case)
+@given(dict_entries=st.lists(st.tuples(felt, felt, felt)))
```
ClementWalter, line 13:
```suggestion
@given(dict_entries=...)
```

isn't enough?

enitrat: I specifically need to generate integers in the valid felt range

cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
     let dict_ptr = dict_ptr + DictAccess.SIZE;
     return ();
 }
+
+func hashdict_finalize{range_check_ptr}(
```
ClementWalter, line 141: maybe add some comments to explain the behaviors of `merge=FALSE` and `TRUE`

ClementWalter: eventually, I think that this function should be called `dict_update`, similarly to python `dict.update()` because it's what it does: updating the value of dict 1 (original_mapping) here with the values of dict 2 (dict_ptr here). For soundness, we need to add a `drop: bool`

cairo/src/utils/dict.cairo
```diff
@@ -136,3 +137,83 @@ func hashdict_write{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
...
+    }
+
+    tempvar len = dict_ptr - dict_ptr_start;
+    memcpy(original_mapping_end, dict_ptr_start, len);
+    let new_original_mapping_end = original_mapping_end + len;
+    return (original_mapping_start, new_original_mapping_end);
+}
+
+// @notice Given a dict segment (start and end), returns a new dict segment with (key, prev_value, prev_value) for each key.
+// @dev Expectes the given dict to be squashed, with one DictAccess instance per key.
```
ClementWalter, line 177: this doesn't break the function, but just create useless entries

cairo/tests/src/utils/test_dict.cairo
```diff
@@ -1,50 +1,68 @@
-%builtins range_check
+from ethereum.cancun.fork_types import MappingBytes32U256
```
ClementWalter, line 1:
```suggestion
```

cairo/src/utils/dict.cairo
enitrat: done

ClementWalter (approved):

PR #512, opened 2025-01-22
feat: revert opcode

### LINKED ISSUE #469: revert opcode

obatirou:
Closes #469 

Side effect: change how errors are handled as `Revert` is an `EthereumException`

### DIFF
--- a/cairo/ethereum/cancun/vm/instructions/system.cairo
+++ b/cairo/ethereum/cancun/vm/instructions/system.cairo
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: MIT
+
+from ethereum.cancun.vm.stack import pop
+from ethereum.cancun.vm import Evm, EvmImpl
+from ethereum.cancun.vm.exceptions import Revert, OutOfGasError
+from ethereum.cancun.vm.memory import memory_read_bytes, expand_by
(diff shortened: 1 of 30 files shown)


cairo/ethereum/cancun/vm/instructions/system.cairo
```diff
@@ -0,0 +1,63 @@
...
+        // Read output from memory
+        let output = memory_read_bytes(memory_start_index, size);
+        // Set output
+        EvmImpl.set_output(output);
+    }
+
+    // Raise revert
+    tempvar revert = new EthereumException(Revert);
+    return revert;
+}
```
enitrat, line 63: missing stack re-bind?

enitrat: wait actually im realising we might need to re-bind the stacks right before returning _on every occurence_ and not at the end of functions

imagine: you pop `size` from the stack, your charge_gas tells your you're OOG, but the prover just sent you a wrong size, and because you didnt re-bind stack before returning, you're not squashing the last dict access, which makes your tx unsound

Eikix: yep sounds accurate

obatirou: very good point

enitrat (approved):

PR #513, opened 2025-01-22
feat: map runner exceptions to python exceptions

enitrat:


cairo/tests/fixtures/runner.py
```diff
@@ -560,3 +557,20 @@ def cairo_run(
         return _run_python_vm(cairo_program, cairo_file, main_path, request)
 
     return _run_rust_vm(cairo_program, rust_program, cairo_file, main_path, request)
+
+
+def map_to_python_exception(e: Exception):
+    error_str = str(e)
+
+    # Throw a specialized python exception from the error message, if possible
+    error_type = None
```
Eikix, line 566: unused

cairo/tests/fixtures/runner.py
```diff
@@ -560,3 +557,20 @@ def cairo_run(
...
+    error = re.search(r"Error message: (.*)", error_str)
+    error_type = error.group(1) if error else error_str
+    exception_class = __builtins__.get(error_type)
+    if isinstance(exception_class, type) and issubclass(exception_class, Exception):
```
Eikix, line 570: what is type

enitrat: the metaclass at the root of all python types (checking if its an instance of a class)

Eikix (approved):
ainsi soit-il

PR #518, opened 2025-01-22
feat: op_implementation

### LINKED ISSUE #517: write the `op_implementation` jump table
In EELS there's a dict[op, callable] that dispatches the op number to the right function.

write the cairo equivalent which would be a jump table responsible for calling the right opcode.
This will require _all_ opcodes to take the same implicit args.

```
tempvar offset = opcode *4 + 1
jmp rel offset
call foo;
jmp end;
call bar;
jmp end;
...

### PR DESCRIPTION
**Pull request by enitrat**
Closes #517

### DIFF
--- a/cairo/ethereum/cancun/vm/instructions.cairo
+++ b/cairo/ethereum/cancun/vm/instructions.cairo
@@ -1,7 +1,7 @@
 from starkware.cairo.common.cairo_builtins import BitwiseBuiltin, KeccakBuiltin, PoseidonBuiltin
 from starkware.cairo.common.alloc import alloc
 from ethereum.cancun.vm import Evm, EvmStruct, EvmImpl
-from ethereum.cancun.vm.exceptions import Revert, EthereumException
+from ethereum.cancun.vm.exceptions import Revert, EthereumException, InvalidOpcode
 from ethereum_types.numeric import Uint
 from ethereum_types.bytes import Bytes, BytesStruct
 from ethereum.cancun.vm.instructions.arithmetic import (
@@ -153,8 +153,8 @@ func op_implementation{
     evm: Evm,
 }(opcode: felt) -> EthereumException* {
     // call opcode
-    // count 1 for "next line" and 4 steps per opcode: call, opcode, jmp, end
-    tempvar offset = opcode * 4 + 1;
+    // count 1 for "next line" and 4 steps per opcode: call, opcode, ret
+    tempvar offset = opcode * 3 + 1;
 
     // Prepare arguments
     [ap] = range_check_ptr, ap++;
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Review by obatirou — APPROVED** [review]

File: cairo/ethereum/cancun/vm/instructions.cairo [hunk 1]
```diff
@@ -0,0 +1,703 @@
...
+    [ap] = range_check_ptr, ap++;
+    [ap] = bitwise_ptr, ap++;
+    [ap] = keccak_ptr, ap++;
+    [ap] = poseidon_ptr, ap++;
+    [ap] = evm.value, ap++;
+
+    jmp rel offset;
+    call stop;  // 0x0
+    jmp end;
+    call add;  // 0x1 - ADD
```
ClementWalter, line 169: could just `ret` after each opcode instead of rebinding at the end?

cairo/tests/ethereum/cancun/vm/test_instructions.py
```diff
@@ -0,0 +1,24 @@
...
+def test_op_implementation(cairo_run, evm: Evm, opcode):
+    try:
+        cairo_evm = cairo_run("op_implementation", evm, opcode)
+    except Exception as e:
+        with strict_raises(type(e)):
+            op = Ops(opcode)
+            op_implementation[op](evm)
+        return
+
+    op = Ops(opcode)
```
ClementWalter, line 22: could have the strategy returning the `Ops` type directly, and use `int(op)` in the `cairo_run`

cairo/ethereum/cancun/vm/instructions.cairo
```diff
@@ -0,0 +1,703 @@
...
+    range_check_ptr,
+    bitwise_ptr: BitwiseBuiltin*,
+    keccak_ptr: KeccakBuiltin*,
+    poseidon_ptr: PoseidonBuiltin*,
+    evm: Evm,
+}() {
+    with_attr error_message("ValueError") {
+        assert 1 = 0;
+    }
+    return ();
```
obatirou, line 702: Shouldn't it return an invalidOpcode error instead?
In the execute_code function of EELS
It would allow to not handle the ValueError case in execute_code and raise directly the correct error

```
            try:
                op = Ops(evm.code[evm.pc])
            except ValueError:
                raise InvalidOpcode(evm.code[evm.pc])
```

cairo/ethereum/cancun/vm/instructions.cairo
enitrat: done

cairo/tests/ethereum/cancun/vm/test_instructions.py
enitrat: no, because the strategy could not generate invalid opcode (and our 'op_implementation' is the same as op_implementation(Ops(opcode)))

cairo/ethereum/cancun/vm/instructions.cairo
enitrat: yes, good idea

cairo/tests/ethereum/cancun/vm/test_instructions.py
```diff
@@ -0,0 +1,28 @@
...
+@given(evm=..., opcode=uint8.filter(lambda x: x not in unimplemented_opcodes))
+def test_op_implementation(cairo_run, evm: Evm, opcode):
+    try:
+        cairo_evm = cairo_run("op_implementation", evm, opcode)
+    except Exception as e:
+        with strict_raises(type(e)):
+            try:
+                op = Ops(opcode)
+            except ValueError as e:
+                raise InvalidOpcode(opcode)
```
obatirou, line 22: put 0 also and comment that the opcode is not passed in cairo?

obatirou (approved):

cairo/tests/ethereum/cancun/vm/test_instructions.py
enitrat: doesn't really matter as we only do typechacks

PR #529, opened 2025-01-23
fix copy dict hint to support default value

enitrat:
issue was: copy_dict_segment would copy the dict's data but not its default value.

### DIFF
--- a/python/cairo-addons/src/cairo_addons/hints/dict.py
+++ b/python/cairo-addons/src/cairo_addons/hints/dict.py
@@ -58,9 +58,28 @@ def copy_dict_segment(
     memory: MemoryDict,
     ap: RelocatableValue,
 ):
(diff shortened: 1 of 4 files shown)


python/cairo-addons/src/cairo_addons/hints/dict.py
```diff
@@ -58,9 +58,28 @@ def copy_dict_segment(
     memory: MemoryDict,
     ap: RelocatableValue,
 ):
-    dict_tracker = dict_manager.get_tracker(ids.parent_dict.dict_ptr)
-    copied_data = dict_tracker.data
-    ids.new_dict_ptr = dict_manager.new_dict(segments, copied_data)
+    from collections import defaultdict
```
obatirou, line 61: should hints be diff tested?
creating a test that would ensure at least that they do not raise

enitrat: that what does not raise?

PR #530, opened 2025-01-23
Add a test for benchmarking tx verification cost

ClementWalter:


cairo/src/utils/bytes.cairo
```diff
@@ -288,77 +288,132 @@ func bytes_to_bytes8_little_endian{range_check_ptr}(dst: felt*, bytes_len: felt,
...
-    let bytes8 = [ap - 2];
-    let bytes8_index = [ap - 1];
-
-    let bytes_len = [fp - 4];
-    let bytes = cast([fp - 3], felt*);
-    let pow256 = cast([fp], felt*);
-    let current_byte = bytes[bytes_len - 1 - bytes_index];
-    let current_pow = pow256[bytes8_index];
+    tempvar less_than_8;
+    %{ ids.less_than_8 = int(ids.bytes_len < 8) %}
```
Eikix, line 292: why do we do these operations inside hints? they look inexpensive

also, what happens if the prover 'lies' here? e.g. bytes_len = 10 but the prover writes `ids.less_than_8 = int(True)`

cairo/src/utils/bytes.cairo
```diff
@@ -288,77 +288,132 @@ func bytes_to_bytes8_little_endian{range_check_ptr}(dst: felt*, bytes_len: felt,
...
+    let bytes8 = cast([ap - 2], felt*);
+    let bytes = cast([ap - 1], felt*);
+
+    assert [bytes8] = bytes[0] + bytes[1] * 256 + bytes[2] * 256 ** 2 + bytes[3] * 256 ** 3 + bytes[
+        4
+    ] * 256 ** 4 + bytes[5] * 256 ** 5 + bytes[6] * 256 ** 6 + bytes[7] * 256 ** 7;
+    tempvar continue_loop;
+    tempvar bytes8 = bytes8 + 1;
+    tempvar bytes = bytes + 8;
+    %{ ids.continue_loop = int(ids.bytes_len - (ids.bytes8 - memory[fp - 5]) * 8 >= 8) %}
```
Eikix, line 311: same question here, why do we need to do this in a hint?
and what happens if the prover lies 

I think we end up with the wrong conversion

Eikix: nvm i think that in remaining_i the two assertions are meant to make sure continue_loop is sound and less_than_8 are sound

ClementWalter: Generally speaking there is no good reason to do the control flow in Cairo apart from avoiding to have dozen of hints. Whenever there is a branching, you can non deterministic jump and assert on the branch that the condition is true

ClementWalter: Actually missing a range check. The current constraint only enforces that the finish is consistent with the given bytes_len, but not that there is no overflow on the number  of iterations

ClementWalter: you can look at `poseidon_hash_many` in the corelib for similar strategies, and see also the slack convo

cairo/src/utils/bytes.cairo
```diff
@@ -288,77 +288,132 @@ func bytes_to_bytes8_little_endian{range_check_ptr}(dst: felt*, bytes_len: felt,
...
+    static_assert bytes8 == [ap - 2];
+    static_assert bytes == [ap - 1];
+
+    jmp skip_full_word_loop if less_than_8 != 0;
+
+    full_word_loop:
+    let bytes8 = cast([ap - 2], felt*);
+    let bytes = cast([ap - 1], felt*);
+
+    assert [bytes8] = bytes[0] + bytes[1] * 256 + bytes[2] * 256 ** 2 + bytes[3] * 256 ** 3 + bytes[
```
enitrat, line 305: i think we could benefit from adding soundness assumptions as comments in the codebase. here we know that if the prover lied on less_than_8 then accessing bytes[i], i <= will cause a memory error

enitrat: or at least have this as function doc

enitrat (dismissed):

enitrat (approved):

PR #532, opened 2025-01-23
Fix/stack rebind

### LINKED ISSUE #514: re-bind stack after each access
As soon as a dict_read has been made in a stack access, the EVM's stack must be re-binded.

Typically this means that:
- if you do a _single_ stack.pop/push but if fails (overflow/underflow) then no dict_read was made, meaning it's fine not to rebind it
- otherwise, you must rebind it before every early return.

For simplicity - let's re-bind the stack after each access.

We should also think about a way to ensure that, in serde, when serializing a dict, the end pointer should always be an un-allocated memory cell (proving that the tracking has been made correctly).

Eikix:
closes #514


Eikix:
It seems that in `get_storage_original`, and `sstore` (which calls `get_storage_original`), there is a remapping of the State's original mapping that ends up breaking the check dict consistency check in `serde.py`
I'm trying to figure out if that's inevitable

cairo/ethereum/cancun/vm/instructions/arithmetic.cairo
```diff
@@ -35,32 +35,35 @@ func add{
...
     if (cast(err3, felt) != 0) {
+        EvmImpl.set_stack(stack);
         return err3;
     }
 
     // OPERATION
     let (result, _) = uint256_add([x.value], [y.value]);
     with stack {
         let err4 = push(U256(new U256Struct(result.low, result.high)));
+        EvmImpl.set_stack(stack);
```
obatirou, line 59: why not use set_pc_and_stack like the other ?

cairo/tests/utils/serde.py
```diff
@@ -496,6 +502,16 @@ def _serialize_mapping_struct(
         segment_size = pointers["dict_ptr"] - pointers["dict_ptr_start"]
         dict_ptr = pointers["dict_ptr_start"]
 
+        # Invariant Testing:
+        # We need to ensure that the last dict_ptr points properly
+        # We skip this invariant check if we're serializing `original_mapping` fields
+        # since they might have been updated by reading the `original_storage_trie` field of the state.
```
enitrat, line 508: this is unclear to me: are we skipping this in a very single case only?

cairo/tests/utils/serde.py
```diff
@@ -666,6 +685,24 @@ def serialize_state(self, ptr) -> State:
...
+                parent_storage_dict,
+                (
+                    "ethereum",
+                    "cancun",
+                    "fork_types",
+                    "MappingTupleAddressBytes32U256Struct",
+                ),
+            )
+
+            is_original_state = (
```
enitrat, line 703:
```suggestion
            is_root_state = (
```

cairo/tests/utils/serde.py
```diff
@@ -693,24 +733,14 @@ def serialize_state(self, ptr) -> State:
                         ),
                         parent_storage_dict,
                         Mapping[Tuple[Address, Bytes32], U256],
+                        check_dict_consistency=not is_original_state,
```
enitrat, line 736:
```suggestion
                        # Skip checks for the root state, whose storage_tries tracking is done in the original_storage_tries field.
                        # We can skip the consistency check on dicts
                        check_dict_consistency=not is_root_state,
```

cairo/tests/utils/serde.py
```diff
@@ -679,6 +716,9 @@ def serialize_state(self, ptr) -> State:
                         ),
                         parent_main_dict,
                         Mapping[Address, Optional[Account]],
+                        # If the parent pointers are None, the state is the original state,
+                        # We can skip the consistency check on dicts
+                        check_dict_consistency=not is_original_state,
```
enitrat, line 721:
```suggestion
```

I don't believe this one is required

cairo/ethereum/cancun/vm/instructions/block.cairo
```diff
@@ -287,6 +298,7 @@ namespace Internals {
         with stack {
             let err = push(U256(new U256Struct(block_hashes.value.low, block_hashes.value.high)));
             if (cast(err, felt) != 0) {
+                EvmImpl.set_stack{evm=evm}(stack);
```
obatirou, line 301: missing one before the ok below ?

cairo/tests/utils/serde.py
```diff
@@ -666,6 +685,24 @@ def serialize_state(self, ptr) -> State:
         parent_storage_dict = self._get_trie_parent_ptr(raw_state["_storage_tries"])
 
         while parent_main_dict and parent_storage_dict:
+
+            tmp_parent_main_dict = self._get_mapping_parent_ptr(
```
enitrat, line 689: rename `parent_xx` to `current_xx` and `tmp_parent` to `parent`

cairo/ethereum/cancun/vm/instructions/environment.cairo
```diff
@@ -395,13 +408,15 @@ func returndatacopy{
         Uint(GasConstants.GAS_VERY_LOW + return_data_copy_gas_cost + extend_memory.value.cost.value)
     );
     if (cast(err, felt) != 0) {
+        EvmImpl.set_stack(stack);
         return err;
     }
 
     let memory = evm.value.memory;
```
obatirou, line 415: move this to where it used.
This reference should be next to its usage

enitrat (commented):
👏👏

cairo/ethereum/cancun/vm/instructions/arithmetic.cairo
enitrat: in error cases, PC is not incremented

cairo/ethereum/cancun/vm/instructions/storage.cairo
```diff
@@ -76,36 +77,41 @@ func sload{
...
         }
     }
 
     // PROGRAM COUNTER
-    let env = evm.value.env;
-    EnvImpl.set_state{env=env}(state);
-    EvmImpl.set_env(env);
-    EvmImpl.set_pc_stack(Uint(evm.value.pc.value + 1), stack);
-    EvmImpl.set_accessed_storage_keys(new_accessed_storage_keys);
+    EvmImpl.set_pc(Uint(evm.value.pc.value + 1));
```
obatirou, line 114: use set_pc_stack like others?

cairo/ethereum/cancun/vm/instructions/arithmetic.cairo
obatirou: What I mean is that it is out of the error case and not updated at the end. Which is not like other test where it is rebinded in error case and at the end (which I find easier to follow)

cairo/ethereum/cancun/vm/instructions/block.cairo
enitrat: i think there's a bigger issue in this function, which is that it's taking both `evm` and a `stack` implicit argument.

so either stop takin the stack as implicit argument, and explicitly rebind the stack to the EVM, OR only take the stack as implicit argument and pass the other values explicitly.

Eikix: > i think there's a bigger issue in this function, which is that it's taking both evm and a stack implicit argument.

i dont think it is, this is a util that takes evm as explicit arg

cairo/ethereum/cancun/vm/instructions/environment.cairo
Eikix: i dont understand

enitrat:
@Eikix something in Serde broke with your last commits

cairo/ethereum/cancun/vm/instructions/block.cairo
```diff
@@ -302,6 +302,7 @@ namespace Internals {
                 return err;
             }
         }
+        EvmImpl.set_stack{evm=evm}(stack);
```
enitrat, line 305: this is useless because evm is not returned. Actually you should remove all                 `EvmImpl.set_stack{evm=evm}(stack);` from this function as stack is not taken from the evm.

cairo/tests/utils/serde.py
```diff
@@ -716,9 +713,6 @@ def serialize_state(self, ptr) -> State:
                         ),
                         parent_main_dict,
```
enitrat, line 714:
```suggestion
                        current_main_dict,
```

cairo/tests/utils/serde.py
```diff
@@ -733,14 +727,14 @@ def serialize_state(self, ptr) -> State:
                         ),
                         parent_storage_dict,
```
enitrat, line 728:
```suggestion
                        current_storage_dict,
```

PR #540, opened 2025-01-24
Use garaga for ec_add and placeholder for ec_mul

ClementWalter:
Mainly extracting the ec ops defined initially only for secp256k1 into their generic part and curve specific


cairo/src/utils/uint256.cairo
ClementWalter, line 1: just because the sw corelib defined UInt384 and Uint256 so I add a lot of painful small issue and decided to alias

cairo/src/utils/uint384.cairo
ClementWalter, line 1: see above for Uint256

cairo/src/precompiles/ecmul.cairo
```diff
@@ -1,19 +1,16 @@
...
-// @title EcMul Precompile related functions.
-// @notice This file contains the logic required to run the ec_mul precompile
-// using alt_bn128 library
-// @author @pedrobergamini
-// @custom:namespace PrecompileEcMul
-namespace PrecompileEcMul {
-    const PRECOMPILE_ADDRESS = 0x07;
-    const GAS_COST_EC_MUL = 6000;
+namespace PrecompileEcAdd {
+    const PRECOMPILE_ADDRESS = 0x06;
```
enitrat, line 12: ecadd in ecmul file?

enitrat:
```suggestion
namespace PrecompileEcMul {
    const PRECOMPILE_ADDRESS = 0x07;
```

cairo/src/precompiles/ecmul.cairo
```diff
@@ -1,19 +1,16 @@
...
+    const GAS_COST_EC_ADD = 150;
```
enitrat, line 13:
```suggestion
    const GAS_COST_EC_MUL = 6000;
```

cairo/src/precompiles/ecmul.cairo
ClementWalter: oupsy, copy pasta

PR #570, opened 2025-01-27
dev: remove useless assertion on stack depth in process_message

enitrat:
see:

> edit: actually BEFORE calling process_message_create we ensure that the message depth is < STACK_DEPTH_LIMIT, so you can disregard my message. I think the following case is just never triggered and the assertion is never triggered

> I also have an unresolved question about what happens when doing a CREATE opcode that hits the message stack limit:

> Let's say i'm executing a simple CREATE opcode, starting from process_message

> `process_message(begin_transaction:1) -> execute_code -> CREATE -> generic_create -> process_create_message (begin_transaction: 2) -> process_message❌ fails because stack too deep

> This error is caught in the execute_code, which fills the evm.error, then returns to process_message in which we rollback_transaction, which brings our amount of snapshots back to 1 and not zero

### DIFF
--- a/cairo/ethereum/cancun/vm/interpreter.cairo
+++ b/cairo/ethereum/cancun/vm/interpreter.cairo
@@ -47,7 +47,7 @@ func process_create_message{
     bitwise_ptr: BitwiseBuiltin*,
     keccak_ptr: KeccakBuiltin*,
     poseidon_ptr: PoseidonBuiltin*,
-}(message: Message, env: Environment) -> (Evm, EthereumException*) {
+}(message: Message, env: Environment) -> Evm {
     alloc_locals;
 
     let state = env.value.state;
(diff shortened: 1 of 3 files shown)


cairo/ethereum/cancun/vm/interpreter.cairo
```diff
@@ -157,16 +148,12 @@ func process_message{
...
-    // Check if depth exceeds limit by checking if (depth - limit) is non-negative
-    let is_depth_exceeded = is_nn(message.value.depth.value - STACK_DEPTH_LIMIT);
-    if (is_depth_exceeded != FALSE) {
-        tempvar err = new EthereumException(StackDepthLimitError);
-        tempvar evm = Evm(cast(0, EvmStruct*));
-        return (evm, err);
-    }
+    // EELS checks whether the stack depth limit is reached here.
+    // However, this case is never triggered, because `generic_call` and `generic_create`
+    // check the stack depth limit before calling `process_message`.
```
Eikix, line 156: process_message can also be called by [process_message_call](https://github.com/ethereum/execution-specs/blob/d9febcb7332d62194968dcdb7775adf497c93626/src/ethereum/cancun/vm/interpreter.py#L90) right?

although normally, process_message_call is used only in tx initialization, which suggests it's unlikely the stack can be full

enitrat: exactly

Eikix (approved):

PR #579, opened 2025-01-27
fix: fix incorrect field name in HeaderStruct in blocks.cairo

mdqst:
I noticed that the field `receipt_root` in `HeaderStruct` is incorrect.
The proper name should be `receipts_root` (with an "s" at the end).


enitrat:
receipt is the correct name (mirrors EELS)

PR #580, opened 2025-01-27
feat: prepare_message

### LINKED ISSUE #548: `prepare_message`
Implement `prepare_message`

For the precompiles, we can use some combination of `dw` and memcpy:

```
    dw PRECOMPILE_0;
    dw 0; //prev_value
    dw 1; //new_value
    dw PRECOMPILE_1;
```

and then memcpy this in the dict segment; then increment the dict_ptr_end and update the tracker with `%{update_tracker%}`

We can also memcpy the `preaccessed_addresses` to this `accessed_addresses` because the dedup work will have been done upstream.

enitrat:
Closes #548

### DIFF
--- a/cairo/ethereum/cancun/fork_types.cairo
+++ b/cairo/ethereum/cancun/fork_types.cairo
@@ -8,6 +8,10 @@ from ethereum.utils.numeric import is_zero
 
 using Address = Bytes20;
 
+struct OptionalAddress {
+    value: felt*,
+}
+
 struct SetAddressDictAccess {
     key: Address,
     prev_value: bool,
--- a/cairo/ethereum/cancun/fork_types.cairo
+++ b/cairo/ethereum/cancun/fork_types.cairo
@@ -9,6 +9,8 @@ from ethereum.utils.numeric import is_zero
 using Address = Bytes20;
 
 struct OptionalAddress {
+    // If `value` is the null ptr, the address is treated as None, else, treat `value` as a felt
+    // that represents the underlying address
     value: felt*,
 }
 
--- a/cairo/ethereum/cancun/utils/message.cairo
+++ b/cairo/ethereum/cancun/utils/message.cairo
@@ -1,0 +1,175 @@
+from ethereum.cancun.vm import Message, Environment, MessageStruct, EvmStruct, Evm, EnvImpl
+from ethereum.cancun.fork_types import (
+    Address,
+    OptionalAddress,
+    SetAddress,
+    SetAddressStruct,
+    SetTupleAddressBytes32,
+    SetAddressDictAccess,
+)
+from ethereum_types.bytes import Bytes, BytesStruct, Bytes20
+from ethereum_types.numeric import U256, Uint, bool
+from ethereum.cancun.state import get_account, State, StateStruct
+from ethereum.cancun.utils.address import compute_contract_address
+from ethereum.cancun.transactions import To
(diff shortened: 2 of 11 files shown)


Eikix:
Formatter failed

cairo/tests/ethereum/cancun/utils/test_message.py
```diff
@@ -0,0 +1,82 @@
...
+        data: Bytes,
+        gas: Uint,
+        code_address: Optional[Address],
+        should_transfer_value: bool,
+        is_static: bool,
+        preaccessed_addresses: Set[Address],
+        preaccessed_storage_keys: Set[Tuple[Address, Bytes32]],
+    ):
+        caller, env = caller_and_env
+        cairo_message = cairo_run_py(
```
Eikix, line 55: Is it part of the PR to use cairo_run and implement track_precompiles in rust?

cairo/ethereum/cancun/utils/message.cairo
```diff
@@ -0,0 +1,175 @@
...
+        tempvar msg_data = msg_data;
+        tempvar code = code_account.value.code;
+        tempvar current_target = current_target;
+        tempvar code_address = code_address;
+        tempvar range_check_ptr = range_check_ptr;
+        tempvar bitwise_ptr = bitwise_ptr;
+        tempvar keccak_ptr = keccak_ptr;
+        tempvar poseidon_ptr = poseidon_ptr;
+    }
+    let state_ = cast([ap - 9], StateStruct*);
```
Eikix, line 122: nit: i noticed it works to reassign the variables by name, like so:
`let poseidon_ptr = poseidon_ptr;` not sure why

cairo/ethereum/cancun/fork_types.cairo
```diff
@@ -8,6 +8,10 @@ from ethereum.utils.numeric import is_zero
 
 using Address = Bytes20;
 
+struct OptionalAddress {
+    value: felt*,
```
Eikix, line 12: why is it a pointer to a felt and not an Address?

Eikix (changes requested):
a couple of questions

cairo/ethereum/cancun/fork_types.cairo
enitrat: because the way the Optional system was built in https://github.com/kkrt-labs/keth/pull/437 is that the optional type contains the inner type directly instead of being a pointer to the outer type, thus i'm just re-using the same principle

enitrat: and changing to a pointer to Address would require too many changes in the codebase to be worth (as it requires changing all usages of OptionalAccount as well)

Eikix: why is it a felt pointer and not a felt then?

cairo/tests/ethereum/cancun/utils/test_message.py
enitrat: good catch; done

cairo/ethereum/cancun/fork_types.cairo
Eikix, line 12:
```suggestion
    // If this value is the null ptr, the address is empty, if it's not, then treat it as a felt that represents the underlying address
    value: felt*,
```

Eikix (dismissed):
one last suggestion otherwise lgtm

PR #583, opened 2025-01-28
Uv workspaces for cairo libs

ClementWalter:
I just wanted to have the test runner for a small cairo-ec member. Then I needed to do this all.


python/cairo-addons/src/cairo_addons/testing/hooks.py
```diff
@@ -0,0 +1,228 @@
...
+
+    # Select tests
+    for item in cairo_items:
+        cairo_program = session.cairo_programs[item.fspath]
+        test_hash = xxhash.xxh64(
+            program_hash(cairo_program)
+            + file_hash(item.fspath)
+            + item.nodeid.encode()
+            + file_hash(Path(__file__).parent / "runner.py")
+        ).hexdigest()
```
ClementWalter, line 217: see before we had also 

```
            + file_hash(Path(__file__).parent / "utils" / "serde.py")
            + file_hash(Path(__file__).parent / "utils" / "args_gen.py")
            + file_hash(Path(__file__).parent / "utils" / "strategies.py")
```

but I didn't know how to keep this in the context of this refacto

enitrat (approved):

PR #594, opened 2025-01-28
feat: blooms

### LINKED ISSUE #587: implement bloom.py
implement `logs_bloom` and `add_to_bloom`

enitrat:
Closes #587


cairo/tests/ethereum/cancun/test_bloom.py
```diff
@@ -0,0 +1,22 @@
...
+    def test_add_to_bloom(self, cairo_run, bloom: MutableBloom, bloom_entry: bytes):
+        cairo_bloom = cairo_run("add_to_bloom", bloom, bloom_entry)
+        add_to_bloom(bloom, bloom_entry)
+        assert cairo_bloom == bloom
+
+    @given(logs=...)
+    def test_logs_bloom(self, cairo_run, logs: Tuple[Log, ...]):
+        cairo_bloom = cairo_run("logs_bloom", logs)
+        bloom = logs_bloom(logs)
+        print([hex(b) for b in bloom])
```
Eikix, line 21: print

cairo/tests/utils/args_gen.py
```diff
@@ -152,6 +152,10 @@ class Memory(bytearray):
     pass
 
 
+class MutableBloom(bytearray):
```
Eikix, line 155: why do we need to define the class?

Eikix (changes requested):

cairo/ethereum/cancun/bloom.cairo
```diff
@@ -0,0 +1,167 @@
...
+            dict_ptr_start=bloom.value.dict_ptr_start,
+            dict_ptr=cast(bloom_ptr, Bytes1DictAccess*),
+            len=bloom.value.len,
+        ),
+    );
+
+    return ();
+}
+
+func logs_bloom{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: KeccakBuiltin*}(
```
Eikix, line 95: I wonder if logs bloom should be entirely done in a hint, and we simply assert that for each log in our list of logs, the log bloom has bit flipped to true for each log / topic?

not sure it changes a lot, except that we don't need a mutable bloom struct that accumulates the bloom

cairo/src/utils/utils.cairo
```diff
@@ -1075,4 +1075,263 @@ namespace Helpers {
         }
         jmp read;
     }
+
+    func pow2(i: felt) -> (res: felt) {
```
Eikix, line 1079: should this be moved to python/cairo-core

cairo/tests/utils/args_gen.py
enitrat: bytearray is not serialized by default as a list. we do like we did for memory.

cairo/src/utils/utils.cairo
enitrat: now that it exists, i guess so!

cairo/ethereum/cancun/bloom.cairo
Eikix: we can start with this and check profiler
or ask @ClementWalter

enitrat: our codebase is not ready for this yet (as this would require significant work on the rust side). But we can keep it for later.

edit: Actually im thinking about how it would look like:

do the bit-flip operation in python, which would return a segment with the correct values. 
Then, iterate on this segment, verifying that: 
- for all topic/logs, the value at the given bit index is set
- for the rest of the values, the value is zero.

HOWEVER
in order to ensure all values are zero, we'd need to iterate another time on the segment, ensuring that un-touched bytes are zero, which would require tracking touched bytes, which would require a dict, so perhaps it's not that good in the end.

Eikix: since values are 1s or 0s, you could get away with:
- count the 1s, check that for all topic/logs, the value at the given bit index is set, check that the sum of all values in the segments are equal to the sum of the previously counted 1s?

cairo/ethereum/cancun/bloom.cairo
```diff
@@ -0,0 +1,168 @@
...
+
+    tempvar index = 255;
+    tempvar bloom_end = cast(mutable_bloom.value.dict_ptr, DictAccess*);
+
+    loop:
+    let index = [ap - 2];
+    let dict_ptr = cast([ap - 1], DictAccess*);
+
+    let (value) = dict_read{dict_ptr=dict_ptr}(index);
+    assert bloom_buffer[index] = value;
```
Eikix, line 126: why do we need a bloom_buffer?

cairo/ethereum/cancun/bloom.cairo
```diff
@@ -0,0 +1,168 @@
...
+    assert bitwise_ptr.y = BIT_MASK_11_BITS;
+    tempvar bit_to_set = bitwise_ptr.x_and_y;
+    let bitwise_ptr = bitwise_ptr + BitwiseBuiltin.SIZE;
+
+    let bit_index = BIT_MASK_11_BITS - bit_to_set;
+    let (byte_index, bit_position) = divmod(bit_index, 8);
+
+    // Calculate 1 << (7 - bit_position) using multiplication
+    // This is equivalent to 2^bit_shift where bit_shift = 7 - bit_position
+    let (_, bit_shift_value) = divmod(7 - bit_position, 8);
```
Eikix, line 69:
```suggestion
let bit_shift_value = 7 - bit_position
```

Eikix (dismissed):
one nit but lgtm

cairo/ethereum/cancun/bloom.cairo
```diff
@@ -0,0 +1,168 @@
...
+    felt_to_bytes16_little(hash_low_bytes, hash.value.low);
+    tempvar hash_bytes = Bytes(new BytesStruct(hash_low_bytes, 16));
+    _add_bloom_index(hash_bytes, 0);
+    _add_bloom_index(hash_bytes, 2);
+    _add_bloom_index(hash_bytes, 4);
+
+    return ();
+}
+
+// @dev does not expect to be called with indexes other than 0, 2, 4
```
obatirou, line 51: should this be asserted ?

cairo/ethereum/cancun/bloom.cairo
enitrat: moving values from a mutable DS (dict) to an array

enitrat:
removed unused `U256_to_Uint` in my rebase

cairo/ethereum/cancun/bloom.cairo
enitrat: https://github.com/kkrt-labs/keth/issues/600

obatirou (approved):

cairo/ethereum/cancun/bloom.cairo
enitrat: good one

PR #595, opened 2025-01-28
feat: validate_transaction

### LINKED ISSUE #546: `validate_transaction`
Implement `validate_transaction`:
- [ ] Calculate the intrinsic cost of the tx with `calculate_intrinsic_cost` and verify the tx gas limit covers this cost
- [ ] Ensure the nonce is not U64::Max
- [ ] If the tx is a create tx (to = None), verify the calldata size

obatirou:
Closes #546 

Creates transactions_types file to avoid circular imports. 
Also move the calculate_intrinsic_cost to transaction.py for the same reason (and to match EELS).

### DIFF
--- a/cairo/ethereum/cancun/transactions.cairo
+++ b/cairo/ethereum/cancun/transactions.cairo
@@ -1,122 +1,162 @@
-from ethereum_types.bytes import BytesStruct, Bytes, Bytes0, TupleBytes32
-from ethereum_types.numeric import Uint, U256, U64
-from ethereum.cancun.fork_types import Address, TupleVersionedHash
+from starkware.cairo.common.math_cmp import is_le_felt, is_not_zero
+from starkware.cairo.common.bool import FALSE, TRUE
 
-const TX_BASE_COST = 21000;
-const TX_DATA_COST_PER_NON_ZERO = 16;
-const TX_DATA_COST_PER_ZERO = 4;
-const TX_CREATE_COST = 32000;
-const TX_ACCESS_LIST_ADDRESS_COST = 2400;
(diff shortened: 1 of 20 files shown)


Eikix:
another round of trunk

cairo/ethereum/cancun/transactions.cairo
```diff
@@ -1,122 +1,162 @@
-from ethereum_types.bytes import BytesStruct, Bytes, Bytes0, TupleBytes32
-from ethereum_types.numeric import Uint, U256, U64
-from ethereum.cancun.fork_types import Address, TupleVersionedHash
-
-const TX_BASE_COST = 21000;
```
enitrat, line 5: tx types being in transactions.py i wouldn't have moved them out (unless there's another reason)

obatirou:
> another round of trunk

this should not happen. I did not touch those files. It seems there is a problem with trunk as it should have been caught before merging

cairo/ethereum/cancun/transactions.cairo
```diff
@@ -1,122 +1,162 @@
...
+    if (is_gas_insufficient != FALSE) {
+        tempvar res = bool(FALSE);
+        return res;
+    }
+
+    let is_nonce_out_of_range = U256_le(U256(new U256Struct(2 ** 64 - 1, 0)), tx_nonce);
+    if (is_nonce_out_of_range.value != FALSE) {
+        tempvar res = bool(FALSE);
+        return res;
+    }
```
enitrat, line 151: simpler check: is_nonce_out_of_range =  (tx_nonce.low < 2**64 -1)+ tx_nonce.high

enitrat (approved):

PR #606, opened 2025-01-29
Feat/secpk1 recover

### LINKED ISSUE #564: Implement secp256k1_recover from the existing garaga-adapted signature verification utils using try_recover_eth_address.
- [ ] Create an ethereum.crypto.elliptic_curver file, in which you implement `secp256k1_recover` using the logic of the exisiting `try_recover_eth_address` and adapting to the new types.

Eikix:
closes #564


cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -0,0 +1,62 @@
...
+    }
+    let max_value = Uint256(secp256k1.P_LOW_128 - 1, secp256k1.P_HIGH_128);
+    let x_uint256 = uint384_to_uint256(public_key_point.x);
+    assert_uint256_le(x_uint256, max_value);
+    let y_uint256 = uint384_to_uint256(public_key_point.y);
+    assert_uint256_le(y_uint256, max_value);
+
+    tempvar x = U256(new U256Struct(x_uint256.low, x_uint256.high));
+    tempvar y = U256(new U256Struct(y_uint256.low, y_uint256.high));
+    return (x=x, y=y);
```
Eikix, line 61: here if we return Bytes we'll have to unpack it into u64s for keccak
If we leave it as U256, we'll be able to use `keccak_uint256s_bigend`, identically as `signature.cairo`

enitrat: ok, let's do both:

rename the current one to secp256k1_recover_uint256_bigends
create secp256k1_recover that returns bytes, which is the one you test ?

cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -0,0 +1,63 @@
...
+    add_mod_ptr: ModBuiltin*,
+    mul_mod_ptr: ModBuiltin*,
+    bitwise_ptr: BitwiseBuiltin*,
+    keccak_ptr: KeccakBuiltin*,
+    poseidon_ptr: PoseidonBuiltin*,
+}(r: U256, s: U256, v: U256, msg_hash: Hash32) -> (x: U256, y: U256) {
+    alloc_locals;
+
+    // reverse endianness of msg_hash
+    let (msg_hash_reversed) = uint256_reverse_endian([msg_hash.value]);
```
Eikix, line 36: not sure why I have to reverse endianness for things to work out

cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -0,0 +1,63 @@
...
+
+    // Convert inputs to UInt384 for try_recover_public_key
+    let r_uint384 = uint256_to_uint384([r.value]);
+    let s_uint384 = uint256_to_uint384([s.value]);
+    // parameter `v` MUST be a `U256` with `low` value equal to y parity and high value equal to 0
+    // see: <https://github.com/ethereum/execution-specs/blob/master/src/ethereum/cancun/transactions.py#L199>
```
enitrat, line 42: wrong link btw
https://github.com/ethereum/execution-specs/blob/53fbc954b8e9e7dd86060814dfbf836791a22f4c/src/ethereum/crypto/elliptic_curve.py#L49

cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -0,0 +1,63 @@
...
+    let y_parity = v.value.low;
+    let msg_hash_uint384 = uint256_to_uint384(msg_hash_reversed);
+
+    let (public_key_point, success) = try_recover_public_key(
+        msg_hash=msg_hash_uint384, r=r_uint384, s=s_uint384, y_parity=y_parity
+    );
+
+    if (success == 0) {
```
enitrat, line 50: in which cases can success be zero? should this return the point at infinity if so?

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
```diff
@@ -79,3 +92,270 @@ func sign_to_uint384_mod_secp256k1(sign: felt) -> UInt384 {
         return res;
     }
 }
+
+// @notice Similar to `recover_public_key`, but handles the case where 'x' does not correspond to a point on the
+// curve gracefully.
+// @param msg_hash The signed message hash.
```
enitrat, line 98:
```suggestion
// @param msg_hash The signed message hash, big-endian
```

cairo/ethereum/crypto/elliptic_curve.cairo
enitrat: because Hash32 is little-endian

enitrat (approved):

PR #609, opened 2025-01-29
feat: process_transaction

### LINKED ISSUE #547: `process_transaction`
Implement `process_transaction`. Requires #546 
Requires `prepare_message`

obatirou:
Close #547


obatirou:
Blocked by https://github.com/kkrt-labs/keth/pull/616

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,323 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+    // Create preaccessed storage keys
+    let (preaccessed_storage_keys_ptr) = dict_new_empty();
+    tempvar preaccessed_storage_keys = SetTupleAddressBytes32(
+        new SetTupleAddressBytes32Struct(
+            dict_ptr_start=cast(preaccessed_storage_keys_ptr, SetTupleAddressBytes32DictAccess*),
+            dict_ptr=cast(preaccessed_storage_keys_ptr, SetTupleAddressBytes32DictAccess*),
+        ),
+    );
+
+    if (tx.value.legacy_transaction.value == 0) {
```
obatirou, line 341: Equivalent to
https://github.com/ethereum/execution-specs/blob/d9febcb7332d62194968dcdb7775adf497c93626/src/ethereum/cancun/fork.py#L750

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+        assert tx_to = tx.value.access_list_transaction.value.to;
+        assert tx_value = tx.value.access_list_transaction.value.value;
+        assert blob_gas_fee = Uint(0);
+    }
+
+    let effective_gas_fee = tx_gas.value * env.value.gas_price.value;
+
+    // Calculate available gas
+    let intrinsic_cost = calculate_intrinsic_cost(tx);
+    let gas = tx_gas.value - intrinsic_cost.value;
```
enitrat, line 298:
```suggestion
    // Calculate available gas. Does not underflow, as it was checked in validate_transaction before.
    let intrinsic_cost = calculate_intrinsic_cost(tx);
    let gas = tx_gas.value - intrinsic_cost.value;
```

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+
+    // Increment nonce
+    increment_nonce{state=state}(sender);
+
+    // Deduct gas fee from sender
+    with_attr error_message("OverflowError") {
+        assert sender_account.value.balance.value.high = 0;
+    }
+    let sender_balance_after_gas_fee = sender_account.value.balance.value.low - effective_gas_fee -
+        blob_gas_fee.value;
```
enitrat, line 308: do we know if this can be problematic to have accounts with very high balances?

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+        new U256Struct(sender_balance_after_gas_fee, 0)
+    );
+    set_account_balance{state=state}(sender, sender_balance_after_gas_fee_u256);
+    EnvImpl.set_state{env=env}(state);
+
+    // Create preaccessed addresses and write coinbase
+    let (preaccessed_addresses_ptr) = dict_new_empty();
+    tempvar preaccessed_addresses_ptr_start = preaccessed_addresses_ptr;
+    let address = env.value.coinbase;
+    hashdict_write{poseidon_ptr=poseidon_ptr, dict_ptr=preaccessed_addresses_ptr}(
```
enitrat, line 323:
```suggestion
    hashdict_write{dict_ptr=preaccessed_addresses_ptr}(
```

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+        new SetTupleAddressBytes32Struct(
+            dict_ptr_start=cast(preaccessed_storage_keys_ptr, SetTupleAddressBytes32DictAccess*),
+            dict_ptr=cast(preaccessed_storage_keys_ptr, SetTupleAddressBytes32DictAccess*),
+        ),
+    );
+
+    if (tx.value.legacy_transaction.value == 0) {
+        let access_list = tx.value.access_list_transaction.value.access_list;
+        process_access_list{
+            poseidon_ptr=poseidon_ptr,
```
enitrat, line 344:
```suggestion
        process_access_list{
```

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+
+    let keccak_ptr = cast([ap - 5], KeccakBuiltin*);
+    let bitwise_ptr = cast([ap - 4], BitwiseBuiltin*);
+    let poseidon_ptr = cast([ap - 3], PoseidonBuiltin*);
+    let preaccessed_addresses = SetAddress(cast([ap - 2], SetAddressStruct*));
+    let preaccessed_storage_keys = SetTupleAddressBytes32(
+        cast([ap - 1], SetTupleAddressBytes32Struct*)
+    );
+
+    tempvar code_address = OptionalAddress(cast(0, Address*));
```
enitrat, line 369: im surprised this compiles, given that the inner is supposed to be:

```suggestion
    tempvar code_address = OptionalAddress(cast(0, felt*));
```

I guess it's ok because it's a pointer of size 1

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+        code_address,
+        bool(1),
+        bool(0),
+        preaccessed_addresses,
+        preaccessed_storage_keys,
+    );
+    let output = process_message_call{env=env}(message);
+
+    // Calculate gas refund
+    assert output.value.refund_counter.value.high = 0;
```
enitrat, line 385: wrap in a with_attr so that if this ever fails we can debug easily

cairo/ethereum/cancun/fork.cairo
```diff
@@ -195,3 +221,322 @@ func make_receipt{range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: Kec
...
+    let coinbase_balance_after_mining_fee_is_zero = U256__eq__(
+        coinbase_balance_after_mining_fee, zero_u256
+    );
+    if (coinbase_balance_after_mining_fee_is_zero.value == FALSE) {
+        set_account_balance{state=state}(env.value.coinbase, coinbase_balance_after_mining_fee);
+        tempvar poseidon_ptr = poseidon_ptr;
+        tempvar state = state;
+    } else {
+        let is_empty = account_exists_and_is_empty{state=state}(env.value.coinbase);
+        if (is_empty.value == TRUE) {
```
enitrat, line 425:
```suggestion
        if (is_empty.value != FALSE) {
```

cairo/tests/ethereum/cancun/test_fork.py
```diff
@@ -98,3 +101,24 @@ def test_make_receipt(
...
+            gas_used_cairo, logs_cairo, error_cairo = cairo_run(
+                "process_transaction", env, tx
+            )
+        except Exception as cairo_exception:
+            try:
+                process_transaction(env, tx)
+            except Exception as python_exception:
+                error = re.search(r"Error message: (.*)", str(cairo_exception))
+                error_type = error.group(1) if error else type(cairo_exception).__name__
+                assert type(python_exception).__name__ == error_type
```
enitrat, line 118: what is this?

enitrat (changes requested):
💯excellent

cairo/tests/ethereum/cancun/test_fork.py
obatirou: We must execute the cairo code before the python one as we cannot do copy of the input (due to frozen properties) so the exact error type raise is not known.

The thing is: cairo will raise an AssertionError when there is an ASSERT_EQ. In this function we use those ASSERT_EQ to raise a specific error and we do not know the type of the error raise unless we do this.
To retrieve the error, we need to compare type raised by the python code. 
So this hack, open to better ideas
 
```
def map_to_python_exception(e: Exception):
    error_str = str(e)

    # Throw a specialized python exception from the error message, if possible
    error = re.search(r"Error message: (.*)", error_str)
    error_type = error.group(1) if error else error_str
    exception_class = __builtins__.get(error_type)
    if isinstance(exception_class, type) and issubclass(exception_class, Exception):
        raise exception_class() from e

    # Fallback to generic exception
    if "An ASSERT_EQ instruction failed" in error_str:
        raise AssertionError(e) from e
    raise Exception(error_str) from e
```

enitrat: I'm not sure I get it 🤔 

We raise ` AssertionError when there is an ASSERT_EQ` if there is no error message the assertion is wrapped with. Otherwise we are supposed to raise with the correct error type, which is taken from the with_attr statement

cairo/ethereum/cancun/fork.cairo
obatirou: good question, eels assume they are uint

cairo/tests/ethereum/cancun/test_fork.py
obatirou: For example this one: 
```
with_attr error_message("InvalidBlock") {
        assert is_valid.value = TRUE;
    }
```

This will raise with an AssertionError because "An ASSERT_EQ instruction failed" is in the error.
But it will also contain the error_message InvalidBlock

Eikix: I think InvalidBlock is a legacy piece of code since it's supposed to be an EthereumException

The rule for now is:
- if the error raised is a python std lib error, you can do 
`with_attr error_message("ExactNameOfTheError") { assert is_valid.value = TRUE; }` and then the runner will convert it to the right python error
- if the error raised is an ethereumexception, apply the same rules as for the other ethereum exceptions

obatirou: InvalidBlock is raised in EELS and catch anywhere. It is even raised by the state_transition function
https://github.com/ethereum/execution-specs/blob/d9febcb7332d62194968dcdb7775adf497c93626/src/ethereum/cancun/fork.py#L177

enitrat: > This will raise with an AssertionError because "An ASSERT_EQ instruction failed" is in the error.

Surprising, this should be handled by the runner. let me check

enitrat: @obatirou pushed a fix. 
@eikix this will also be necessary for your PR

enitrat: @obatirou pushed a fix. 
@eikix this will also be necessary for your PR

obatirou: thank you 
Should we let as is then ?

obatirou: might have to limit the number of examples also as it is very long even with custom strategy

Eikix: > @obatirou pushed a fix. @Eikix this will also be necessary for your PR

ok i'll rebase and use this to modify the signature of recover_sender to raise instead of return the exception

cairo/ethereum/cancun/fork.cairo
enitrat: Uint is actually un-bounded so I believe this can be problematic... we'll figure out

enitrat (approved):

PR #625, opened 2025-01-31
Initial circuit compiler

ClementWalter:
## Why

The idea of this PR is to leverage regular cairo zero syntax to compile circuits for the mod builtins.

## What

- [x] make the cairo base prime a parameter of the test
- [x] allow for running tests against any prime
- [x] write a compile function to extract the circuit from the cairo compilation
- [x] generate templates and scripts to dump results

This PR adds the `compile_circuit` scripts that can be run with uv like

```
uv run compile_circuit src/cairo_ec/circuits.cairo
```

to generate a `circuits_compiled.cairo` file which actually rewrite the initial logic into a function calling the mod bultins.

The generated artifacts and the compiler are diff tested against python reference logic and cairo regular run.

## How

Ask cursor


python/cairo-ec/src/cairo_ec/compiler.py
```diff
@@ -0,0 +1,55 @@
+from starkware.cairo.lang.compiler.encode import decode_instruction
+from starkware.cairo.lang.compiler.identifier_definition import TypeDefinition
+from starkware.cairo.lang.compiler.instruction import Instruction, Register
+from starkware.cairo.lang.compiler.program import Program
+
+from cairo_addons.testing.utils import flatten
+
+
+def circuit_compile(cairo_program: Program, circuit: str):
```
obatirou, line 9: This is far from trivial. Some documentation is necessary.

obatirou (approved):

PR #647, opened 2025-02-04
extract all custom hints to python functions

enitrat:


crates/cairo-addons/src/vm/hint_utils/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+        len_felt.try_into().map_err(|_| MathError::Felt252ToUsizeConversion(Box::new(len_felt)))?;
+
+    let data = vm.get_relocatable(ptr)?;
+
+    let mut bytes = Vec::new();
+    for i in 0..len {
+        let byte_addr = (data + i)?;
+        let byte = vm.get_integer(byte_addr)?.into_owned();
+        bytes.push(byte);
+    }
```
obatirou, line 36: you could use collect like below in deserialize_sequence 
```rust
(0..len)
        .map(|i| {
            let byte_addr = (data + i)?;
            vm.get_integer(byte_addr).map(|b| b.into_owned())
        })
        .collect::<Result<Vec<_>, _>>()
```

Or if you want to keep the manual push, use `Vec::with_capacity(len)` to avoid reallocating at each push if not mistaken

obatirou (approved):

PR #651, opened 2025-02-04
assert is (not) on on curve

ClementWalter:


python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
```diff
@@ -19,3 +21,22 @@ func ec_double(x0: felt, y0: felt, a: felt) -> (felt, felt) {
 
     end:
 }
+
+// @dev Assert that a point is on the curve.
+func assert_on_curve(x: felt, y: felt, a: felt, b: felt) {
```
obatirou, line 26: nit: add ec in function name even for assert

ClementWalter: `ec_assert_on` ?

python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
```diff
@@ -19,3 +21,22 @@ func ec_double(x0: felt, y0: felt, a: felt) -> (felt, felt) {
...
+    assert y * y = (x * x * x + a * x + b);
+
+    return ();
+
+    end:
+}
+
+// @dev Assert that a point is not on the curve.
+// @dev If the point is on the curve, diff = 0 and 1 / diff will fail.
```
enitrat, line 35: which not checking diff != 0 then ?

obatirou (dismissed):

PR #655, opened 2025-02-04
fix: is_zero opcode

enitrat:
the jumptable was calling the is_zero cairo instead of the is_zero opcode.


ClementWalter (commented):
Can we keep the is zero name for the opcode and maybe import with alias when required ?

ClementWalter (approved):

PR #666, opened 2025-02-05
dev: refactor dict_{read/write} for coherency

### LINKED ISSUE #661: bug: dict_{read/write} tracker data is incoherent with hashdict
## why
problem was using dict_write to write data and hashdict_read in places in the codebase

## how
we need to replace the key logic used in all dict_write / read with those of hashdict_write /read

enitrat:
Closes #661

### DIFF
--- a/cairo/src/utils/dict.cairo
+++ b/cairo/src/utils/dict.cairo
@@ -122,35 +122,6 @@ func hashdict_read{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
     return (value=value);
 }
 
-// A wrapper around dict_read that hashes the key before accessing the dictionary if the key
-// does not fit in a felt.
-// @dev This version returns 0, if the key is not found and the dict is NOT a defaultdict.
-// @param key_len: The readnumber of felt values used to represent the key.
-// @param key: The key to access the dictionary.
-// TODO: write the associated squash function.
-func hashdict_get{poseidon_ptr: PoseidonBuiltin*, dict_ptr: DictAccess*}(
-    key_len: felt, key: felt*
-) -> (value: felt) {
-    alloc_locals;
-    local felt_key;
-    if (key_len == 1) {
-        assert felt_key = key[0];
-        tempvar poseidon_ptr = poseidon_ptr;
-    } else {
-        let (felt_key_) = poseidon_hash_many(key_len, key);
-        assert felt_key = felt_key_;
-        tempvar poseidon_ptr = poseidon_ptr;
-    }
-
-    local value;
-    %{ hashdict_get %}
-    dict_ptr.key = felt_key;
-    dict_ptr.prev_value = value;
-    dict_ptr.new_value = value;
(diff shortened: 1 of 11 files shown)


cairo/tests/utils/args_gen.py
```diff
@@ -1110,12 +1128,11 @@ def generate_dict_arg(
...
     # The last element is the original_segment_stop pointer.
     # Because this is a new dict, this is 0 (null ptr).
-    # This does not apply to stack and memory (hash_mode=False), in which case there's only 2 elements.
-    data_to_load = (
-        [dict_ptr, current_ptr, parent_ptr or 0]
-        if (hash_mode is not False)
-        else [dict_ptr, current_ptr]
-    )
+    # This does not apply to Stack, Memory and MutableBloom, in which case there's only 2 elements.
+    if arg_type_origin in (Stack, Memory, MutableBloom):
```
Eikix, line 1132: should we add jumpdest?

cairo/src/utils/dict.cairo
```diff
@@ -29,6 +30,34 @@ func dict_new_empty() -> (res: DictAccess*) {
...
+    dict_ptr.key = key_;
+    dict_ptr.prev_value = value;
+    dict_ptr.new_value = value;
+    let dict_ptr = dict_ptr + DictAccess.SIZE;
+    return (value=value);
+}
+
+// Writes a value to the dictionary, overriding the existing value.
+// @dev The key is as a tuple of size 1 in the tracker.
+func dict_write{dict_ptr: DictAccess*}(key_: felt, new_value: felt) {
```
Eikix, line 51: this means that every dict in the repo should be a default dict since hashdictread assumes it, did we change all instantiations to default_dicts?

ClementWalter: or update the hashdict hint to not call default_factory by default

``` 
@register_hint
def hashdict_read(dict_manager: DictManager, ids: VmConsts, memory: MemoryDict):
    dict_tracker = dict_manager.get_tracker(ids.dict_ptr)
    dict_tracker.current_ptr += ids.DictAccess.SIZE
    preimage = tuple([memory[ids.key + i] for i in range(ids.key_len)])
    # Not using [] here because it will register the value for that key in the tracker.
    ids.value = dict_tracker.data.get(preimage, dict_tracker.data.default_factory())
```

can be updated into

```
@register_hint
def hashdict_read(dict_manager: DictManager, ids: VmConsts, memory: MemoryDict):
    dict_tracker = dict_manager.get_tracker(ids.dict_ptr)
    dict_tracker.current_ptr += ids.DictAccess.SIZE
    preimage = tuple([memory[ids.key + i] for i in range(ids.key_len)])
    # Not using [] here because it will register the value for that key in the tracker.
    ids.value = dict_tracker.data.get(preimage)
    if ids.value is None:
        ids.value = dict_tracker.data.default_factory()
```

enitrat: Done

ClementWalter (approved):

PR #670, opened 2025-02-05
fix: process-transaction src & strategy

enitrat:
- fixes the strategy of process_transaction to only generate valid gas values regarding basefee
- fixes the preaccessed addresses & keys to be instanciated with a default dict with default 0
- fixes the state used in reads post-execution to be the updated on from the env
- saturate blob gas / gas price values to 2**64
- stop early taylor exponentiation when accumulated numerator is > 2**128
- limit excess_blob_gas to 10*target in strategy, to avoid the pythonic taylor exponentiation to run forever.
- fix the assertions made in `process_transaction` to handle raises AND exceptions in return values

### DIFF
--- a/cairo/ethereum/utils/numeric.cairo
+++ b/cairo/ethereum/utils/numeric.cairo
@@ -1,4 +1,4 @@
-from starkware.cairo.common.math_cmp import is_le, is_not_zero
+from starkware.cairo.common.math_cmp import is_le, is_not_zero, is_le_felt
 from starkware.cairo.common.uint256 import uint256_reverse_endian
 from ethereum_types.numeric import Uint, U256, U256Struct, bool, U64
 from ethereum_types.bytes import Bytes32, Bytes32Struct, Bytes20, Bytes
@@ -86,6 +86,7 @@ func ceil32{range_check_ptr}(value: Uint) -> Uint {
     return result;
 }
 
+// @dev: Saturates when the numerator_accumulated * numerator is greater than 2**128 - 1
 func taylor_exponential{range_check_ptr}(factor: Uint, numerator: Uint, denominator: Uint) -> Uint {
     let output = 0;
     let i = 1;
@@ -109,6 +110,14 @@ func _taylor_exponential{range_check_ptr}(
     let output = output + numerator_accumulated;
     let value = numerator_accumulated * numerator;
     let div = denominator * i;
+
+    let saturate = is_le_felt(2 ** 128, value);
+    if (saturate != 0) {
+        // Return current accumulated output/denominator when we hit saturation
+        let (res, _) = divmod(output, denominator);
+        return res;
+    }
+
     let (numerator_accumulated, _) = divmod(value, div);
     let i = i + 1;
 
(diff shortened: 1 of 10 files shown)


ClementWalter (dismissed):
we may want to have `set_new` to make it easier

cairo/tests/ethereum/utils/test_numeric.py
```diff
@@ -11,6 +11,24 @@
 from tests.utils.strategies import uint128
 
 
+def taylor_exponential_limited(
```
ClementWalter, line 14: move into class

ClementWalter (dismissed):

cairo/tests/ethereum/cancun/test_fork.py
```diff
@@ -201,22 +218,25 @@ def test_make_receipt(
...
         except Exception as e:
             with strict_raises(type(e)):
-                process_transaction(env, tx)
+                output_py = process_transaction(env, tx)
+                # The Cairo Runner will raise if an exception is in the return values.
+                if len(output_py) == 3:
+                    raise output_py[2]
+                assert env_cairo == env
+                assert gas_cairo == output_py[0]
+                assert logs_cairo == output_py[1]
```
ClementWalter, line 233: if proces_transaction raises, the following lines are not executed, you need to put them out of the context manager

ClementWalter (changes requested):
probably need to update the assert in `with strict_rases`

cairo/tests/ethereum/cancun/test_fork.py
enitrat: I think I just need to put them right above the raise output-py[2]?

cairo/ethereum/utils/numeric.cairo
```diff
@@ -109,6 +110,14 @@ func _taylor_exponential{range_check_ptr}(
     let output = output + numerator_accumulated;
     let value = numerator_accumulated * numerator;
     let div = denominator * i;
+
+    let saturate = is_le_felt(2 ** 128, value);
+    if (saturate != 0) {
+        // Return current accumulated output/denominator when we hit saturation
+        let (res, _) = divmod(output, denominator);
+        return res;
+    }
```
obatirou, line 119: Note: we can let it as a failsafe but I am not sure this will happen. Taylor exponential is used only in calculate_blob_gas_price. The max it can get in input is MAX_BLOB_GAS_PER_BLOCK. Strategies should reflect that even with MAX_BLOB_GAS_PER_BLOCK*2 this will be greater than 2***128.

The following test is passing
```python
    @given(
        factor=st.just(MIN_BLOB_GASPRICE),
        numerator=st.just(MAX_BLOB_GAS_PER_BLOCK).map(Uint),
        denominator=st.just(BLOB_GASPRICE_UPDATE_FRACTION),
    )
    def test_taylor_exponential(
        self, cairo_run, factor: Uint, numerator: Uint, denominator: Uint
    ):
        assert taylor_exponential(factor, numerator, denominator) == cairo_run(
            "taylor_exponential", factor, numerator, denominator
        )
```

enitrat: > I am not sure this will ever happen with current constatns

This will definitely trigger in ef-tests in case inputs are not the current blob gas target

ClementWalter (approved):

PR #672, opened 2025-02-06
fix: environment patched import fails in EELS as it has no `traces` field

### LINKED ISSUE #671: bug: patched Environment import trigger errors in apply_body
## Context:
apply_body code has:

```python
system_tx_env = vm.Environment(
            caller=SYSTEM_ADDRESS,
            origin=SYSTEM_ADDRESS,
            block_hashes=block_hashes,
            coinbase=coinbase,
            number=block_number,
            gas_limit=block_gas_limit,
            base_fee_per_gas=base_fee_per_gas,
            gas_price=base_fee_per_gas,
            time=block_time,
            prev_randao=prev_randao,
            state=state,
            chain_id=chain_id,
            traces=[],
            excess_blob_gas=excess_blob_gas,
            blob_versioned_hashes=(),
            transient_storage=TransientStorage(),
        )
```

But Environment from args_gen does not have traces field.
```

# All these classes are auto-patched in test imports in cairo/tests/conftests.py
@dataclass
class Environment(
    make_dataclass(
        "Environment",
        [(f.name, f.type, f) for f in fields(EnvironmentBase) if f.name != "traces"],
        namespace={"__doc__": EnvironmentBase.__doc__},
    )
):
    def __eq__(self, other):
        return all(
            getattr(self, field.name) == getattr(other, field.name)
            for field in fields(self)
        )

```

This triggers an error "Trying to call __init__ on Environment with unknown field traces"

## Problem
We patched all imports of Environment to take our redefined class. This applies to inner code of EELS too. This forces us to make sure our modified types have the same fields as EELS classes.

## Proposed fix
Stop removing traces field from the modified class and skip it in serialization into cairo altogether

Eikix:
closes #671

### DIFF
--- a/cairo/tests/utils/args_gen.py
+++ b/cairo/tests/utils/args_gen.py
@@ -320,7 +320,7 @@ class FlatTransientStorage:
 class Environment(
     make_dataclass(
         "Environment",
-        [(f.name, f.type, f) for f in fields(EnvironmentBase)],
+        [(f.name, f.type, f) for f in fields(EnvironmentBase) if f.name != "traces"],
         namespace={"__doc__": EnvironmentBase.__doc__},
     )
 ):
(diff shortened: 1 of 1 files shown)


enitrat:
doing these patches manually does not sound like a good solution. Why not overload the __init__ method ?

ClementWalter (approved):

PR #673, opened 2025-02-06
dev: remove logging perf of coverage in main

Eikix:
le logger de perf de coverage rend la nightly trop verbose
je recommande de supprimer les log info dans le coverage

### DIFF
--- a/python/cairo-addons/src/cairo_addons/testing/coverage.py
+++ b/python/cairo-addons/src/cairo_addons/testing/coverage.py
@@ -1,13 +1,8 @@
-import logging
 from pathlib import Path
-from time import perf_counter
 from typing import List
 
 import polars as pl
 from starkware.cairo.lang.compiler.program import Program
-
-logging.basicConfig(format="%(levelname)-8s %(message)s")
-logger = logging.getLogger("timer")
 
 
 def coverage_from_trace(
@@ -18,8 +13,6 @@ def coverage_from_trace(
         if program.debug_info is None:
             raise ValueError("Program debug info is not available")
 
-        logger.info("Begin coverage")
-        start = perf_counter()
         coverage = (
             pl.DataFrame(
                 {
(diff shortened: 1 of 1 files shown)


enitrat:
make it debug level?

ClementWalter (changes requested):
make it debug

Eikix:
ok will set nightly in info and ths in debug

Eikix:
change into logging total coverage time once

python/cairo-addons/src/cairo_addons/testing/fixtures.py
```diff
@@ -67,15 +72,33 @@ def coverage(cairo_program: Program, cairo_file: Path, worker_id: str):
...
+    yield timed_coverage
 
     # If no coverage is collected, don't dump anything
-    # This can happen if the all the tests raise Cairo exceptions
     if not reports:
         return
 
+    # Update global total at the end of the fixture
+    global _total_coverage_time
+    _total_coverage_time += coverage_time
```
ClementWalter, line 96: Idon't get the need for a global _total_coverage_time

python/cairo-addons/src/cairo_addons/testing/fixtures.py
ClementWalter, line 1: I don't like the fact to use a global in fixture. if we want to accumulate time, I'd rather create a session scoped fixture, or hide this global mechanism in the coverage module, and just logs it here in the fixutre

ClementWalter (changes requested):
since this is far from any prio, I wouldn't mind either just putting debug to the logger as discussed earlier, of remove this logger, but no accumulation like that

Eikix:
> since this is far from any prio, I wouldn't mind either just putting debug to the logger as discussed earlier, of remove this logger, but no accumulation like that

ok i vote to remove it then, it's so verbose it's unreadable
if we need it we can profile the coverage

ClementWalter (approved):

PR #693, opened 2025-02-07
add assert_is_on_curve

ClementWalter:


python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -108,17 +35,28 @@ func get_random_point{
...
+    assert_is_on_curve(x=x, y=&y_try, a=a, b=b, g=g, is_on_curve=&is_on_curve, p=p);
 
-    let (y, is_on_curve) = try_get_point_from_x(x=x_384, v=0, a=a, b=b, g=g, p=p);
+    assert is_on_curve.d3 = 0;
+    assert is_on_curve.d2 = 0;
+    assert is_on_curve.d1 = 0;
 
-    if (is_on_curve != 0) {
-        let point = G1Point(x=x_384, y=y);
+    if (is_on_curve.d0 != 0) {
```
zmalatrax, line 53: Why not using `try_get_point_from_x` ?
Avoid function call for readability, performance?

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -1,103 +1,30 @@
...
-    assert input[7] = UInt384(is_on_curve, 0, 0, 0);
-
-    assert add_mod_ptr[0] = ModBuiltin(
-        p=[p], values_ptr=input, offsets_ptr=add_offsets_ptr, n=add_mod_n
-    );
-    assert mul_mod_ptr[0] = ModBuiltin(
-        p=[p], values_ptr=input, offsets_ptr=mul_offsets_ptr, n=mul_mod_n
-    );
+    assert_is_on_curve(x=x, y=&y_try, a=a, b=b, g=g, is_on_curve=&is_on_curve, p=p);
+    // TODO: Add a check for v
```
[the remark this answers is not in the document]
ClementWalter: I just realized this was missing indeed, but out of the scope of this refacto

python/cairo-ec/src/cairo_ec/ec_ops.cairo
ClementWalter: actually I'd like to remove `try_get_point_from_x` I guess, I can

python/cairo-addons/src/cairo_addons/hints/curve.py
```diff
@@ -61,8 +61,8 @@ def compute_y_from_x_hint(ids: VmConsts, segments: MemorySegmentManager):
     x = uint384_to_int(ids.x.d0, ids.x.d1, ids.x.d2, ids.x.d3)
     rhs = (x**3 + a * x + b) % p
 
-    ids.is_on_curve = is_quad_residue(rhs, p)
-    if ids.is_on_curve == 1:
+    is_on_curve = is_quad_residue(rhs, p)
+    if is_on_curve == 1:
```
zmalatrax, line 65: Better to follow this pattern when writing hints?
We use a variable as a regular python var, and then we add it to the `ids` through `segments.load_data()` at the end of the hint?

ClementWalter: it's because I've updated the cairo def from `felt` to `Uint384*` so it's a pointer

python/cairo-ec/src/cairo_ec/ec_ops.cairo
zmalatrax: Ok, it's also used in `secp256k1.try_recover_public_key`, but if it's short enough I guess it can be inlined

python/cairo-addons/src/cairo_addons/hints/curve.py
zmalatrax: ohh indeed

zmalatrax (approved):

PR #705, opened 2025-02-07
use ret instead of 0 return

### LINKED ISSUE #704: dev: use low level ret after an assert 0 = 1
## Why

In some places, we need to raise surely with an assert 0 = 1 (branch should not be run) nut still need to return something to comply with the compiler expected return type.

Currently, we build useless objects that are actually never returned, and it impacts falsy the coverage.

## What

Don't build fake object for the sake of making the compiler happy.

## How

Just use a low level `ret`.

Eikix:
closes #704

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -686,6 +686,7 @@ func close_transaction{
         tempvar parent_main_trie_ptr = cast(parent_main_trie, felt);
         if (cast(parent_main_trie_ptr, felt) == 0) {
             assert 0 = 1;
+            ret;
         }
     }
     let parent_trie_start = parent_main_trie.dict_ptr_start;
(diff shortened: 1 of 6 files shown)


ClementWalter (changes requested):
just use when we had to built a return struct or type, but don't add new returns

cairo/src/utils/signature.cairo
```diff
@@ -27,19 +27,19 @@ namespace Signature {
...
             if (x.high == secp256k1.N_HIGH_128) {
                 assert [range_check_ptr] = secp256k1.N_LOW_128 - 1 - x.low;
                 tempvar range_check_ptr = range_check_ptr + 1;
-                return ();
+                ret;
             } else {
                 assert [range_check_ptr] = secp256k1.N_HIGH_128 - 1 - x.high;
                 tempvar range_check_ptr = range_check_ptr + 1;
-                return ();
+                ret;
```
ClementWalter, line 42: ?

Eikix:
> just use when we had to built a return struct or type, but don't add new returns

thanks, done

obatirou (approved):

ClementWalter (approved):

PR #722, opened 2025-02-07
Use raise util

ClementWalter:
- **Add to coverage when runner raises**
- **Move errors util in cairo addons**
- **Add raise util**
- **Use raise and raise_jmp instead of assert**

Solves: #745


python/cairo-core/src/cairo_core/control_flow.cairo
```diff
@@ -0,0 +1,19 @@
...
+//      at the end of the function, which would then not be covered.
+func raise(message: felt) {
+    with_attr error_message("{message}") {
+        [ap] = 0, ap++;
+        [ap] = 1, ap++;
+        dw 0x40127ffe7fff7fff;
+    }
+}
+
+raise_jmp:
```
Eikix, line 13: why do we need both raise and raise_jmp?

cairo/ethereum/cancun/state.cairo
```diff
@@ -685,8 +684,7 @@ func close_transaction{
     with_attr error_message("IndexError") {
         tempvar parent_main_trie_ptr = cast(parent_main_trie, felt);
         if (cast(parent_main_trie_ptr, felt) == 0) {
-            assert 0 = 1;
-            ret;
+            raise('IndexError');
```
Eikix, line 687: what happens in this case? it looks like there are two with_attr error_message nested, the one in line 684 and the one in raise

python/cairo-core/src/cairo_core/control_flow.cairo
ClementWalter: a function needs to end with either a jmp, a ret or a dw. The high level `return` would check for signature consistency, the `ret` would add an uncovered line

cairo/ethereum/cancun/state.cairo
ClementWalter: it's possible to have several nested `with_attr` , but here it's just forgotten to be removed

python/cairo-core/src/cairo_core/control_flow.cairo
```diff
@@ -0,0 +1,19 @@
+// @dev Raise by asserting [ap - 1] == [ap - 2];
+//      The instruction is dw to have the last instruction run failing,
```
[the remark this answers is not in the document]
ClementWalter: this is just the compiled `assert [ap - 1] = [ap - 2]`, see cairovm.gs !

cairo/ethereum/cancun/utils/address.cairo
```diff
@@ -35,8 +36,7 @@ func to_address{range_check_ptr}(data: UnionUintU256) -> Address {
     }
 
     with_attr error_message("Type not valid") {
-        assert 0 = 1;
-        ret;
+        jmp raise.raise_label;
```
enitrat, line 39: why not invoking raise('type not valid')?

ClementWalter: you need to end a function by either a ret or a jmp or a dw

ClementWalter:
note: doing this work I've realized the `map_to_python_exception` util, which is redundant with the `cairo_error` context manager. A follow up PR should be done to remove on of them

enitrat (approved):

python/cairo-core/src/cairo_core/control_flow.cairo
```diff
@@ -0,0 +1,12 @@
...
+//      while using assert [ap - 1] = [ap - 2] would require to add ret;
+//      at the end of the function, which would then not be covered.
+func raise(message: felt) {
+    with_attr error_message("{message}") {
+        raise_label:
+        [ap] = 0, ap++;
+        [ap] = 1, ap++;
+        dw 0x40127ffe7fff7fff;
```
Eikix, line 10: this is assert [ap-1] = [ap-2]?

cairo/ethereum/cancun/utils/address.cairo
Eikix, line 39: why does it count as not covered?

PR #761, opened 2025-02-11
fix: get storage original 

### LINKED ISSUE #760: bug: get_storage_original should get from state's current storage trie if snapshot is null
# EDIT

We need a way in process_transaction to initialize the original_storage_tries to be a copy of the current state.
If we use the existing `trie_get_TrieTupleAddressBytes32U256` then we need a way to finalize this dict at the end of the tx, i.e. at the end of process transaction

# Archived:
in state.cairo, in get_storage_original, we:

```

    let new_original_storage_tries = state.value.original_storage_tries;
    let value = trie_get_TrieTupleAddressBytes32U256{trie=new_original_storage_tries}(address, key);

```

which suggests we assume original_storage_tries is never none. 

in the case of a top most tx, snapshot is null, as such, we must get the storage slot from the current state storage trie.

Eikix:
closes #760


Eikix:
Alternatively, as discussed with @obatirou we could modify `begin_transaction` to set State.original_storage_tries to a copy of the `_storage_tries`.

cc @enitrat

cairo/ethereum/cancun/fork.cairo
```diff
@@ -311,8 +313,20 @@ func process_transaction{
         assert is_valid.value = TRUE;
     }
 
-    // Get sender account
+    // Set original_storage_tries
     let state = env.value.state;
+    let storage_tries = state.value._storage_tries;
+    let original_storage_tries = copy_TrieTupleAddressBytes32U256{trie=storage_tries}();
```
Eikix, line 319: We need a new logic to properly squash / merge this new "read only" trie -> @ClementWalter calls it "rebase" -> when you are at the end of the tx, you need to rebase the final storage trie on the original trie, such that when you squash, all the read operations of original state trie are coherent

cairo/ethereum/cancun/state.cairo
```diff
@@ -616,6 +588,33 @@ func begin_transaction{
...
     local __fp__: felt* = fp_and_pc.fp_val;
 
+    // Set original storage tries if not already set
+    if (cast(state.value.original_storage_tries.value, felt) == 0) {
+        let storage_tries = state.value._storage_tries;
+        // cast in dict access type
+        let dict_ptr_start = cast(storage_tries.value._data.value.dict_ptr_start, DictAccess*);
+        let dict_ptr_end = cast(storage_tries.value._data.value.dict_ptr, DictAccess*);
+        let (new_dict_ptr_start, new_dict_ptr_end) = dict_copy(dict_ptr_start, dict_ptr_end);
+        tempvar original_storage_tries_data = MappingTupleAddressBytes32U256(
```
enitrat, line 598:
```suggestion
        let storage_tries = state.value._storage_tries;
        let (new_dict_ptr_start, new_dict_ptr_end) = dict_copy( cast(storage_tries.value._data.value.dict_ptr_start, DictAccess*), cast(storage_tries.value._data.value.dict_ptr, DictAccess*));
        tempvar original_storage_tries_data = MappingTupleAddressBytes32U256(
```

fmt and lgtm

Issue #731, opened 2025-02-10
dev: compute the storage roots of each address from the Trie[Tuple[Address, Bytes32], U256]


Eikix:
Implement a function that takes as an input our "storage tries" of type Trie[Tuple[Address, Bytes32], U256] and ouputs a Mapping[Address, Hash32]: the storage root for each address.

UPDATE: Timebox to create a strategy / PoC

Strategy:

Create a function `storage_roots` that will:

- Take as input the `State`
- squash the storage tries of type `Trie[Tuple[Address, Bytes32], U256]`
- Create a Mapping[Address, TrieBytes32U256] type
- Iterate over the keys of this trie
  -  Get the tuple (address, bytes32) from the dict_tracker
  - call a function `build_storage_trie_for_address` which takes (address, key, value) and 
     - if the trie doesn't exist in the mapping: allocate a segment with default_dict_new(0);
     - Update it with trie_set(trie, key, value)
- We end up with a complete Mapping[Address, TrieBytes32U256]

Now, squash that mapping.
Create a new Mapping[Address, Bytes32] that will contain the storage roots
Iterate over the keys, s.t. for each key:
- call `root` on the value of type TrieBytes32U256
- squash the trie after having called `root` (last access ever)
- Add the root to the mapping for that address

Return the Mapping[Address, Bytes32]

PR #764, opened 2025-02-11
feat: storage roots

enitrat:
Closes https://github.com/kkrt-labs/keth/issues/731


cairo/ethereum/cancun/trie.cairo
```diff
@@ -1765,6 +1822,35 @@ func _get_bytes32_preimage_for_key{poseidon_ptr: PoseidonBuiltin*}(
...
+    let (local preimage_data: felt*) = alloc();
+    local preimage_len;
+    %{ get_preimage_for_key %}
+
+    // Verify preimage
+    if (preimage_len != 3) {
+        raise('preimage_len != 3');
+    }
+
+    let (preimage_hash) = poseidon_hash_many(preimage_len, preimage_data);
```
ClementWalter, line 1840: if preimage_len needs to be 3, you don't need to populate it in a hint, just use 3.

cairo/ethereum/cancun/state.cairo
```diff
@@ -890,3 +930,212 @@ func destroy_touched_empty_accounts{poseidon_ptr: PoseidonBuiltin*, state: State
...
+    let storage_tries = state.value._storage_tries;
+    let storage_tries_start = cast(storage_tries.value._data.value.dict_ptr_start, DictAccess*);
+    let storage_tries_end = cast(storage_tries.value._data.value.dict_ptr, DictAccess*);
+
+    let (squashed_storage_tries_start, squashed_storage_tries_end) = dict_squash(
+        storage_tries_start, storage_tries_end
+    );
+
+    // Create a Mapping[Address, Trie[Bytes32, U256]] that will contain the "flat" tries, where we
+    // will the storage trie of each address
```
obatirou, line 952:
```suggestion
    // Create a Mapping[Address, Trie[Bytes32, U256]] that will contain the "flat" tries, where we
    // will get the storage trie of each address
```

cairo/ethereum/cancun/trie.cairo
enitrat: the hint is generic and assigns the primage len - as it could be used with preimages of any size (see _get_bytes_preimage_for_key), as such I will assign it in the hint

ClementWalter: the hint would work with `ids.preimage_len` already set to 3

this can just be

```
let preimage_len = 3;
%{ get_preimage_for_key }%
let (preimage_hash) = poseidon_hash_many(preimage_len, preimage_data);
```

ClementWalter (approved):

cairo/ethereum/cancun/trie.cairo
enitrat: done

PR #787, opened 2025-02-12
fix: skip None values in _prepare_trie_inner_storage

enitrat:


enitrat:
we can ignore codecov as this will be covered in apply_body

Eikix (approved):
lgtm

PR #788, opened 2025-02-12
feat: use DictManager's `preimage` field to track preimages of hashed keys

### LINKED ISSUE #795: update logic regarding tracking of preimages
Our current logic is based on iterating through all dict entries to get the preimages of a hashed keys.
Instead, we can create a "preimages: Dict[felt252, Tuple[int, ...]]" that would track the preimage of each key

enitrat:
Closes #795

### DIFF
--- a/cairo/tests/ethereum/cancun/test_trie.py
+++ b/cairo/tests/ethereum/cancun/test_trie.py
@@ -27,7 +27,7 @@ from ethereum.cancun.trie import (
 )
 from ethereum_types.bytes import Bytes, Bytes32
 from ethereum_types.numeric import U256, Uint
-from hypothesis import example, given
+from hypothesis import Verbosity, example, given, settings
 from hypothesis import strategies as st
 
 from cairo_addons.testing.errors import cairo_error, strict_raises
@@ -79,6 +79,7 @@ class TestTrie:
         assert common_prefix_length(a, b) == cairo_run("common_prefix_length", a, b)
 
     @given(a=..., b=...)
+    @settings(verbosity=Verbosity.quiet)
     def test_common_prefix_length_should_fail(
         self, cairo_programs, cairo_run_py, a: Bytes, b: Bytes
     ):
@@ -101,6 +102,7 @@ class TestTrie:
         )
 
     @given(x=nibble.filter(lambda x: len(x) != 0), is_leaf=...)
+    @settings(verbosity=Verbosity.quiet)
     def test_nibble_list_to_compact_should_raise_when_wrong_remainder(
         self, cairo_programs, cairo_run_py, x, is_leaf: bool
     ):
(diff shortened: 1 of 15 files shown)


crates/cairo-addons/src/vm/dict_manager.rs
```diff
@@ -82,6 +85,42 @@ impl PyTrackerMapping {
...
+/// the preimages with manager.preimages[index] = preimage
+#[pyclass(name = "PreimagesMapping", unsendable)]
+pub struct PyPreimagesMapping {
+    inner: Rc<RefCell<RustDictManager>>,
+}
+
+#[pymethods]
+impl PyPreimagesMapping {
+    fn __getitem__(&self, key: PyMaybeRelocatable) -> PyResult<PyDictKey> {
+        Ok(self.inner.borrow().preimages.get(&key.into()).cloned().unwrap().into())
```
obatirou, line 98: maybe use ok_or_else instead of unwrap to return a proper error?

Eikix:
tests fail

enitrat:
> tests fail

can't repro locally 🗡️

cairo/tests/ethereum/cancun/test_trie.py
```diff
@@ -101,6 +102,7 @@ def test_nibble_list_to_compact(self, cairo_run, x, is_leaf: bool):
         )
 
     @given(x=nibble.filter(lambda x: len(x) != 0), is_leaf=...)
+    @settings(verbosity=Verbosity.quiet)
```
Eikix, line 105: why are we using quiet verbosity?

we can directly replace in CI profile in conftest quiet verbosity

crates/cairo-addons/src/vm/dict_manager.rs
```diff
@@ -82,6 +85,51 @@ impl PyTrackerMapping {
     }
 }
 
+// Object returned by DictManager.preimages enabling access to the preimages by index and mutating
+/// the preimages with manager.preimages[index] = preimage
+#[pyclass(name = "PreimagesMapping", unsendable)]
+pub struct PyPreimagesMapping {
+    inner: Rc<RefCell<RustDictManager>>,
```
Eikix, line 92: Why are we manipulating this object?

cairo/tests/ethereum/cancun/test_trie.py
enitrat: quiet for those that contain the cairo_program fixture only, so that the full program is not displayed even in verbose settings (this is what causes the logs to be bloated in CI)

crates/cairo-addons/src/vm/dict_manager.rs
enitrat: the preimages are a field of the RustDictManager

Eikix (approved):

PR #790, opened 2025-02-12
chore: update readme

slumondai:
Just fixing a probable typo

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ dependencies and run commands. To install uv:
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
-Moreover, the project uses [rust](https://www.rust-lang.org/) to manage rust
+Moreover, the project uses [cargo](https://github.com/rust-lang/cargo) to manage rust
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,7 @@ dependencies and run commands. To install uv:
 curl -LsSf https://astral.sh/uv/install.sh | sh
 ```
 
-Moreover, the project uses [rust](https://www.rust-lang.org/) to manage rust
+Moreover, the project uses [cargo](https://github.com/rust-lang/cargo) to manage rust
```
Eikix, line 29: thanks for your help!
this makes me think that this line altogether does not convey valuable information, since cargo comes automatically with the rust binary.
we should simply say that we use rust in parts of the project and as such, rust must be installed

slumondai: Hello, it is a bit soon to call this help. 
I notice that [lib GMP](https://gmplib.org/) must be installed to run the tests on a mac, it could be more important to include this in the readme. 
Closing the PR for now

PR #791, opened 2025-02-12
fix: hashdict_read hint should not skip zero values

enitrat:
The problem is that `if value` would return `False` if `value := 0`. But if we delete the entry in a trie (by writing 0), then the next read on the same (address, key) should be returning `0` as `prev_value`, not `default_factory()`. This caused an issue when squashing the storage_tries when computing the storage roots.


Eikix (approved):
lgtm

Eikix (approved):
re: lgtm

PR #792, opened 2025-02-12
fix: iterate on squashed storage tries in storage roots computation

enitrat:
The iteration was done on the the unsquashed tries, meaning that multiple entries would appear for the same address.


Eikix (approved):
lgtm

PR #807, opened 2025-02-13
dev: update pyo3 to 0.23

### LINKED ISSUE #802: dev: migrate to PyO3 0.23.x
To use the garaga-rs hints, we need to update PyO3. This will come with require some refactoring because 0.22 -> 0.23 had a[ lot of API changes.](https://pyo3.rs/v0.23.3/migration.html?highlight=intopy#migrating-from-older-pyo3-versions)

enitrat:
The 0.23 update has some incompabilities with 0.22. 
Because Garaga requires pyo3 ^0.23, I updated our dependency, which required migrating some traits.

Closes #802


python/cairo-addons/tests/test_felt.py
```diff
@@ -1,60 +0,0 @@
-from hypothesis import given
-from hypothesis import strategies as st
-from starkware.cairo.lang.cairo_constants import DEFAULT_PRIME
-
-from cairo_addons.vm import Felt
-
-felt = st.integers(min_value=0, max_value=DEFAULT_PRIME - 1)
-
-
-class TestFelt:
```
Eikix, line 10: why are we deleting it

python/cairo-addons/tests/test_memory_segments.py
```diff
@@ -22,14 +22,6 @@ def test_add_temporary_segment(self, runner):
         assert ptr.segment_index == -1
         assert ptr.offset == 0
 
-    def test_load_data_felt(self, runner):
```
Eikix, line 25: why are we deleting it

Eikix (approved):
lgtm

PR #809, opened 2025-02-14
fix: use default_dict_finalize where necessary

### LINKED ISSUE #798: replace "squash_dict" by "default_dict_finalize" for default dicts
# Why

We use default dicts for various operations. While `squash_dict` ensures consistency in dict accesses, it does not ensure that the first access's value was coherent with the default value of the dict.

# What

Replace squashes of default dict with a call to `default_dict_finalize`. Similar to `squash_dict`, you will create a wrapper function that will also assign the dict_tracker of the input segment to the squashed segment.

Locate all default_dict that use pointers to values. We will need to keep these pointers accessible in order to finalize with the right pointer.

Normally - the only case of this happening is for U256Struct*, the default value being u256(0,0). For that we can define 
inside a constants file a 
```
U256_ZERO:
dw 0;
dw 0;
```
then `let ptr = get_label_location(U256_ZERO)`
and `cast(ptr, U256Struct*)`

obatirou:
Close #798

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -79,12 +79,16 @@ from legacy.utils.dict import (
     get_keys_for_address_prefix,
     dict_update,
     dict_copy,
-    dict_squash,
+    default_dict_finalize,
 )
 
 EMPTY_ROOT:
 dw 0x6ef8c092e64583ffa655cc1b171fe856;  // low
 dw 0x21b463e3b52f6201c0ad6c991be0485b;  // high
+
+U256_ZERO:
+dw 0;
+dw 0;
 
 struct AddressTrieBytes32U256DictAccess {
     key: Address,
@@ -937,7 +941,7 @@ func destroy_touched_empty_accounts{poseidon_ptr: PoseidonBuiltin*, state: State
 }
 
 func empty_transient_storage{range_check_ptr}() -> TransientStorage {
-    tempvar default_value = new U256Struct(0, 0);
+    let (default_value) = get_label_location(U256_ZERO);
     let (dict_ptr) = default_dict_new(cast(default_value, felt));
     let dict_start = cast(dict_ptr, TupleAddressBytes32U256DictAccess*);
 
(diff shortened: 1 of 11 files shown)


cairo/tests/utils/args_gen.py
```diff
@@ -1156,6 +1156,8 @@ def default_factory():
             return default_value
 
         data = defaultdict(default_factory, data)
+    else:
+        default_value = 0
```
enitrat, line 1160: this is not correct - we can't assume that the default value is actually zero. for a trie[bytes32, u256] the default value needs to pointer to u256(0,0), and because the trie strategies are not generating default dicts, this will cause problems

cairo/tests/utils/strategies.py
```diff
@@ -174,6 +174,7 @@ def non_default_strategy(default):
                 non_default_strategy(default),
                 min_size=min_size,
                 max_size=15,
+                dict_class=defaultdict,
```
enitrat, line 177: what's the default_factory for this default_dict ? does it need to be specified?

cairo/ethereum/cancun/state.cairo
```diff
@@ -977,8 +981,8 @@ func storage_roots{
     let storage_tries_start = cast(storage_tries.value._data.value.dict_ptr_start, DictAccess*);
     let storage_tries_end = cast(storage_tries.value._data.value.dict_ptr, DictAccess*);
 
-    let (squashed_storage_tries_start, squashed_storage_tries_end) = dict_squash(
-        storage_tries_start, storage_tries_end
+    let (squashed_storage_tries_start, squashed_storage_tries_end) = default_dict_finalize(
+        storage_tries_start, storage_tries_end, 0
```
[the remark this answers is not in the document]
enitrat:
```suggestion
    let (squashed_storage_tries_start, squashed_storage_tries_end) = default_dict_finalize(
        storage_tries_start, storage_tries_end, cast(storage_tries.value.default.value, felt)
```

enitrat (changes requested):

cairo/tests/utils/args_gen.py
enitrat: if anything we probably need to assert that the dict passed is a default dict

cairo/tests/utils/strategies.py
obatirou: good point

cairo/tests/utils/args_gen.py
obatirou: the fix is wrong and was done to rapidly 
lookin into it now

enitrat (approved):
💯

Issue #774, opened 2025-02-12
dev: test_runtime.cairo does not properly test behaviour of our jump destination


Eikix:
# WHY

[apply_body](https://github.com/kkrt-labs/keth/pull/769/commits/5d6a6e0337110474815dea9af76c664eea545a12) feature uncovered a problem in jumpdest -> `initialize_jumpdest` creates a dict of type `Dict[int, {0,1}]` instead of `Dict[tuple(int), {0,1}]` as per our codebase rules on preimages.
this means that current tests don't cover jumpdest logic

# WHAT

test_runtime.cairo

# HOW

proposal: modify the test to reproduce behaviour of jumps, i.e. dict_read operations on the valid_jumpdest dictionary

Eikix:
Strategy:
- `test__get_valid_jump_destinations` -> delete
- in `test__finalize_jumpdests`, call get_valid_jumpdests, and replace hint by successive reads (pass a SetUint in test__finalize_jumpdests as argument and recursively dict_read), and lastly call finalize
- in `test__assert_valid_jumpdest`, first call get_valid_jumpdest, pass `Code: Bytes` and a unique jumpdest as argument

ClementWalter:
is this still valid?

enitrat:
Hashdicts handle buth Tuple[int] and int as keys

PR #810, opened 2025-02-14
feat: garaga rs hints

### LINKED ISSUE #617: dev: migrate garaga hints from try_recover_public_key to rust
utils in secp256k1.cairo in `cairo_ec` use garaga hints for sig verification
migrate them to rust to increase run speed

enitrat:
Closes #617


python/cairo-addons/src/cairo_addons/hints/utils.py
```diff
@@ -136,3 +136,30 @@ def initialize_jumpdests(
...
+        for i in range(min(ids.valid_jumpdest.key, 32))
+    ]
+    # Check if any PUSH may prevent this to be a JUMPDEST
+    ids.is_no_push_case = int(
+        not any([0x60 + i <= byte <= 0x7F for i, byte in enumerate(bytecode)])
+    )
+
+
+@register_hint
+def continue_general_case(
```
Eikix, line 155: i suggest adding jumpdest_ as a prefix to this function name for clarity

cairo/tests/legacy/utils/test_bytes_legacy.py
```diff
@@ -33,7 +33,7 @@ def test_should_raise_when_value_sup_31_bytes(self, cairo_run, n):
         @given(n=integers(min_value=256, max_value=2**248 - 1))
         @settings(verbosity=Verbosity.quiet)
         def test_should_raise_when_byte_value_not_modulo_base(
-            self, cairo_programs, cairo_run, n
+            self, cairo_programs, cairo_run_py, n
```
Eikix, line 36: why are we using cairo_run_py here?

python/cairo-ec/src/cairo_ec/circuit_utils.cairo
```diff
@@ -6,12 +6,13 @@ func hash_full_transcript{poseidon_ptr: PoseidonBuiltin*}(limbs_ptr: felt*, n: f
...
-    let elements_end = &limbs_ptr[n * N_LIMBS];
+    local elements_end: felt* = limbs_ptr + (n * N_LIMBS);
 
     tempvar elements = limbs_ptr;
     tempvar pos_ptr = cast(poseidon_ptr, felt*);
 
     loop:
+    let N_LIMBS_ = N_LIMBS;
     tempvar has_six_uint384_remaining;
     %{ has_six_uint384_remaining_hint %}
```
Eikix, line 17: i suggest against adding `_hint` suffix in a hint's name

cairo/tests/legacy/utils/test_bytes_legacy.py
enitrat: we can't patch_hint with the rust vm

cairo/tests/ethereum/cancun/vm/test_runtime.py
```diff
@@ -7,13 +7,15 @@
 from cairo_addons.testing.errors import cairo_error
 from tests.utils.solidity import get_contract
 
+pytestmark = pytest.mark.python_vm
```
obatirou, line 10: What is missing to pass it to rust?
Could we make a follow up issue

Same where relevant

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
```diff
@@ -230,6 +232,8 @@ func try_recover_public_key{
     tempvar range_check96_ptr_init = range_check96_ptr;
     tempvar range_check96_ptr_after_circuit = range_check96_ptr + 224 + (4 + 117 + 108 - 1) *
         N_LIMBS;
+    let x = cast(poseidon_ptr, felt*) - 3;
+    // %{print_maybe_relocatable_hint%}
```
Eikix, line 236: delete these unused lines

cairo/tests/ethereum/cancun/vm/test_runtime.py
enitrat: `test_runtime.cairo` is full of hints in tests for now, this will be covered by #774

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -0,0 +1,268 @@
...
+
+pub const HINTS: &[fn() -> Hint] = &[
+    build_msm_hints_and_fill_memory,
+    compute_y_from_x_hint,
+    fill_add_mod_mul_mod_builtin_batch_one,
+    decompose_scalar_to_neg3_base,
+    fill_add_mod_mul_mod_builtin_batch_117_108,
+];
+
+pub fn build_msm_hints_and_fill_memory() -> Hint {
```
obatirou, line 45: Add documentation saying it is to builds the Multi-Scalar Multiplication (MSM) hints and fills memory with SECP256K1 curve point data. 
What is RLC?

Nit: it could be broke down in smaller function for an easier review.

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -0,0 +1,268 @@
...
+                vm,
+                ids_data,
+                ap_tracking,
+            )?;
+            Ok(())
+        },
+    )
+}
+
+// Adapted from sympy implementation
```
obatirou, line 189: which implementation exactly ? What commit ?

crates/cairo-addons/src/vm/hint_definitions/maths.rs
```diff
@@ -98,7 +105,23 @@ pub fn felt252_to_bytes_be() -> Hint {
...
+    Hint::new(
+        String::from("is_positive_hint"),
+        |vm: &mut VirtualMachine,
+         _exec_scopes: &mut ExecutionScopes,
+         ids_data: &HashMap<String, HintReference>,
+         ap_tracking: &ApTracking,
+         _constants: &HashMap<String, Felt252>|
+         -> Result<(), HintError> {
+            let value = get_integer_from_var_name("value", vm, ids_data, ap_tracking)?;
+            let is_positive = if value <= Felt252::from(u128::MAX) { 1 } else { 0 };
```
obatirou, line 118: I am not sure about this 
* `DEFAULT_PRIME = 2**251 + 17 * 2**192 + 1`
* `u128 max = 2**128 - 1`
```
// Returns the sign of value: -1 if value < 0, 1 if value > 0.
// value is considered positive if it is in [0, STARK//2[
// value is considered negative if it is in ]STARK//2, STARK[
// If value == 0, returned value can be either 0 or 1 (undetermined).
func sign{range_check_ptr}(value) -> felt {
    const STARK_DIV_2_PLUS_ONE = (-1) / 2 + 1;  // == prime//2 + 1
    const STARK_DIV_2_MIN_ONE = (-1) / 2 - 1;  // == prime//2 - 1
    tempvar is_positive: felt;
    %{ is_positive_hint %}
    if (is_positive != 0) {
        assert_le_felt(value, STARK_DIV_2_MIN_ONE);
        return 1;
    } else {
        assert_le_felt(STARK_DIV_2_PLUS_ONE, value);
        return -1;
    }
}
```

obatirou: A differential fuzz testing between python and rust implementation for all hints could be beneficial.
Current tests of higher level function might not catch those kind of differences

python/cairo-ec/src/cairo_ec/circuit_utils.cairo
```diff
@@ -6,12 +6,13 @@ func hash_full_transcript{poseidon_ptr: PoseidonBuiltin*}(limbs_ptr: felt*, n: f
...
     local BASE = 2 ** 96;
 
-    let elements_end = &limbs_ptr[n * N_LIMBS];
+    local elements_end: felt* = limbs_ptr + (n * N_LIMBS);
 
     tempvar elements = limbs_ptr;
     tempvar pos_ptr = cast(poseidon_ptr, felt*);
 
     loop:
+    let N_LIMBS_HINT = N_LIMBS;
```
obatirou, line 15: This is only for the rust hint?
You cannot access N_LIMBS directly ?

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
obatirou: print is still here

crates/cairo-addons/src/vm/hint_definitions/maths.rs
Eikix: > A differential fuzz testing between python and rust implementation for all hints could be beneficial. Current tests of higher level function might not catch those kind of differences. Could be done in a follow up issue

We talked about it, it's a bit involved / time consuming but it could be useful yes! I wonder how easy it is to do
Down the line, we technically won't be using the cairoVM py (since there'll only be one M31 cairo runner)!

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -0,0 +1,268 @@
...
+            let b = Uint384::from_var_name("b", vm, ids_data, ap_tracking)?.pack();
+            let p = Uint384::from_var_name("p", vm, ids_data, ap_tracking)?.pack();
+            let g = Uint384::from_var_name("g", vm, ids_data, ap_tracking)?.pack();
+            let x = Uint384::from_var_name("x", vm, ids_data, ap_tracking)?.pack();
+            let v = get_integer_from_var_name("v", vm, ids_data, ap_tracking)?.to_biguint();
+
+            let rhs = (pow(x.clone(), 3) + a * x + b) % p.clone();
+
+            // Currently, only the secp256k1 field is supported
+            let (rhs_felt, g_felt) = if p.to_str_radix(16).to_uppercase() ==
```
obatirou, line 145: this check is not present in python

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -0,0 +1,268 @@
...
+                let sqrt_felt = rhs_felt.sqrt().unwrap().0;
+                let has_same_parity = (v % 2_u32) == (element_to_biguint(&sqrt_felt) % 2_u32);
+                if has_same_parity {
+                    sqrt_felt
+                } else {
+                    -sqrt_felt
+                }
+            } else {
+                (rhs_felt * g_felt).sqrt().unwrap().0
+            };
```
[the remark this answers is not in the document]
obatirou: I think the handling of errors could be done with map_err and send back a better message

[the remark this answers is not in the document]
enitrat: to be honest it's easier to panic with an unwrap failed (as the call stack unfolds and gives you the precise line where the error occured) compared to a more obscure HintError. Thus I won't be modifying this

crates/cairo-addons/src/vm/hint_definitions/curve.rs
enitrat: We can only support secp256k1 in rust

crates/cairo-addons/src/vm/hint_definitions/maths.rs
enitrat: you're right, this should be DEFAULT_PRIME / 2

crates/cairo-addons/src/vm/hint_definitions/curve.rs
obatirou: try_get_point_from_x is generalist but can only be used for secp256k1. Maybe a naming problem ?
Why pass the arguments if the curve is already be known. The python version should also add the check maybe?

Note try_get_point_from_x has still an important todo: Add a check for v

python/cairo-ec/src/cairo_ec/circuit_utils.cairo
obatirou: An idea why ?
Should an issue be created ?

crates/cairo-addons/src/vm/hint_definitions/curve.rs
enitrat: > Why pass the arguments if the curve is already be known. The python version should also add the check maybe?

The current implementation relying on lambdaworks cryptography requires the `F` Prime order to be known at compile time. The python limitation does not have this limitation. Considering that this is only ever called with the SECP256R1_PRIME_FIELD_ORDER I don't want to make this behavior generic and waste resources on unused code

python/cairo-ec/src/cairo_ec/circuit_utils.cairo
enitrat: the way variables are accessed is a bit different, I don't think it's worth tracking as it's non-blocking and minor

crates/cairo-addons/src/vm/hint_definitions/maths.rs
```diff
@@ -98,7 +105,24 @@ pub fn felt252_to_bytes_be() -> Hint {
...
+        String::from("is_positive_hint"),
+        |vm: &mut VirtualMachine,
+         _exec_scopes: &mut ExecutionScopes,
+         ids_data: &HashMap<String, HintReference>,
+         ap_tracking: &ApTracking,
+         _constants: &HashMap<String, Felt252>|
+         -> Result<(), HintError> {
+            let value = get_integer_from_var_name("value", vm, ids_data, ap_tracking)?;
+            let is_positive =
+                if value <= Felt252::from(Felt252::MAX.to_biguint() / 2_u32) { 1 } else { 0 };
```
obatirou, line 119:
```suggestion
           let is_positive = (value <= Felt252::from(Felt252::MAX.to_biguint() / 2_u32)) as i32;
```

obatirou: should work

crates/cairo-addons/src/vm/hint_definitions/curve.rs
enitrat: sorry my bad, K1 not R1 - indeed

obatirou (approved):

PR #820, opened 2025-02-17
dev: secpk1 endianness

### LINKED ISSUE #708: dev: refactor secpk1 recover to not use low level endianness
# why

we want to encapsulate endianness in utils

# what

refactor  secpk1_recover_big_ends to use type conversions U256_to_bytes_be or le

# how
- in secp256k1_recover_uint256_bigends, remove endianness assumption and return Hash32 type for x and y
- in public_key_point_to_eth_address_be, use keccak little endian 
- refrain from using low level utils to reverse endianness of a word in the codebase, encapsulate these assumptions in test utils

Eikix:
closes #708 

what this PR does:
- refactor numeric into cairo_core
- refactor bytes into cairo_core such that secpk1 util in cairo curve does not import types from ethereum (circular deps)
- change types from secp recover to match codebase style


enitrat (changes requested):
- refactor numeric into cairo_core
- refactor bytes into cairo_core such that secpk1 util in cairo curve does not import types from ethereum (circular deps)

Is this desired? they're ethereum specific types. This seems out of scope for this PR; I would revert as this is a big; unrelated change.

Eikix:
> * refactor numeric into cairo_core
> * refactor bytes into cairo_core such that secpk1 util in cairo curve does not import types from ethereum (circular deps)
> 
> Is this desired? they're ethereum specific types. This seems out of scope for this PR; I would revert as this is a big; unrelated change.

TL:DR:
- @ClementWalter mentioned that our try_recover_public_key should not "bear" endianness details, which suggests that we shouldn't manipulate Uint256 but rather Bytes32. This in turn forces me to import Bytes32 in cairo_curve, which is a circular dependency.
- I thought we could migrate types in cairo_core, such that now our utils can use our types -> U256 and Bytes32

Eikix:
> > * refactor numeric into cairo_core
> > * refactor bytes into cairo_core such that secpk1 util in cairo curve does not import types from ethereum (circular deps)
> > 
> > Is this desired? they're ethereum specific types. This seems out of scope for this PR; I would revert as this is a big; unrelated change.
> 
> TL:DR:
> 
> * @ClementWalter mentioned that our try_recover_public_key should not "bear" endianness details, which suggests that we shouldn't manipulate Uint256 but rather Bytes32. This in turn forces me to import Bytes32 in cairo_curve, which is a circular dependency.
> * I thought we could migrate types in cairo_core, such that now our utils can use our types -> U256 and Bytes32

Looking back I have probably fixed circular dependencies by migrating `public_key_point_to_eth_address` to ethereum.crypto but still ugly to import Bytes32 from ethereum into cairo_core

Eikix:
> can you split in 2 prs easily?

sure, but i split already into 1 atomic commit, what is the purpose of splitting the PRs?

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
```diff
@@ -321,23 +327,17 @@ func try_recover_public_key{
...
-    assert elements[0] = x;
-    assert elements[1] = y;
-    let (point_hash: Uint256) = keccak_uint256s_bigend(n_elements=2, elements=elements);
-
-    // The Ethereum address is the 20 least significant bytes of the keccak of the public key.
-    let (_, high_low) = unsigned_div_rem(point_hash.high, 2 ** 32);
-    let eth_address = point_hash.low + RC_BOUND * high_low;
-    return eth_address;
+
+    let max_value = Uint256(secp256k1.P_LOW_128 - 1, secp256k1.P_HIGH_128);
```
enitrat, line 323: what's this check for?

cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -46,29 +48,28 @@ func secp256k1_recover_uint256_bigends{
...
 func public_key_point_to_eth_address{
     range_check_ptr, bitwise_ptr: BitwiseBuiltin*, keccak_ptr: KeccakBuiltin*
-}(x: U256, y: U256) -> Address {
-    let eth_address_felt_be = public_key_point_to_eth_address_be(x=[x.value], y=[y.value]);
-    let eth_address = Address_from_felt_be(eth_address_felt_be);
-    return eth_address;
+}(public_key: Bytes) -> Address {
+    alloc_locals;
+    let point_hash = keccak256(public_key);
+    let (low_high, _) = unsigned_div_rem(point_hash.value.low, 2 ** 96);
```
Eikix, line 71: I have to come out clean on this one, I don't understand why it works

keccak256 returns a Hash32, which wraps Uint256.
I don't know how to get the least significant 20 bytes of that Uint256 using arithmetic. I tried many ways to turn it around, and this way works.
Maybe when my brain is less fried it'll become clear.

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
```diff
@@ -141,9 +141,10 @@ func try_recover_public_key{
 
     let (y, is_on_curve) = try_get_point_from_x(x=&r, v=y_parity, a=&a, b=&b, g=&g, p=&p);
     if (is_on_curve == 0) {
-        tempvar public_key_x = Bytes32(new Bytes32Struct(0, 0));
-        tempvar public_key_y = Bytes32(new Bytes32Struct(0, 0));
-        return (public_key_x=public_key_x, public_key_y=public_key_y, success=0);
+        let (data) = alloc();
+        memset(data, 0, 64);
+        tempvar public_key = Bytes(new BytesStruct(data, 64));
+        return (public_key=public_key, success=0);
```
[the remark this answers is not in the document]
Eikix: to match the type of EELS, it's a small price to pay in terms of unoptimized perf since we do Uint256 -> Bytes -> U64 words array

instead of Uint256 -> u64 words array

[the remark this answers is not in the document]
Eikix: I still have trouble wrapping my head around a Uint256 that represents a little endian bytearray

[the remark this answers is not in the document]
Eikix: maybe it's not such a small price

cairo/ethereum/crypto/elliptic_curve.cairo
enitrat: I hope this helps

```
from ethereum_types.numeric import U256
from ethereum.cancun.fork_types import Address
from ethereum.crypto.hash import keccak256

point_hash = keccak256(b"hello")
print(f"0x{point_hash.hex()}") # 0x1c8aff950685c2ed4bc3174f3472287b56d9517b9c948127319a09a7a36deac8

# What you have in Cairo
point_hash_int = U256.from_le_bytes(point_hash)
point_hash_low = point_hash_int._number & 2**128-1
point_hash_high = point_hash_int._number >> 128

print(f"0x{hex(point_hash_low)}") # 0x0x7b2872344f17c34bedc2850695ff8a1c
print(f"0x{hex(point_hash_high)}") # 0x0xc8ea6da3a7099a312781949c7b51d956

# You want to take the last 20 bytes of the public key from its u256 little-endian representation
# = Take the upper 4 bytes from the low [12:16] and 16 bytes from the high [0:16]
low_high = point_hash_low // 2**96 # 4 last bytes bytes
print(f"0x{hex(low_high)}") # 0x0x7b287234

# Assemble them into a 20-byte value
eth_address_int = point_hash_high << 32 | low_high
print(f"0x{hex(eth_address_int)}") # 0x0xc8ea6da3a7099a312781949c7b51d956 | 7b287234

# Serialize them back into bytes
serialized = eth_address_int.to_bytes(20, "little")

assert Address(serialized) == Address(point_hash[12:32])


```

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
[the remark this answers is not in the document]
enitrat: This is not a good intuition to have, converting from u256 -> bytes and then packing the bytes has a price, we know that byte conversion is one of the most expensive operation we do in the codebase, this should be motivated by numbers instead.

I ran the profiling, this is a ~30% increase 7800 to 10000 steps, and I personally think it would be fine to just return a Bytes32

enitrat (changes requested):

python/cairo-ec/src/cairo_ec/curve/secp256k1.cairo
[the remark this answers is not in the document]
Eikix: Ok will revert to Bytes32, thanks for taking the time to profile

cairo/ethereum/crypto/elliptic_curve.cairo
Eikix: Maybe the part I didn't get is that in EELS it's said, 'take the 20 least significant bytes of the hash', but here we take the 20 most significant bytes, and I was confused

Eikix: Anyway; it helps:)

Eikix: Also why do high and low look big endian in their hex representation?

Eikix: gpt mentions:
"Even though they came from a little-endian interpretation of the hash, they are stored as standard Python integers, which always print in big-endian order."

hex strings are a bit confusing to me

enitrat: >  'take the 20 least significant bytes of the hash', but here we take the 20 most significant bytes, and I was confused

The least significant bytes are on the right (end)

cairo/ethereum/crypto/elliptic_curve.cairo
```diff
@@ -46,29 +47,33 @@ func secp256k1_recover_uint256_bigends{
...
-    tempvar x = U256(new U256Struct(x_uint256.low, x_uint256.high));
-    tempvar y = U256(new U256Struct(y_uint256.low, y_uint256.high));
-    return (x=x, y=y);
+    return (public_key_point_x, public_key_point_y);
 }
 
+// @notice Converts a public key point to the corresponding Ethereum address.
+// @param x The x coordinate of the public key point.
+// @param y The y coordinate of the public key point.
+// @return The Ethereum address, interpreted as a 20-byte big-endian value.
```
enitrat, line 64:
```suggestion
// @return The Ethereum address, interpreted as a 20-byte big-endian value.
```

big-endian ?

Eikix:
```suggestion
// @return The Ethereum address, interpreted as a 20-byte little-endian value.
```

enitrat (approved):

PR #821, opened 2025-02-17
temp/test-ci

enitrat:
dont merge


Eikix:
update: @obatirou is investigating locally a potential culprit: `check_transaction`

PR #823, opened 2025-02-18
fix: nightly fail

obatirou:
This PR is to reproduce the nightly fail that takes too much time.


obatirou:
Check already done with 
```
let len = tx.value.blob_transaction.value.blob_versioned_hashes.value.len;
        with_attr error_message("InvalidBlock") {
            assert_not_zero(len);
        }
```
So len of blob_versioned_hashes is never 0 when entering _check_versioned_hashes_version

PR #845, opened 2025-02-24
dev: add ef-tests to CI

### LINKED ISSUE #837: dev: run ef-tests in CI / nightly
Once all precompiles are implemented (even if not sound) we should be able to run all EF-Tests.

Add CI jobs to run them in CI (sampled, max 2k tests, so that runtime is <=10min) and in nightly (all of them)

update: there seems to be a big amount of memory used when running these tests in parallel. we should try to validate if this is a memory leak issue or not, and maybe run the tests sequentially in the meanwhile.

enitrat:
Closes #837


.github/workflows/nightly_fuzzing.yml
```diff
@@ -20,47 +20,72 @@ jobs:
...
+              "text": ":rotating_light: *Keth Nightly Unit Tests Failed!*: <@${{ secrets.SLACK_USER_1 }}> <@${{ secrets.SLACK_USER_2 }}>\n"
             }
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      - name: Fail workflow if tests failed
+        if: steps.unit-tests.outcome == 'failure'
+        run: exit 1
 
+  ef-tests-full:
+    runs-on: ubuntu-latest-16-cores
```
obatirou, line 61: use 64 cores to avoid long running one to block other tests

.github/workflows/nightly_fuzzing.yml
```diff
@@ -20,47 +20,72 @@ jobs:
...
+    env:
+      HYPOTHESIS_PROFILE: nightly
```
obatirou, line 63: is this needed ?

.github/workflows/nightly_fuzzing.yml
```diff
@@ -20,47 +20,72 @@ jobs:
...
+      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v2
+        with:
+          enable-cache: true
```
obatirou, line 69: should this be enabled ?

enitrat: I don't see a reason not to reuse the uv cache

.github/workflows/nightly_fuzzing.yml
obatirou: It should not

.github/workflows/nightly_fuzzing.yml
obatirou: oh you right, it thought it was the pytest cache

.github/workflows/python_tests.yml
```diff
@@ -47,19 +42,39 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: astral-sh/setup-uv@v3
+        with:
+          enable-cache: true
+          cache-dependency-glob: uv.lock
+      - uses: actions/setup-python@v5
+        with:
+          python-version-file: .python-version
+      - name: Run sampled EF tests
```
enitrat, line 76: should we add codecov as well?

cairo/tests/ef_tests/cancun/test_state_transition.py
```diff
@@ -13,7 +13,10 @@
     run_blockchain_st_test,
 )
 
-pytestmark = pytest.mark.cairo_file(f"{Path().cwd()}/cairo/ethereum/cancun/fork.cairo")
+pytestmark = [
+    pytest.mark.cairo_file(f"{Path().cwd()}/cairo/ethereum/cancun/fork.cairo"),
+    pytest.mark.max_steps(100_000_000),
```
enitrat, line 18: max 100M steps per ef-test run: anything above is extremely likely to be >1B anyway, crashing the runner

.github/workflows/nightly_fuzzing.yml
```diff
@@ -52,19 +47,50 @@ jobs:
...
+          cache-dependency-glob: uv.lock
+      - uses: actions/setup-python@v5
+        with:
+          python-version-file: .python-version
+      - name: Run full EF tests
+        id: ef-tests
+        continue-on-error: true
+        # Running on 48 cores to avoid out of memory issues
+        run: |
+          uv run --reinstall pytest -n 48 -m "not slow" --durations=0 -v -s --log-cli-level=DEBUG cairo/tests/ef_tests/ --ignore-glob='cairo/tests/ef_tests/fixtures/*'
```
obatirou, line 83: are some ef_tests marked slow ? 
`-m "not slow"  `

.github/workflows/python_tests.yml
obatirou: If it takes car of the max 2000 and radomization of tests it would be great but seems tricky.

obatirou (dismissed):

PR #859, opened 2025-02-25
dev: optimize access addresses / keys

### LINKED ISSUE #569: dev: "fork" accessed_addresses and accessed_storage_keys from their parents when creating child_evm
in generic_call or generic_create, we create new, empty dicts for the accessed storage keys of a child context. we use the tracker of the previous segment as the initial_dict, and copy all the entries of the dict access segment.

we could simply "fork" it from the parent dict with the original_mapping mechanism.

Definition of done:
- delete dict_copy logic
- update `incorporate_child_on_success` to use `dict_update` to merge the values with the parent dict (instead of `squash_and_update`, which is a function for dicts with unrelated histories)
- in `generic_call`, `generic_create`, don't use `dict_copy`, but do as in `copy_TrieAddressOptionalAccount` with `%{copy_dict_segment}`

### LINKED ISSUE #863: `dict_copy` and `copy_segment` hints are duplicated
We only need a single hint that takes the tracker of dict A, copies it, and assigns it to dict B. 
Delete one of them.
Update the cairo code accordingly.

enitrat:
Closes #569 
Closes #863

### DIFF
--- a/cairo/legacy/utils/dict.cairo
+++ b/cairo/legacy/utils/dict.cairo
@@ -67,7 +67,7 @@ func dict_copy{range_check_ptr}(dict_start: DictAccess*, dict_end: DictAccess*) 
     // Update the DictTracker's current_ptr to point to the end of the copied dict.
     let current_tracker_ptr = new_dict_ptr;
     let new_tracker_ptr = new_end;
-    %{update_dict_tracker%}
+    %{ update_dict_tracker %}
     return (new_dict_ptr, new_end);
 }
 


cairo/tests/ethereum/cancun/vm/instructions/test_system.py
```diff
@@ -244,7 +244,7 @@ def test_create2(self, cairo_run, evm: Evm):
     )
     def test_generic_call(
         self,
-        cairo_run,
+        cairo_run_py,
```
obatirou, line 247: why py?

cairo/ethereum/cancun/vm/instructions/system.cairo
```diff
@@ -134,25 +134,27 @@ func generic_call{
...
-    let (accessed_addresses_copy_start, accessed_addresses_copy) = dict_copy(
-        cast(evm.value.accessed_addresses.value.dict_ptr_start, DictAccess*),
-        cast(evm.value.accessed_addresses.value.dict_ptr, DictAccess*),
-    );
-    let (accessed_storage_keys_copy_start, accessed_storage_keys_copy) = dict_copy(
-        cast(evm.value.accessed_storage_keys.value.dict_ptr_start, DictAccess*),
-        cast(evm.value.accessed_storage_keys.value.dict_ptr, DictAccess*),
-    );
+    // Fork the accessed_addresses dict segment
+    local new_dict_ptr: DictAccess*;
```
obatirou, line 138: the comment said `this could be optimized using a non-copy mechanism.`
There is a copy of the dict segment.
dict_copy use a memcpy and update the tracker in the hint and the copy_dict_segment only update the tracker, right?
So the cairo memory is not copied. Is this what it is expected?

enitrat: > dict_copy use a memcpy and update the tracker in the hint and the copy_dict_segment only update the tracker, right?
So the cairo memory is not copied. Is this what it is expected?

Yes, exactly - this is the expected behavior. The segment is not copied. But the naming is terrible and confusing. I'll update it.

enitrat: Done. removed the dict_copy hint as per rationale in https://github.com/kkrt-labs/keth/issues/863. Modified the name of the remaining hint for more clarity

cairo/legacy/utils/dict.cairo
```diff
@@ -50,16 +50,21 @@ func dict_write{dict_ptr: DictAccess*}(key_: felt, new_value: felt) {
     return ();
 }
 
+// @notice Copies a dict segment from the dict_start to the new_end.
```
obatirou, line 53:
```suggestion
// @notice Copies a dict segment from dict_start
```

obatirou (dismissed):

obatirou (approved):

PR #886, opened 2025-02-27
feat: add ec_mul 

### LINKED ISSUE #178: feat: ec_mul
## Why

Missing currently ec_mul operation in `ec_ops` while there is an ec_mul precompile.

## What

Add the ec_mul operation, following the ecip_1P circuit (no double and add algo).

## How

tbd

### LINKED ISSUE #853: feat: handle scalars < 2**128 in ecip circuits
Currently, only scalars above $2^128$ can be used in ECIP circuits.

While it is a somehow negligible issue for `ec_recover`, in the context of `ec_mul`, it should be handled

### LINKED ISSUE #854: feat: handle scalars above group order in ec_mul precompile
According to [EIP-196](https://eips.ethereum.org/EIPS/eip-196), the `EC_MUL` precompile should handle scalars superior to the group order (`n`), as well as the field characteristic (`p`)

> ## Test Cases:
> Inputs to test:
> Curve points which would be valid if the numbers were taken mod p (should fail).
> Both contracts should succeed on empty input.
> Truncated input that results in a valid curve point.
> Points not on curve (but valid otherwise).
> **Multiply point with scalar that lies between the order of the group and the field (should succeed).
> Multiply point with scalar that is larger than the field order (should succeed).**

Currently, the ECIP circuits only handle scalars below `n` of the curve being used.

Therefore, when using the precompile `ec_mul` to compute $Q = kP$, if $k > n$, we need to compute $r \equiv k \mod n = \text{divmod(k, n)}$, and compute $Q = rP$ (with ecip on 1 point).

Considering an elliptic curve over a prime finite field $\mathbb{F}_p$, the characteristic of the field is $p$, and the order of the curve is $n$, while $G$ is the generator point. 

By definition there are $n$ points on the curve and we have $nG = O$ (point at infinity).
Any point of the curve $P$ can be uniquely represented by a scalar multiplication $mG$, with $`m \in (0, n)`$
So, with $`k = q*n + r`$ and $P = mG$, we have $`Q = kP = (q*n + r)P = qnmG + rP = qm \cdot nG + rP = qmO + rP = rP = [k\mod n]P`$

zmalatrax:
Closes #853 
Closes #854
Closes #178


python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
```diff
@@ -517,18 +517,24 @@ func ecip_1p(
...
     assert a0_y * a0_y = a0_x * a0_x * a0_x + a * a0_x + b;
-    // Assert q_low is on curve
-    assert q_low_y * q_low_y = q_low_x * q_low_x * q_low_x + a * q_low_x + b;
-    // Assert q_high is on curve
-    assert q_high_y * q_high_y = q_high_x * q_high_x * q_high_x + a * q_high_x + b;
-    // Assert q_high_shifted is on curve
-    assert q_high_shifted_y * q_high_shifted_y = q_high_shifted_x * q_high_shifted_x *
-        q_high_shifted_x + a * q_high_shifted_x + b;
+    // Assert q_low is on curve or point at infinity
+    assert q_low_y * q_low_y = is_on_curve_q_low * (q_low_x * q_low_x * q_low_x + a * q_low_x + b) + (1 - is_on_curve_q_low) * q_low_x;
```
ClementWalter, line 532: `is_on_curve_q_low = 0 => q_low_y**2 == q_low_x` but why does this mean that both are 0 ?

zmalatrax: If both are 0 then it's the point at infinity (x=0, y=0)

zmalatrax: `is_on_curve` should be refactored as `is_point_at_infinity`

python/cairo-ec/src/cairo_ec/ec_ops.cairo
ClementWalter, line 1: missing a test for ec_mul

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -157,6 +158,114 @@ pub fn build_msm_hints_and_fill_memory() -> Hint {
     )
 }
 
+/// Builds Multi-Scalar Multiplication (MSM) hints and fills memory for elliptic curve operations.
+///
+/// This function processes point coordinates and scalar values to prepare data for MSM operations
+/// on the alt_bn128 curve. It:
+///
+/// 1. Extracts point coordinates (x,y) and scalar values ()
```
obatirou, line 166: scalar value is empty here
```suggestion
/// 1. Extracts point coordinates (x,y) and scalar values (k)
```

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -157,6 +158,114 @@ pub fn build_msm_hints_and_fill_memory() -> Hint {
...
+         ids_data: &HashMap<String, HintReference>,
+         ap_tracking: &ApTracking,
+         _constants: &HashMap<String, Felt252>|
+         -> Result<(), HintError> {
+            const N_LIMBS: usize = 4;
+            let p_addr = get_relocatable_from_var_name("p", vm, ids_data, ap_tracking)?;
+            let x = Uint384::from_base_addr(p_addr, "p.x", vm)?.pack();
+            let y = Uint384::from_base_addr((p_addr + N_LIMBS).unwrap(), "p.y", vm)?.pack();
+
+            let scalar = Uint256::from_var_name("scalar", vm, ids_data, ap_tracking)?.pack();
```
obatirou, line 203: nit:
should this variable in cairo be called k instead ?
And the input to the function ec_mul be called scalar

zmalatrax: The operation is `Q = kP`, `k` and `P` being the input ; to compute the actual value with ecip, we reduce `k` mod n (the group order): $`\text{scalar} = k \mod n`$

I'm not opinionated on the naming, we can invert the variable names if it's smoother

obatirou: I see. after reflexion might not be needed indeed

python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
```diff
@@ -517,18 +517,33 @@ func ecip_1p(
...
     b: felt,
     base_rlc: felt,
+    is_on_curve_q_low: felt,
+    is_on_curve_q_high: felt,
 ) {
+    // Assert is_on_curve_q_low is a boolean flag
+    assert is_on_curve_q_low * (1 - is_on_curve_q_low) = 0;
+    // Assert is_on_curve_q_high is a boolean flag
+    assert is_on_curve_q_high * (1 - is_on_curve_q_high) = 0;
     // Assert g is on curve
```
obatirou, line 527:
```suggestion
    // Assert p is on curve
```

python/cairo-ec/src/cairo_ec/uint384.cairo
```diff
@@ -120,3 +120,12 @@ func uint384_is_neg_mod_p{
         return 0;
     }
 }
+
+// Returns x % p
+// It does not assert that x is in [0, p), nor if the computed value is effectively equal to x mod p.
+func uint384_div_rem{}(x: UInt384, p: UInt384) -> (UInt384, UInt384) {
```
ClementWalter, line 126:
```suggestion
func uint384_div_rem(x: UInt384, p: UInt384) -> (UInt384, UInt384) {
```

ClementWalter: what is this non sound function ?

python/cairo-ec/tests/circuits/test_circuits.py
```diff
@@ -56,7 +56,10 @@ class TestModOps:
         @given(data=st.data())
         @settings(verbosity=Verbosity.quiet)
         def test_add(self, cairo_program, cairo_run, prime_cls, st_prime, data):
-            inputs = {"x": data.draw(st_prime), "y": data.draw(st_prime)}
+            inputs = {
+                "x": 2**384 - 3,
+                "y": 0,
```
ClementWalter, line 61:
```suggestion
                "y": 0
```

python/cairo-ec/src/cairo_ec/uint384.cairo
zmalatrax: I used it to get the quotient and remainder of the input scalar `k` of `ec_mul`, to reduce it modulo `n` (order of the curve), and then I check that `q` is a multiple of `n`, and `r` is less than `n - 1`

```cairo
func ec_mul{...}(...) {
  // ...
  // k = quo * n + rem
  // And scalar = rem, casted as UIn256
  // We assert that the quotient is a multiple of n,
  // and that the remainder is lower than n.
  let (quo, rem) = uint384_div_rem(k, n);
  let q_mul_n = mul(new quo, &n, &n);
  uint384_eq_mod_p([q_mul_n], zero_u384, n);
  let scalar = uint384_to_uint256(rem);
  assert_uint256_le(scalar, n_min_one);
}
```

zmalatrax: But this check might need to be done within `uint384_div_rem` instead ?

At first, I wanted to check that `assert k = add(mul(new quo, &n, &n), new rem, &n)`
But `mul(new quo, &n, &n)` returns 0 when quo is a multiple of n so it didn't work

python/cairo-ec/tests/circuits/test_circuits.py
zmalatrax: oh, forgot to revert a debug commit, mb

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,78 @@ func alt_bn128_mul{
...
+    // Check that x0 is within the interval [0, modulus)
+    let is_x0_out_of_range = U256_le(ALT_BN128_PRIME, x0_value);
+    if (is_x0_out_of_range.value != 0) {
+        return oog_err;
+    }
+    // Check that y0 is within the interval [0, modulus)
+    let is_y0_out_of_range = U256_le(ALT_BN128_PRIME, y0_value);
+    if (is_y0_out_of_range.value != 0) {
+        return oog_err;
     }
```
obatirou, line 208:
```suggestion
    // Check that x0 is within the interval [0, modulus)
    let is_x0_out_of_range = U256_le(ALT_BN128_PRIME, x0_value);
    if (is_x0_out_of_range.value != 0) {
        tempvar err = new EthereumException(OutOfGasError);
        return err;
    }
    // Check that y0 is within the interval [0, modulus)
    let is_y0_out_of_range = U256_le(ALT_BN128_PRIME, y0_value);
    if (is_y0_out_of_range.value != 0) {
        tempvar err = new EthereumException(OutOfGasError);
        return err;
    }
```

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,78 @@ func alt_bn128_mul{
...
+    let gas_cost = Uint(6000);
+    let err = charge_gas(gas_cost);
+    if (cast(err, felt) != 0) {
+        return err;
+    }
+
+    local a: UInt384 = UInt384(alt_bn128.A0, alt_bn128.A1, alt_bn128.A2, alt_bn128.A3);
+    local b: UInt384 = UInt384(alt_bn128.B0, alt_bn128.B1, alt_bn128.B2, alt_bn128.B3);
+    local g: UInt384 = UInt384(alt_bn128.G0, alt_bn128.G1, alt_bn128.G2, alt_bn128.G3);
+    local modulus: UInt384 = UInt384(alt_bn128.P0, alt_bn128.P1, alt_bn128.P2, alt_bn128.P3);
```
enitrat, line 184:
```suggestion
    tempvar a = UInt384(alt_bn128.A0, alt_bn128.A1, alt_bn128.A2, alt_bn128.A3);
    tempvar b = UInt384(alt_bn128.B0, alt_bn128.B1, alt_bn128.B2, alt_bn128.B3);
    tempvar g = UInt384(alt_bn128.G0, alt_bn128.G1, alt_bn128.G2, alt_bn128.G3);
    tempvar modulus = UInt384(alt_bn128.P0, alt_bn128.P1, alt_bn128.P2, alt_bn128.P3);
```

move this block closer right next to where these variables are used

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,78 @@ func alt_bn128_mul{
...
-    let n_value = U256_from_be_bytes(n_bytes);
+    let k_value = U256_from_be_bytes(k_bytes);
 
-    // Gas
-    let gas_cost = Uint(6000);
-    let err = charge_gas(gas_cost);
-    if (cast(err, felt) != 0) {
-        return err;
+    tempvar ALT_BN128_PRIME = U256(new U256Struct(alt_bn128.P_LOW_128, alt_bn128.P_HIGH_128));
+    tempvar oog_err = new EthereumException(OutOfGasError);
```
enitrat, line 198: define inside the if branches

crates/cairo-addons/src/vm/hint_definitions/circuits.rs
```diff
@@ -12,21 +16,44 @@ use cairo_vm::{
...
+            let quo_limbs = split(&q, 4, 96);
+            let rem_limbs = split(&r, 4, 96);
+            write_result_to_ap(quo_limbs[0], 8, vm)?;
+            write_result_to_ap(quo_limbs[1], 7, vm)?;
+            write_result_to_ap(quo_limbs[2], 6, vm)?;
+            write_result_to_ap(rem_limbs[3], 5, vm)?;
+            write_result_to_ap(rem_limbs[0], 4, vm)?;
+            write_result_to_ap(rem_limbs[1], 3, vm)?;
+            write_result_to_ap(rem_limbs[2], 2, vm)?;
+            write_result_to_ap(rem_limbs[3], 1, vm)
```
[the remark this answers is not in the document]
enitrat: yes, also it's better to not rely on ap offsets, but rely on variable names.

This could be `write_collection_from_var_name("q", quo_limbs, vm, ids_data, ap_tracking)`

if it doesnt work - write_collection_to_addr with the address of q

python/cairo-addons/src/cairo_addons/hints/curve.py
```diff
@@ -151,6 +151,76 @@ def build_msm_hints_and_fill_memory(ids: VmConsts, memory: MemoryDict):
...
+        + rlc_coeff_u384_cast_offset
+        + ecip_circuit_constants_offset
+    )
+    ecip_circuit_q_offset = 32 * N_LIMBS
+    fill_felt_ptr(rlc_components, memory, ids.range_check96_ptr + memory_offset)
+    fill_felt_ptr(
+        q_low_high_high_shifted,
+        memory,
+        ids.range_check96_ptr + memory_offset + ecip_circuit_q_offset,
+    )
```
enitrat, line 221: pref our local functions

```
            let offset =
                (range_check96_ptr.segment_index, range_check96_ptr.offset + memory_offset).into();
            write_collection_to_addr(offset, &rlc_components, vm)?;

            let offset = range_check96_ptr + (memory_offset + ecip_circuit_q_offset);
            write_collection_to_addr(offset.unwrap(), &q_low_high_high_shifted, vm)?;
```

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -112,10 +131,209 @@ func ec_add{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: Mod
...
+    mul_mod_ptr: ModBuiltin*,
+    poseidon_ptr: PoseidonBuiltin*,
+}(p: G1Point, k: UInt384, modulus: UInt384) -> G1Point {
+    alloc_locals;
+    let (__fp__, _) = get_fp_and_pc();
+
+    local a: UInt384 = UInt384(alt_bn128.A0, alt_bn128.A1, alt_bn128.A2, alt_bn128.A3);
+    local b: UInt384 = UInt384(alt_bn128.B0, alt_bn128.B1, alt_bn128.B2, alt_bn128.B3);
+    local g: UInt384 = UInt384(alt_bn128.G0, alt_bn128.G1, alt_bn128.G2, alt_bn128.G3);
+    local n: UInt384 = UInt384(alt_bn128.N0, alt_bn128.N1, alt_bn128.N2, alt_bn128.N3);
```
enitrat, line 148: use tempvar + define closer to usage

python/cairo-ec/src/cairo_ec/uint384.cairo
enitrat: yes, the check must be done inside this function. The functions should always (besides very rare exceptions) be sound

enitrat (changes requested):

python/cairo-addons/src/cairo_addons/hints/curve.py
zmalatrax: Is there a local function similar to `write_collection_to_addr` in python?

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,78 @@ func alt_bn128_mul{
...
-    tempvar n_value = n_value;
-
-    tempvar data = data;
-    tempvar error: EthereumException*;
-    tempvar output: Bytes;
+    // Check that p0 is on curve
+    let x0_uint384 = uint256_to_uint384([x0_value.value]);
+    let y0_uint384 = uint256_to_uint384([y0_value.value]);
+    let p0 = G1Point(x=x0_uint384, y=y0_uint384);
+    let k_uint384 = uint256_to_uint384([k_value.value]);
```
obatirou, line 210: Put usage next to usage if possible. This k_uint384 is not related to p0 being on curve or not

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,78 @@ func alt_bn128_mul{
...
+    if (is_on_curve == 0) {
+        assert_not_on_curve(new p0.x, new p0.y, a, b, modulus);
+        tempvar err = new EthereumException(OutOfGasError);
+        return err;
     }
 
+    assert_on_curve(new p0.x, new p0.y, a, b, modulus);
+    tempvar range_check96_ptr = range_check96_ptr;
+    tempvar add_mod_ptr = add_mod_ptr;
+    tempvar mul_mod_ptr = mul_mod_ptr;
```
obatirou, line 241: Is this necessary?

python/cairo-ec/src/cairo_ec/curve/alt_bn128.cairo
```diff
@@ -31,3 +32,19 @@ namespace alt_bn128 {
     const P_MIN_ONE_D2 = 0x30644e72e131a029;
     const P_MIN_ONE_D3 = 0x0;
 }
+
+@known_ap_change
+func sign_to_uint384_mod_alt_bn128(sign: felt) -> UInt384 {
```
obatirou, line 37: No unit test?

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -112,10 +131,203 @@ func ec_add{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: Mod
 }
 
 // Multiply an EC point by a scalar. Doesn't check if the input is on curve nor if it's the point at infinity.
-func ec_mul{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: ModBuiltin*}(
-    p: G1Point, scalar: UInt384, g: UInt384, a: UInt384, modulus: UInt384
-) -> G1Point {
-    // TODO: Implement this function.
-    let res = G1Point(UInt384(0, 0, 0, 0), UInt384(0, 0, 0, 0));
+// Different cases
```
obatirou, line 134:
```suggestion
```

python/cairo-ec/src/cairo_ec/ec_ops.cairo
obatirou: Maked solved but still relevant

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -112,10 +131,203 @@ func ec_add{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: Mod
...
+    local a: UInt384 = UInt384(alt_bn128.A0, alt_bn128.A1, alt_bn128.A2, alt_bn128.A3);
+    local b: UInt384 = UInt384(alt_bn128.B0, alt_bn128.B1, alt_bn128.B2, alt_bn128.B3);
+    local g: UInt384 = UInt384(alt_bn128.G0, alt_bn128.G1, alt_bn128.G2, alt_bn128.G3);
+    local n: UInt384 = UInt384(alt_bn128.N0, alt_bn128.N1, alt_bn128.N2, alt_bn128.N3);
+    let n_min_one = Uint256(alt_bn128.N_LOW_128 - 1, alt_bn128.N_HIGH_128);
+
+    tempvar zero_u384 = UInt384(0, 0, 0, 0);
+    let scalar_is_zero = uint384_eq_mod_p(k, zero_u384, modulus);
+    if (scalar_is_zero != 0) {
+        let point_at_infinity = G1Point(UInt384(0, 0, 0, 0), UInt384(0, 0, 0, 0));
```
obatirou, line 154:
```suggestion
        let point_at_infinity = G1Point(zero_u384, zero_u384);
```
Should work?

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -112,10 +131,202 @@ func ec_add{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: Mod
...
+    hash_full_transcript(range_check96_ptr + ecip_circuit_constants_offset, 22);
+    tempvar range_check96_ptr_init = range_check96_ptr;
+    tempvar range_check96_ptr_after_circuit = range_check96_ptr + 1092;
+    let random_point = get_random_point{range_check96_ptr=range_check96_ptr_after_circuit}(
+        seed=[cast(poseidon_ptr, felt*) - 3], a=&a, b=&b, g=&g, p=&modulus
+    );
+    let range_check96_ptr = range_check96_ptr_init;
+
+    // Circuits inputs
+
```
obatirou, line 256:
```suggestion
```

python/cairo-ec/src/cairo_ec/ec_ops.cairo
```diff
@@ -112,10 +131,202 @@ func ec_add{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: Mod
...
+    let ecip_input: UInt384* = cast(range_check96_ptr + ecip_circuit_constants_offset, UInt384*);
+
+    // Random Linear Combination Sum of Discrete Logarithm Division
+    // rlc_sum_dlog_div for 2 points: n_coeffs = 14 + 4 * 2 = 22 (0-21)
+
```
obatirou, line 261:
```suggestion
```

python/cairo-ec/src/cairo_ec/uint384.cairo
```diff
@@ -120,3 +122,26 @@ func uint384_is_neg_mod_p{
...
+    %{ div_rem_hint %}
+
+    let q_mul_p = mul(new q, new p, new p);
+    tempvar zero_u384 = UInt384(0, 0, 0, 0);
+    let is_equal = uint384_eq([q_mul_p], zero_u384);
+    assert is_equal = 1;
+
+    let one_u384 = UInt384(1, 0, 0, 0);
+    let p_min_one = sub(new p, new one_u384, new p);
+    uint384_assert_le(r, p);
```
obatirou, line 144: `let (__fp__, _) = get_fp_and_pc();` + `&` do not work instead of usage of new?

python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
obatirou, line 1: Should the paper on which those formula are based be added to the repo?
Without it, reviewing (or auditing) this code is tricky

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -175,46 +288,87 @@ pub fn compute_y_from_x_hint() -> Hint {
...
+            // throws at it's not the same type between the arms of a conditional variable
+            // declaration;
+            // TODO: See if there is a better way to reduce code duplication to handle multiple
+            // fields.
+            // TODO: The BN254_PRIME_FIELD_ORDER is not made public by garaga, when it does for the
+            // other curves
+            let bn254_prime_field_order: String = String::from_str(
+                "30644E72E131A029B85045B68181585D97816A916871CA8D3C208C16D87CFD47",
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
zmalatrax: `BigUint::get_modulus_from_curve_id(curve_id: usize)` is not a const fn, which indeed should be used

Only way to have it as a const would be having garaga expose `BN254_PRIME_FIELD_ORDER` as it is done for the other curves (e.g. `SECP256K1_PRIME_FIELD_ORDER`), which is public in lambdaworks_math ; though adding lambdaworks_math as a dep just for that const is not an option imo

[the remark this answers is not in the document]
zmalatrax: For consistency, I chose to remove `SECP256K1_PRIME_FIELD_ORDER`, to use `get_modulus_from_curve_id` for both curves:

```
let secp256k1_prime_field_order = get_modulus_from_curve_id(CurveID::SECP256K1);
let bn254_prime_field_order = get_modulus_from_curve_id(CurveID::BN254);
```

wdyt?

python/cairo-ec/tests/circuits/test_circuits.py
```diff
@@ -790,10 +791,10 @@ def test_ecip_2p(self, cairo_program, cairo_run, data, prime):
         def test_ecip_1p(self, cairo_program, cairo_run, data, prime):
             curve_id = CurveID.from_str("secp256k1")
             curve = CURVES[curve_id.value]
-            g = G1Point.gen_random_point(curve_id)
-            points = [g]
+            p = G1Point.gen_random_point(curve_id)
+            points = [p]
 
-            u1 = data.draw(st.integers(min_value=2**128 + 1, max_value=curve.n))
+            u1 = data.draw(st.integers(min_value=0, max_value=curve.n))
```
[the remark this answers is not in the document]
zmalatrax: `min_value` is `None` by default, from the docstring
> Examples from this strategy will shrink towards zero, and negative values
    will also shrink towards positive (i.e. -n may be replaced by +n).

Thus, is `min_value` is not set to 0, hypothesis will take negative values (as integers are `int`)

python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
zmalatrax: The `ecip` circuits have a link to the actual paper, and the equation that are being used (w/ page number) which implements the formula:

> // It verifies that the equation (3) holds ; from the paper
// "Zero Knowledge Proofs of Elliptic Curve Inner Products
// from Principal Divisors and Weil Reciprocity", by Liam Eagen
// (source: https://eprint.iacr.org/2022/596.pdf, p.9).

Do you want me to also add a comment on the use of ECIP on the functions using ECIP? (i.e. `ec_mul` and `alt_bn128_mul`

python/cairo-ec/src/cairo_ec/curve/alt_bn128.cairo
zmalatrax: fait pour `sign_to_uint384_mod_alt_bn128`

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -175,46 +288,87 @@ pub fn compute_y_from_x_hint() -> Hint {
...
+                    ap_tracking,
+                )?;
+                write_collection_from_var_name(
+                    "is_on_curve",
+                    &[is_on_curve.into(), Felt252::ZERO, Felt252::ZERO, Felt252::ZERO],
+                    vm,
+                    ids_data,
+                    ap_tracking,
+                )?;
+            } else if p.to_str_radix(16).to_uppercase() == bn254_prime_field_order {
```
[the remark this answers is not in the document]
zmalatrax: Made 3 if statements instead, w/ early panic if not SECP256K1 or BN254

python/cairo-addons/src/cairo_addons/hints/curve.py
obatirou: You can use `segments.load_data(ptr, data)`

python/cairo-ec/tests/circuits/test_circuits.py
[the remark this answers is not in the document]
obatirou: oh you are absolutely right 
don't why I though this my bad for suggesting this

python/cairo-ec/src/cairo_ec/circuits/ec_ops.cairo
obatirou: Thought is was not already done. I think this enough

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -175,46 +288,89 @@ pub fn compute_y_from_x_hint() -> Hint {
...
+                )?;
+                write_collection_from_var_name(
+                    "is_on_curve",
+                    &[is_on_curve.into(), Felt252::ZERO, Felt252::ZERO, Felt252::ZERO],
+                    vm,
+                    ids_data,
+                    ap_tracking,
+                )?;
+            }
             Ok(())
```
obatirou, line 374: You could do an early return in the if is_secp256k1_p and remove the second if (add a comment only as

python/cairo-ec/src/cairo_ec/uint384.cairo
```diff
@@ -120,3 +122,27 @@ func uint384_is_neg_mod_p{
...
+    %{ div_rem_hint %}
+
+    let q_mul_p = mul(&q, &p, &p);
+    tempvar zero_u384 = UInt384(0, 0, 0, 0);
+    let is_equal = uint384_eq([q_mul_p], zero_u384);
+    assert is_equal = 1;
+
+    let one_u384 = UInt384(1, 0, 0, 0);
+    let p_min_one = sub(&p, new one_u384, &p);
+    uint384_assert_le(r, p);
```
obatirou, line 145:
```suggestion
    uint384_assert_le(r, p_min_one);
```

python/cairo-ec/src/cairo_ec/uint384.cairo
```diff
@@ -120,3 +122,27 @@ func uint384_is_neg_mod_p{
         return 0;
     }
 }
+
+// Returns (q, r) s.t x = q*p + r
+func uint384_div_rem{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: ModBuiltin*}(
```
obatirou, line 127: It seems to me that an assertion is missing: 
* `(q * p) mod p == 0` is checked
*  `r < p` is checked 

but not that `x = q*p + r`
So the prover could fill the memory that pass the 2 first assertions but that do not verify `x = q*p + r`

zmalatrax: Indeed, I first tried with `uint384_eq_mod_p` but it actually does not reduce mod p the inputs when verifying the `assert_eq` circuit ; as it didn't work I didn't continue

But it can actually be done by using `add(x, 0, p)` or mul(x, 1, p) ; will add it

cairo/ethereum/cancun/vm/precompiled_contracts/alt_bn128.cairo
```diff
@@ -171,40 +171,75 @@ func alt_bn128_mul{
...
-    if (cast(error, felt) != 0) {
-        return error;
+    tempvar a = new UInt384(alt_bn128.A0, alt_bn128.A1, alt_bn128.A2, alt_bn128.A3);
+    tempvar b = new UInt384(alt_bn128.B0, alt_bn128.B1, alt_bn128.B2, alt_bn128.B3);
+    tempvar modulus = new UInt384(alt_bn128.P0, alt_bn128.P1, alt_bn128.P2, alt_bn128.P3);
+    tempvar g = new UInt384(alt_bn128.G0, alt_bn128.G1, alt_bn128.G2, alt_bn128.G3);
+    tempvar is_on_curve;
+    tempvar point = p0;
+    %{ is_point_on_curve %}
+    tempvar is_p0_on_curve_uint384 = UInt384(is_on_curve, 0, 0, 0);
```
enitrat, line 230: looks unused

crates/cairo-addons/src/vm/hint_definitions/circuits.rs
```diff
@@ -12,21 +16,38 @@ use cairo_vm::{
...
     has_one_uint384_remaining_hint,
-    felt_to_uint384_split_hint,
     x_mod_p_eq_y_mod_p_hint,
     x_is_neg_y_mod_p_hint,
 ];
 
+pub fn div_rem_hint() -> Hint {
+    Hint::new(
+        String::from("div_rem_hint"),
+        |vm: &mut VirtualMachine,
```
enitrat, line 34: this should've been called u384_div_rem_hint because it's type specific

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -175,46 +288,89 @@ pub fn compute_y_from_x_hint() -> Hint {
...
+                    ap_tracking,
+                )?;
+                write_collection_from_var_name(
+                    "is_on_curve",
+                    &[is_on_curve.into(), Felt252::ZERO, Felt252::ZERO, Felt252::ZERO],
+                    vm,
+                    ids_data,
+                    ap_tracking,
+                )?;
+            }
```
enitrat, line 339:
```suggestion
            if is_secp256k1_p {
                let rhs_felt =
                    FieldElement::<SECP256K1PrimeField>::from_hex_unchecked(&rhs.to_str_radix(16));
                let g_felt =
                    FieldElement::<SECP256K1PrimeField>::from_hex_unchecked(&g.to_str_radix(16));

                let is_on_curve = is_quad_residue(&rhs, &p);
                let square_root = if is_on_curve {
                    let sqrt_felt = rhs_felt.sqrt().unwrap().0;
                    let has_same_parity =
                        (v.clone() % 2_u32) == (element_to_biguint(&sqrt_felt) % 2_u32);
                    if has_same_parity {
                        sqrt_felt
                    } else {
                        -sqrt_felt
                    }
                } else {
                    (rhs_felt * g_felt).sqrt().unwrap().0
                };

                let sqrt_biguint = element_to_biguint(&square_root);
                Uint384::split(&sqrt_biguint).insert_from_var_name(
                    "y_try",
                    vm,
                    ids_data,
                    ap_tracking,
                )?;
                write_collection_from_var_name(
                    "is_on_curve",
                    &[is_on_curve.into(), Felt252::ZERO, Felt252::ZERO, Felt252::ZERO],
                    vm,
                    ids_data,
                    ap_tracking,
                )?;
                return Ok(())
            }
```

crates/cairo-addons/src/vm/hint_definitions/curve.rs
```diff
@@ -175,46 +288,89 @@ pub fn compute_y_from_x_hint() -> Hint {
...
```
enitrat, line 373:
```suggestion
            let rhs_felt =
                FieldElement::<BN254PrimeField>::from_hex_unchecked(&rhs.to_str_radix(16));
            let g_felt =
                FieldElement::<BN254PrimeField>::from_hex_unchecked(&g.to_str_radix(16));

            let is_on_curve = is_quad_residue(&rhs, &p);
            let square_root = if is_on_curve {
                let sqrt_felt = rhs_felt.sqrt().unwrap().0;
                let has_same_parity = (v % 2_u32) == (element_to_biguint(&sqrt_felt) % 2_u32);
                if has_same_parity {
                    sqrt_felt
                } else {
                    -sqrt_felt
                }
            } else {
                (rhs_felt * g_felt).sqrt().unwrap().0
            };

            let sqrt_biguint = element_to_biguint(&square_root);
            Uint384::split(&sqrt_biguint).insert_from_var_name(
                "y_try",
                vm,
                ids_data,
                ap_tracking,
            )?;
            write_collection_from_var_name(
                "is_on_curve",
                &[is_on_curve.into(), Felt252::ZERO, Felt252::ZERO, Felt252::ZERO],
                vm,
                ids_data,
                ap_tracking,
            )?;
```

enitrat (approved):

enitrat:
ok for me besides suggestions

crates/cairo-addons/src/vm/hint_definitions/circuits.rs
zmalatrax: indeed, thx!
this hint will be removed (as it was actually unnecessary)

enitrat (approved):

PR #887, opened 2025-02-27
dev: expose `serialize` in RustVM pythonic hints

enitrat:
Exposes a `serialize` function, available in the scope of all hints, that enables the serialization using `serde` of cairo types, and returns their python representation.

This can be used for the logging / tracing system.

I am welcoming any alternative idea on the currently proposed design - as working with python `globals()` has been quite a hassle.

Also, i'm not sure yet how we can make this entirely transparent with the pythonVM runs. If we realize there's a need for that, we can do it later.

### DIFF
--- a/crates/cairo-addons/src/vm/dynamic_hint.rs
+++ b/crates/cairo-addons/src/vm/dynamic_hint.rs
@@ -37,6 +37,15 @@
 //! %}
 //! ```
 //!
+//! Serialize a variable from `ids` using the `serialize` function.
+//! This uses the factory function `serialize` that is injected into the execution scope in
+//! runner.rs.
+//!
+//! ```cairo
+//! tempvar evm = Evm(...)
+//! %{ serialize(ids.evm) %}
+//! ```
+//!
 //! ## Limitations
 //!
 //! - Direct mutation of `ids` variables is not supported.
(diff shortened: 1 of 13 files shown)


crates/cairo-addons/src/vm/dynamic_hint.rs
```diff
@@ -188,21 +225,27 @@ impl DynamicPythonHintExecutor {
...
-            locals
+            // Add the ids dictionary to context
+            bounded_context
                 .set_item("ids", py_ids_dict)
                 .map_err(|e| DynamicHintError::PyDictSet(e.to_string()))?;
 
+            let injected_py_code = r#"
+from functools import partial
+
+serialize = partial(globals()["serialize"], segments=globals()["segments"], program_identifiers=globals()["py_identifiers"], dict_manager=globals()["dict_manager"])
```
enitrat, line 241: this makes `serialize` available in all pythonic hints

crates/cairo-addons/src/vm/dynamic_hint.rs
```diff
@@ -188,21 +221,25 @@ impl DynamicPythonHintExecutor {
...
                 .set_item("ids", py_ids_dict)
                 .map_err(|e| DynamicHintError::PyDictSet(e.to_string()))?;
 
-            // Execute the Python code
-            let hint_code_c_string = CString::new(hint_code)
+            // Inject the serialize function factory into the execution scope
+            let injected_py_code = r#"
+from functools import partial
+
+serialize = partial(globals()["serialize"], segments=globals()["segments"], program_identifiers=globals()["py_identifiers"], dict_manager=globals()["dict_manager"])
```
ClementWalter, line 235:
```suggestion
serialize = partial(serialize, segments=segments, program_identifiers=py_identifiers, dict_manager=dict_manager)
```

ClementWalter (approved):

PR #903, opened 2025-02-28
fix: precompiles warmup

### LINKED ISSUE #902: bug: track_precompiles doesnt take into accounts already warmed up precompiles
if a tx has in accessed addresses precompiles, then our `track_precompiles` hint followed by the memcpy of the constant values defined in `dw` of the label `precompiled_addresses_set` will collide with the already accessed addresses.

Thus, we should fallback to a manual loop.

enitrat:
Closes #902

### DIFF
--- a/cairo/ethereum/cancun/utils/message.cairo
+++ b/cairo/ethereum/cancun/utils/message.cairo
@@ -21,42 +21,9 @@ from starkware.cairo.common.registers import get_fp_and_pc, get_label_location
 from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.dict_access import DictAccess
 
-from legacy.utils.dict import hashdict_write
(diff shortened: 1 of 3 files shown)


ClementWalter (approved):

cairo/ethereum/cancun/utils/message.cairo
```diff
@@ -140,16 +107,12 @@ func prepare_message{
     hashdict_write{dict_ptr=preaccessed_addresses_ptr}(1, current_target_, 1);
     hashdict_write{dict_ptr=preaccessed_addresses_ptr}(1, &caller.value, 1);
 
-    let dict_ptr = preaccessed_addresses_ptr;
-    %{ track_precompiles %}
```
obatirou, line 144: Not used anywhere anymore.
Hints need to be deleted.

obatirou (approved):

PR #905, opened 2025-02-28
dev: logger

### LINKED ISSUE #888: dev: implement tracing / logging through hints
## Why

Following https://github.com/kkrt-labs/keth/pull/887 we'll be able to run any python code AND to serialize all cairo objects into their pythonic representation in hints.

We'd like to be able to run traces in hints. 

## What

Logging (first step):
- Implement a way to easily log values at key steps of the execution, to debug ef-tests more easily

## How

Make it so that we can capture the logs emitted to stdout to get a better view into the execution cycle of keth. we will be able to compare it to execution-specs implementing a similar system, and compare the two side-by-side.

### LINKED ISSUE #920: dev: add CLI flag to disable pythonic hint execution
To avoid emitting traces and executing pythonic hints (which can slow down execution), add a CLI flag that we can pass to the rust runner. If that flag is activated, skip execution of python hints.

enitrat:
Adds a logger in `log-cli-level="TRACE"` mode. 

Closes #888
Closes #920


cairo/tests/conftest.py
```diff
@@ -60,16 +134,29 @@ def pytest_configure(config):
...
+    # Mock the Extended type
+    ethereum_rlp.rlp.Extended = Union[Sequence["Extended"], bytearray, bytes, Uint, FixedUnsigned, str, bool]  # type: ignore # noqa: F821
+
+    # Patching evm_trace:
+    # - Problem: Global patches of `ethereum.trace.evm_trace` are not reflected in places where `evm_trace` is imported in EELS.
+    # - Cause: `ethereum.cancun.vm.interpreter` (and other modules) imports `evm_trace` locally (e.g., `from ethereum.trace import evm_trace`)
+    #   at module load time, caching the original `discard_evm_trace`. Patching `ethereum.trace.evm_trace` later didn’t
+    #   update this local reference due to Python’s import caching.
+    # - Solution: Explicitly patch both `ethereum.trace.evm_trace` globally and
+    #   `ethereum.cancun.vm.interpreter.evm_trace` locally (and other places where `evm_trace` is imported).
```
ClementWalter, line 154: still thinking that there is most probably a way to patch globally, with unittest.mock or other options. I don't mind leaving this as is, but leave a TODO to investigate other options

crates/cairo-addons/Cargo.toml
```diff
@@ -40,4 +40,4 @@ pyo3-build-config = "0.23.3" # Should match pyo3 version
 extension-module = ["pyo3/extension-module"]
 default = ["extension-module", "pyo3/experimental-inspect"]
 # Dynamic hints feature - not enabled by default
```
ClementWalter, line 42: update comment

crates/cairo-addons/src/vm/runner.rs
```diff
@@ -208,9 +220,21 @@ except Exception as e:
         Ok(PyDictManager { inner: dict_manager })
     }
 
-    fn run_until_pc(&mut self, address: PyRelocatable, resources: PyRunResources) -> PyResult<()> {
-        let mut hint_processor =
-            HintProcessor::default().with_run_resources(resources.inner).build();
+    #[pyo3(signature = (address, resources, enable_pythonic_hints=false))]
+    fn run_until_pc(
```
ClementWalter, line 224: if the option is given at `new` I don't think that we should have a second time the param. What happens if you given False in new and True here?

crates/cairo-addons/src/vm/vm_consts.rs
```diff
@@ -147,6 +147,15 @@ fn create_var_type(
                     CairoVarType::Struct { name: base_type.to_string(), members, size }
                 }
                 None => {
+                    if base_type == "(fp_val: felt, pc_val: felt*)" {
+                        // Manual handling of fp_val and pc_val, which causes issues in both VMs.
```
ClementWalter, line 151: what kind of issue ?

log_analyzer.py
ClementWalter, line 1: move in scripts

python/cairo-addons/src/cairo_addons/testing/hooks.py
```diff
@@ -91,6 +97,12 @@ def pytest_addoption(parser):
         type=int,
         help="The seed to set random with.",
     )
+    parser.addoption(
+        "--disable-pythonic-hints",
+        action="store_true",
+        default=False,
+        help="Disable execution of pythonic hints: True or False",
+    )
```
ClementWalter, line 105: where is this used ?

crates/cairo-addons/src/vm/vm_consts.rs
enitrat: in RustVM it just tells you that (fp_al: felt, pc_val: felt*) cannot be bound to a type (and thus it crashes at this step, when loading the vm consts)
in the PythonVM it raises an exception upon trying to access fp_and_pc (same kind of issues)

cairo/tests/conftest.py
enitrat: probably but I could not get anything other to work

crates/cairo-addons/src/vm/runner.rs
enitrat: added as a field in the runner, set upon new

python/cairo-addons/src/cairo_addons/testing/hooks.py
enitrat: deleted

ClementWalter (approved):
create a follow up to instanciate several loggers with different formatting instead of the `[FLAGS]` that you put

PR #925, opened 2025-03-03
fix: account_has_storage in cases of destroyed storage

### LINKED ISSUE #924: bug: account_has_storage_keys returns wrong value on destroyed accounts
If the account is destroyed, the entry is still present in the dict segment; but the associated value is zero.

As such, the hint `get_keys_for_address_prefix` will return some keys.

enitrat:
This is a rough patch of this issue by adding another hint that filtrates for zero values. This is not the definitive design, which will be refactored to handle proofs of non-inclusion.

This is the same as get_keys_for_address_prefix, only we do not collect zero-values.

Closes #924

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -510,18 +510,38 @@ func account_has_code_or_nonce{poseidon_ptr: PoseidonBuiltin*, state: State}(
     return res;
 }
 
+// @notice Returns all (address, key) pairs for a given address.
+// This is a temporary fix specifically made for the `account_has_storage` function.
+// We need to make a deeper refactor once we have a sound way to handle storage keys.
+func get_storage_keys_for_address{dict_ptr: DictAccess*}(
+    prefix_len: felt, prefix: felt*
+) -> ListTupleAddressBytes32 {
+    alloc_locals;
+
+    local keys_len: felt;
+    local keys: TupleAddressBytes32*;
+    %{ get_storage_keys_for_address %}
+
+    // warning: this is unsound as the prover can return any list of keys.
+    tempvar res = ListTupleAddressBytes32(new ListTupleAddressBytes32Struct(keys, keys_len));
+    return res;
+}
+
 func account_has_storage{poseidon_ptr: PoseidonBuiltin*, state: State}(address: Address) -> bool {
     alloc_locals;
 
     let fp_and_pc = get_fp_and_pc();
     local __fp__: felt* = fp_and_pc.fp_val;
 
-    let storage_tries = state.value.original_storage_tries;
+    let storage_tries = state.value._storage_tries;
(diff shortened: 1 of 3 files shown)


cairo/ethereum/cancun/state.cairo
```diff
@@ -510,18 +510,38 @@ func account_has_code_or_nonce{poseidon_ptr: PoseidonBuiltin*, state: State}(
     return res;
 }
 
+// @notice Returns all (address, key) pairs for a given address.
+// This is a temporary fix specifically made for the `account_has_storage` function.
+// We need to make a deeper refactor once we have a sound way to handle storage keys.
+func get_storage_keys_for_address{dict_ptr: DictAccess*}(
```
obatirou, line 516: nit: put both this one and get_keys_for_address_prefix next to each other in this file or in legacy.utils.dict.

crates/cairo-addons/src/vm/hint_definitions/hashdict.rs
```diff
@@ -207,6 +209,108 @@ pub fn get_keys_for_address_prefix() -> Hint {
...
+                            }
+                        } else {
+                            None
+                        }
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+
```
obatirou, line 273:
```suggestion
            let matching_preimages: Vec<&Vec<MaybeRelocatable>> = tracker
                .get_dictionary_ref()
                .iter()
                .filter_map(|(key, value)| {
                    match key {
                        DictKey::Compound(key_parts) => {
                            // Chain Option methods instead of nested if lets
                            value
                                .get_int()
                                .filter(|&v| v != Zero::zero())
                                .and_then(|_| {
                                    // Use if expression instead of if/else with None
                                    if key_parts.len() >= prefix.len() && key_parts[..prefix.len()] == prefix[..] {
                                        Some(key_parts)
                                    } else {
                                        None
                                    }
                                })
                        }
                        _ => None,
                    }
                })
                .collect();

```

python/cairo-addons/src/cairo_addons/hints/hashdict.py
```diff
@@ -82,6 +82,30 @@ def get_keys_for_address_prefix(
...
 
 @register_hint
+def get_storage_keys_for_address(
+    dict_manager: DictManager,
+    ids: VmConsts,
+    segments: MemorySegmentManager,
+    memory: MemoryDict,
+):
+    dict_tracker = dict_manager.get_tracker(ids.dict_ptr)
+    prefix = tuple([memory[ids.prefix + i] for i in range(ids.prefix_len)])
```
obatirou, line 92:
```suggestion
    prefix = tuple(memory[ids.prefix + i] for i in range(ids.prefix_len))
```
Should work

ClementWalter (approved):

PR #942, opened 2025-03-04
fix: blob-related ef-tests failures

### LINKED ISSUE #929: bug: tx validation fails on blob gas costs
`"cairo/tests/ef_tests/cancun/test_state_transition.py::test_general_state_tests[eip4844_blobs - src/GeneralStateTestsFiller/Pyspecs/cancun/eip4844_blobs/test_excess_blob_gas.py::test_correct_increasing_blob_gas_costs[fork_Cancun-parent_blobs_4-parent_excess_blobs_1129-blockchain_test-new_blobs_1]]"` fails on the second tx of a block. Considering that the second tx is not even strting, I think it gets killed during validation (and it should not)

### LINKED ISSUE #937: tracking: failing ef-tests
```
FAILED cairo/tests/ef_tests/cancun/test_state_transition.py::test_general_state_tests[bcWalletTest - walletReorganizeOwners_Cancun] - Failed: Timeout >300.0s
```

List of failing ef-tests. Investigate & fix them.

enitrat:
Closes #937
Closes #929

### DIFF
--- a/cairo/ethereum/cancun/vm/gas.cairo
+++ b/cairo/ethereum/cancun/vm/gas.cairo
@@ -281,16 +281,16 @@ func calculate_total_blob_gas{range_check_ptr}(tx: Transaction) -> Uint {
     return total_blob_gas;
 }
 
-// @dev: Saturates at 2**64 - 1
+// @dev: Saturates at 2**128 - 1
 func calculate_blob_gas_price{range_check_ptr}(excess_blob_gas: U64) -> Uint {
     let blob_gas_price = taylor_exponential(
         Uint(GasConstants.MIN_BLOB_GASPRICE),
         Uint(excess_blob_gas.value),
         Uint(GasConstants.BLOB_GASPRICE_UPDATE_FRACTION),
     );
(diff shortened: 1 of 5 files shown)


cairo/ethereum/cancun/vm/gas.cairo
```diff
@@ -300,9 +300,14 @@ func calculate_data_fee{range_check_ptr}(excess_blob_gas: U64, tx: Transaction)
     alloc_locals;
     // saturate at 2**64 - 1
     let total_blob_gas = calculate_total_blob_gas(tx);
-    // saturate at 2**64 - 1
+    // saturate at 2**128 - 1
```
obatirou, line 303: Should calculate_total_blob_gas be saturate also at 2**128 - 1 ?

enitrat: Not necessarily. This is kind of overfitting for tests, 2**64 would've been fine as well as the TX would've been killed in a later part of the tx

obatirou: ok fine for me

PR #962, opened 2025-03-10
dev: MPT library for EELS

Eikix:
> UPDATE: How to move forward on this PR?
> 
> Problems & Challenges:
> 
> * too many new LoCs
> * code quality not up to codebase standard
> * logic that does not serve our codebase directly
> 
> Recommended solution:
> 
> * Break down the PR in smaller PRs that directly serve a purpose for keth.
> 
> Next actionable:
> 
> * Open a PR with an eth rpc minimal module with methods for eth_getProof, eth_getCode
> * Open a PR with a way to convert a zkpi json file to a EthereumTries object and a method to flatten the tries to obtain a EELS state object.

- load an `EthereumState` object from ZKPI which is a flattened view of all MPT nodes (State Trie and Storage Tries) and Codes touched during block, as well as an `accessList` field of all addresses and storage keys touched during block, and a state_root.
- entrypoint to transform `EthereumState` into an EELS `State` object
- entrypoints to get/set/delete nodes in the EthereumState


python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+)
+
+EMPTY_CODE_HASH = Hash32(
+    bytes.fromhex("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470")
+)
+
+EMPTY_BYTES_RLP = b"\x80"
+
+
+def nibble_path_to_hex(nibble_path: Bytes) -> Bytes:
```
enitrat, line 51:
```suggestion
def nibble_path_to_hex(nibble_path: Bytes) -> str:
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+    Represents the (partial) state of Ethereum for a given block.
+    Includes all account MPT leaf nodes and storage MPT leaf nodes and all intermediary nodes touched in a block,
+    related codes, an access list of addresses and storage keys that were accessed in the block, and a state root.
+    """
+
+    nodes: Mapping[Bytes32, Bytes]
+    codes: Mapping[Bytes32, Bytes]
+    access_list: Mapping[Address, Optional[List[Bytes32]]]
+    state_root: Hash32
+
```
enitrat, line 76:
```suggestion
    nodes: Mapping[Bytes32, Bytes]        # Stores MPT nodes by their hash
    codes: Mapping[Bytes32, Bytes]        # Stores contract code by code hash
    access_list: Mapping[Address, Optional[List[Bytes32]]]  # Tracks accessed addresses and storage
    state_root: Hash32                    # Root hash of the state trie
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+                [
+                    nibble_path[i] * 16 + nibble_path[i + 1]
+                    for i in range(0, len(nibble_path) - 1, 2)
+                ]
+            )
+        ).hex()
+    )
+
+
+class EthereumState:
```
enitrat, line 65: PartialState ? Confusing with EELS' "State" imo

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+
+    def to_state(self) -> State:
+        """
+        Convert the EthereumState to a State object.
+
+        Returns
+        -------
+        State
+            A State object representing the Ethereum state
+        """
```
enitrat, line 110: see here -> State object representing Ethereum State, on a class named EthereumState, that returns State 🥴

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+
+        # Step 1: Recursively explore the state trie to get to the leaves
+        _main_trie: Trie[Address, Optional[Account]] = Trie(secured=True, default=None)
+        _storage_tries: Dict[Address, Trie[Bytes32, U256]] = {}
```
enitrat, line 114:
```suggestion
        _main_trie: Trie[Address, Optional[Account]] = Trie(secured=True, default=None, _data=defaultdict(lambda: None))
```

enitrat: just thinking that this _might_ be an issue at a later point when converting the State object to a Cairo state during `serde`

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+
+        for address in self.access_list.keys():
+
+            account = self.get(keccak256(address))
```
enitrat, line 118: is this computed once only? or is there a risk of hashing the same address multiple times? (if so we can add memoization)

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+                    ),
+                )
+                continue
+
+            decoded = rlp.decode(account)
+            if len(decoded) != 4:
+                raise ValueError(f"Invalid account length: {len(decoded)}")
+
+            if decoded[3] == EMPTY_CODE_HASH:
+                code = b""
```
enitrat, line 137:
```py
from dataclasses import dataclass
from ethereum_types.bytes import Bytes32

@dataclass
class RLPAccount:
    nonce: bytes
    balance: bytes
    storage_root: bytes
    code_hash: Bytes32

decoded = RLPAccount(*rlp.decode(account))
```

would be nice to have a type here so as not to guess what each accessed index here (all these nits can be fed to a LLM and one-shotted later). We can zip all PR comments and give it to sonnet

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+            else:
+                code = self.codes.get(decoded[3], None)
+                # TODO: This is a hack to get the code for codes not present in the EthereumState object
+                # This is due to accessing code only through EXTCODEHASH opcode
```
enitrat, line 141: This means, that EXTCODEHASH opcodes don't require loading the account code, but because our (and EELS') modeling of accounts does not store codehash, we would need the entire code here?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+                ),
+            )
+
+            # Process storage for this account if it has storage keys in the access list
+            storage_root = decoded[2]
+            if not storage_root:
+                raise ValueError(f"Storage root is None for address: {address}")
+
+            if self.access_list.get(address) is not None:
+                for key in self.access_list[address]:
```
enitrat, line 186:
```suggestion
            if address in self.access_list:
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+                    value = self.get(keccak256(key), Hash32(storage_root))
+                    if value is None:
+                        logger.debug(
+                            f"Value is None for key: 0x{key.hex()} for address: 0x{address.hex()}"
+                        )
```
enitrat, line 191: how can value be None here, when the fallbask is `storage_root` which can't be `None` (checkd before)?

are we supposed to fallback on the root ?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+                        continue
+                    if _storage_tries.get(address) is None:
+                        _storage_tries[address] = Trie(secured=True, default=U256(0))
+                    trie_set(
```
enitrat, line 195:
```suggestion
                        _storage_tries[address] = Trie(secured=True, default=U256(0), _data=defaultdict(lambda: U256(0)))
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        if is_leaf:
+            return LeafNode(rest_of_key=current_path, value=value)
+        else:
+            return ExtensionNode(key_segment=current_path, subnode=value)
+    else:
+        raise ValueError(f"Unknown node structure: {type(decoded)}")
+
+
+# Redefinition of encode_internal_node from ethereum.cancun.trie
+# without keccak256 of the RLP encoded node if its length is greater than 32 bytes
```
enitrat, line 1333: what's the reason? to reduce # of keccak done?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+            return None
+
+    def get_storage_root(self, address: Address) -> Optional[Hash32]:
+        """
+        Get the storage root for an account.
+        """
+        account = self.get(keccak256(address))
+        if account is None:
+            return EMPTY_TRIE_ROOT_HASH
+        decoded = rlp.decode(account)
```
enitrat, line 283: (echoing to a previous comment with a class for decoded accounts, we could do this validation when creating the class instance), as I see it's used in a lot of places

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        """
+        account = self.get(keccak256(address))
+        if account is None:
+            return None
+
+        decoded = rlp.decode(account)
+        if len(decoded) != 4:
+            raise ValueError(f"Invalid account length: {len(decoded)}")
+
+        # TODO: Fix situation where code is not present in self.codes
```
enitrat, line 300: This is done, as we fetch the code from the node if it's not in the zkpi right?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+
+        node_data = self.nodes.get(node_hash)
+
+        if node_data is None:
+            raise ValueError(f"Node not found: 0x{node_hash.hex()}")
+
+        node = self._decode_node(node_data)
+        return self._process_node(node, nibble_path)
+
+    def _process_node(self, node: InternalNode, nibble_path: Bytes) -> Optional[Bytes]:
```
enitrat, line 335: This is recursive until we reach a leaf node right?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        path : Bytes
+            The path to delete
+        root_hash : Hash32, optional
+            The root hash to start from, defaults to the trie's state_root
+
+        Returns
+        -------
+        Optional[Hash32]
+            The new root hash after deletion, or None if the path wasn't found
+        """
```
enitrat, line 482:
```suggestion
        """
        Delete a value from the trie at the given path.

        The deletion process in a Merkle Patricia Trie follows these steps:

        1. Navigate to the target node using the path:
            - At each branch node, follow the appropriate nibble
            - At each extension node, verify the shared prefix matches
            - Continue until reaching a leaf node

        2. When the target node is found:
            - If it's a leaf node and the path matches exactly, remove it
            - If it's not found, return without changes

        3. After deletion, restructure the trie to maintain invariants:
            - If a branch node is left with only one child, convert it to an extension node
            - If an extension node's child is deleted, remove the extension node
            - Merge extension nodes with their child extension nodes when possible
            - Update all affected node hashes up to the root

        For example, if deleting 'abc' from a trie:
        ```
        Before:                After:
        [Root]                [Root]
          |                     |
        [Branch]              [Branch]
         /    \                /    \
        'ab'   'd'          'ab'   'd'
          |                   |
        'c':123             'x':456
        'x':456
        ```

        Parameters
        ----------
        path : Bytes
            The path to delete
        root_hash : Hash32, optional
            The root hash to start from, defaults to the trie's state_root

        Returns
        -------
        Optional[Hash32]
            The new root hash after deletion, or None if the path wasn't found.
            Returns the original root_hash if no changes were made.
        """
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1372 @@
...
+        value : Bytes
+            The RLP-encoded value to store
+        root_hash : Hash32, optional
+            The root hash to start from, defaults to the trie's state_root
+
+        Returns
+        -------
+        Hash32
+            The new root hash after insertion/update
+        """
```
enitrat, line 800:
```suggestion
    """
    Insert or update a value in the trie at the given path.

    The upsert process in a Merkle Patricia Trie follows these steps:

    1. Special case - Empty trie:
       - If the trie is empty (root_hash == EMPTY_TRIE_ROOT_HASH)
       - Create a new leaf node with the entire path and value
       - No need for complex path splitting or node creation

    2. Path traversal and node creation:
       a) At a Branch node:
          - If path is empty: (not supported in this implementation)
          - Otherwise: Follow the next nibble
          - If that branch is empty: Create new leaf node
          - If branch exists: Recursively upsert into that branch

       b) At an Extension node:
          - Find common prefix between path and node's key_segment
          - If paths diverge: Create a branch node at divergence point
          - If extension is prefix of path: Recursively upsert remaining path
          - Handle path compression by merging extension nodes when possible

       c) At a Leaf node:
          - If paths match exactly: Update value
          - If paths differ: Create branch node at first different nibble
          - Add both paths (existing and new) to the branch
          - Create extension node if common prefix exists

    3. Node encoding and storage:
       - Encode modified nodes using RLP
       - For nodes ≥ 32 bytes: Store separately and use hash as reference
       - For nodes < 32 bytes: Store directly in parent node
       - Update node hashes up to root

    Example of inserting 'abc':123 into a trie:
    ```
    Before:                After:
    [Root]                [Root]
      |                     |
    [Branch]              [Branch]
      |                   /    \
     'd'               'ab'    'd'
      |                 |      |
    456              'c':123  456
    ```

    Parameters
    ----------
    path : Bytes
        The path where the value should be stored
    value : Bytes
        The RLP-encoded value to store
    root_hash : Hash32, optional
        The root hash to start from, defaults to the trie's state_root

    Returns
    -------
    Hash32
        The new root hash after insertion/update
    """
```

enitrat (commented):
lgtm overall, although i would need to spend more time on the deletion / upsertion process

python/mpt/src/mpt/__init__.py
Eikix: normally the access_list is a set, so only once

python/mpt/src/mpt/__init__.py
Eikix: good idea

python/mpt/src/mpt/__init__.py
Eikix: it appears to me that the suggestion does not work, we're replacing an iteration of storage keys for a given address by an if statement

python/mpt/src/mpt/__init__.py
Eikix: is this correct?

python/mpt/src/mpt/__init__.py
Eikix: yes

python/mpt/src/mpt/__init__.py
Eikix: no, it's because in encode_internal_node from ethereum lib, it automatically hashes the result of the RLP encoding if its length is >= 32, but sometimes in our implementation I thought we might not want to hash in some instances. 

after checking all the times i called this function, we always hash. i'll replace this by the real function

Eikix: ERRATUM: i need to separate the two steps -> RLPencode & hash, to store encoded nodes at a specific hash
EELS's `encode_internal_node` does not let me do that

python/mpt/src/mpt/__init__.py
enitrat: yeah the suggestion is off by one line. the goal is to replace 

> if self.access_list.get(address) is not None:

by `if address in self.access_list:`

python/mpt/src/mpt/__init__.py
enitrat: it's off by one again, got trolled by vscode

cairo/scripts/zkpi_to_eels.py
```diff
@@ -279,6 +279,7 @@ def sanity_check_fixture(fixture: Dict[str, Any]) -> None:
         parent_beacon_block_root=block.header.parent_beacon_block_root,
         excess_blob_gas=calculate_excess_blob_gas(chain.blocks[-1].header),
     )
+    logger.info("Patched state root: 0x" + output.state_root.hex())
```
enitrat, line 282: what's this here?

pyproject.toml
```diff
@@ -44,10 +45,12 @@ dev-dependencies = [
 ]
 
 [tool.uv.sources]
+ethereum = { git = "https://github.com/kkrt-labs/execution-specs.git", branch = "kakarot-changes" }
```
enitrat, line 48: this should be in the `mpt` package no?

python/mpt/src/mpt/py.typed
enitrat, line 1: what is it ?

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,243 @@
+import logging
+from typing import Dict, Optional
+
+from ethereum.cancun.fork_types import EMPTY_ACCOUNT, Account, Address
+from ethereum.cancun.state import State, get_account, get_storage
+from ethereum_types.bytes import Bytes32
+from ethereum_types.numeric import U256
+
+# Set up logger
+logger = logging.getLogger("mpt.state_diff")
```
enitrat, line 10:
```suggestion
logger = logging.getLogger()
```

not sure an arg is required?

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,243 @@
...
+
+    logger.debug(
+        f"Computing state diff: {len(pre_addresses)} pre addresses, {len(post_addresses)} post addresses"
+    )
+
+    # Find deleted accounts
+    deletions = pre_addresses - post_addresses
+    if deletions:
+        logger.error(
+            f"Warning: Since EIP-6780, we should not see any accounts marked for deletion at the level of the block - found {len(deletions)}"
```
enitrat, line 89: should this exit rather than log ?

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,302 @@
...
+    def test_get(self):
+        mpt = StateTries.from_json(f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}")
+
+        random_address = list(mpt.access_list.keys())[0]
+
+        key = keccak256(random_address)
+
+        result = mpt.get(key)
+
+        assert result is not None
```
enitrat, line 50: is `is not None` sufficient"? Don't we want expected values to prevent eventual regressions?

enitrat:
@CodiumAI-Agent /review

enitrat:
@CodiumAI-Agent /describe

cairo/scripts/stf_from_zkpi.py
```diff
@@ -0,0 +1,565 @@
...
+    # Get storage keys for this address from the access list
+    storage_keys = ethereum_state.access_list.get(address)
+    if not storage_keys:
+        logger.error(f"No storage keys in access list for address: 0x{address.hex()}")
+        return
+
+    # Get storage root for this address
+    storage_root = ethereum_state.get_storage_root(address)
+    if not storage_root:
+        print(f"No storage root found for address: 0x{address.hex()}")
```
enitrat, line 406:
```suggestion
        logger.error(f"No storage root found for address: 0x{address.hex()}")
```

pyproject.toml
Eikix: ah, well i didn't make changes to ethereum, where is the ethereum dependency in main?

python/mpt/src/mpt/py.typed
Eikix: no idea, it got generated by uv init

python/mpt/src/mpt/state_diff.py
Eikix: i've noticed edge cases where an empty account is in the pre-state and it is deleted in the post state

i could catch that case and exit if it's not in this case

cairo/scripts/zkpi_to_eels.py
Eikix: old print, dleeting

cairo/scripts/stf_from_zkpi.py
```diff
@@ -0,0 +1,565 @@
...
+    address : Address
+        The address to check
+    block_number : int
+        The block number to check at
+    """
+    from ethereum_rlp import rlp
+
+    logger = logging.getLogger("account_checksum")
+
+    print(f"\n🔍 Checking storage for address: 0x{address.hex()}")
```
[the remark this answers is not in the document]
Eikix: > @CodiumAI-Agent /improve use logger instead

waddaheck

pyproject.toml
enitrat: probably a rebase conflict on your branch then

Eikix:
> @CodiumAI-Agent /review

bruv what?

README.md
```diff
@@ -129,6 +129,31 @@ The script will load the ZKPI data for the specified block, convert it to the
...
 artifacts to the output directory.
 
+### Running State Transition Function from ZKPI
+
+Execute an Ethereum block stateless in Python. That is: execute and verify the
+Ethereum state transition function in Python (no-proof mode) using a ZK-Prover
+Input (ZKPI) file, use the `stf_from_zkpi.py` script:
+
+```bash
+uv run stf_from_zkpi <BLOCK_NUMBER>
```
ClementWalter, line 139: naming problem, we already have `zkpi_TO_xxx`, please don't introduce `xxx_FROM_yyy`

cairo/scripts/stf_from_zkpi.py
```diff
@@ -0,0 +1,565 @@
...
+from mpt import StateTries
+from mpt.state_diff import StateDiff
+
+
+def configure_logging():
+    logging.basicConfig(
+        level=logging.DEBUG,
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    )
+    logging.getLogger().setLevel(logging.DEBUG)
```
ClementWalter, line 38: why a function?

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,561 @@
...
+    U256,
+    Uint,
+)
+from scripts.zkpi_to_eels import normalize_transaction
+
+from mpt import StateTries
+from mpt.state_diff import StateDiff
+
+
+def stf_from_zkpi(block_number: int):
```
ClementWalter, line 33:
```suggestion
def zkpi_to_stf(block_number: int):
```

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,561 @@
...
+    If the state root does not match, it will attempt to identify the problematic address(es) and storage key(s) in the state.
+
+    Parameters
+    ----------
+    block_number : int
+        The block number to process
+
+    Requires CHAIN_RPC_URL to be set in the environment variables.
+    """
+    input_file = f"data/1/inputs/{block_number}.json"
```
ClementWalter, line 47: why not taking the file path in input instead of enforcing a given path ?

Eikix: good point

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,566 @@
...
+    If the state root does not match, it will attempt to identify the problematic address(es) and storage key(s) in the state.
+
+    Parameters
+    ----------
+    block_number : int
+        The block number to process
+
+    Requires CHAIN_RPC_URL to be set in the environment variables.
+    """
+    input_file = f"{path}/{block_number}.json"
```
ClementWalter, line 47: why 2 args ?

README.md
```diff
@@ -129,6 +129,31 @@ The script will load the ZKPI data for the specified block, convert it to the
...
 artifacts to the output directory.
 
+### Running State Transition Function from ZKPI
+
+Execute an Ethereum block stateless in Python. That is: execute and verify the
+Ethereum state transition function in Python (no-proof mode) using a ZK-Prover
+Input (ZKPI) file, use the `zkpi_to_stf.py` script:
+
+```bash
+uv run zkpi_to_stf <BLOCK_NUMBER>
```
ClementWalter, line 139: update since script was updated

README.md
```diff
@@ -129,6 +129,31 @@ The script will load the ZKPI data for the specified block, convert it to the
...
+```
+
+This command loads the ZKPI data for the specified block (it assumes the prover
+inputs for that block were generated previously and stored in
+data/1/inputs/<BLOCK_NUMBER>), executes all transactions in the block, and
+verifies that the computed state root matches the expected state root from the
+block header. The script performs detailed validation. If any discrepancies are
```
ClementWalter, line 146:
```suggestion
block header. If any discrepancies are
```

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,566 @@
...
+from ethereum_spec_tools.evm_tools.loaders.fixture_loader import Load
+from ethereum_spec_tools.evm_tools.loaders.fork_loader import ForkLoad
+from ethereum_spec_tools.evm_tools.loaders.transaction_loader import TransactionLoad
+from ethereum_types.bytes import Bytes, Bytes0, Bytes32
+from ethereum_types.numeric import (
+    U64,
+    U256,
+    Uint,
+)
+from scripts.zkpi_to_eels import normalize_transaction
```
ClementWalter, line 27: fix isort config, this is internal dep

cairo/scripts/zkpi_to_stf.py
Eikix: path is optional such that we can do both: `uv run zkpi_to_stf 22065671` and `uv run zkpi_to_stf 22065671 --path data/2/zkpi`

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,566 @@
...
+                        value=hex_to_u256(tx["value"]),
+                        data=Bytes(hex_to_bytes(tx["data"])),
+                        v=hex_to_u256(tx["v"]),
+                        r=hex_to_u256(tx["r"]),
+                        s=hex_to_u256(tx["s"]),
+                    )
+                    if isinstance(tx, dict)
+                    else Bytes(hex_to_bytes(tx))
+                )  # Non-legacy txs are hex strings
+                for tx in encoded_transactions
```
ClementWalter, line 130: why 2 loops ?

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,566 @@
...
+        state_diff = StateDiff.from_pre_post(pre_state_copy, post_state)
+        ethereum_state.update_from_state_diff(state_diff)
+        success = (
+            ethereum_state.state_root.hex()
+            == data["blocks"][0]["header"]["stateRoot"][2:]
+        )
+        assert (
+            ethereum_state.state_root.hex()
+            == data["blocks"][0]["header"]["stateRoot"][2:]
+        )
```
ClementWalter, line 193: I'd rather use `state_transition` directly, with a patched `state_root`. At least make sure to reproduce exactly its body, with a comment

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,566 @@
...
+        for ancestor in data["witness"]["ancestors"][::-1]
+    ]
+    blockchain = BlockChain(
+        blocks=blocks,
+        state=pre_state,
+        chain_id=U64(data["chainConfig"]["chainId"]),
+    )
+
+    # For each new block in blocks, create a Block object and apply the state transition
+    for block in data["blocks"]:
```
ClementWalter, line 89: why a loop

cairo/scripts/zkpi_to_stf.py
ClementWalter, line 1: so this script is more like `run_block_with_eels` or something like that

Eikix: yeah but it runs the block from a zkpi

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,299 @@
...
+
+# Standard test account
+TEST_ACCOUNT = Account(
+    nonce=Uint(10), balance=U256(1000), code=bytes.fromhex("deadbeef")
+)
+
+
+class TestStateTries:
+    def test_from_json(self):
+        mpt = StateTries.from_json(f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}")
```
ClementWalter, line 38: where pytest fixture?

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,299 @@
...
+        assert mpt is not None
+
+    def test_get(self):
+        mpt = StateTries.from_json(f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}")
+        state = mpt.to_state()
```
ClementWalter, line 43: don't "allocate" variable far before it's in use. And here, you can just call `mpt.to_state()` in the arg of `get_account`

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,299 @@
...
+
+        random_address = list(mpt.access_list.keys())[0]
```
ClementWalter, line 45: where fixture?

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,299 @@
...
+        ) == get_account(state, random_address)
+
+    def test_delete(self):
+        mpt = StateTries.from_json(f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}")
+
+        # Using an address from the JSON file that we know exists
+        test_address = Address(
+            bytes.fromhex("30325619135da691a6932b13a19b8928527f8456")
+        )
+
```
ClementWalter, line 63: no hard coded value

python/mpt/tests/test_mpt.py
```diff
@@ -0,0 +1,299 @@
...
+    def test_upsert(self):
+        mpt = StateTries.from_json(f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}")
+
+        encoded_account = encode_account(TEST_ACCOUNT, EMPTY_TRIE_ROOT_HASH)
+
+        # Using an address from the JSON file
+        test_address = Address(
+            bytes.fromhex("30325619135da691a6932b13a19b8928527f8456")
+        )
```
ClementWalter, line 78: no hardcoded value

python/mpt/tests/test_mpt.py
ClementWalter, line 1: rewrite everything with appropriate fixture and fuzzing when random accesses are needed

cairo/scripts/zkpi_to_stf.py
Eikix: proposing something in my next commit

README.md
```diff
@@ -129,6 +129,44 @@ The script will load the ZKPI data for the specified block, convert it to the
...
+   storage values
+
+This provides a way to validate Ethereum state transitions without maintaining a
+full node state, using only the specific trie nodes required for the block's
+execution.
+
+Usage:
+
+```bash
+python cairo/scripts/zkpi_to_stf.py <block_number> [--path <path_to_input_files>]
```
ClementWalter, line 153: I still don't understand why we want 2 args and not only a zkpi file path

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,559 @@
...
+        sys.exit(1)
+    except json.JSONDecodeError:
+        logging.error(f"Invalid JSON in input file: {input_file}")
+        sys.exit(1)
+    except Exception as e:
+        logging.error(f"Error loading state from file: {e}")
+        sys.exit(1)
+
+    with open(input_file, "r") as f:
+        data = json.load(f)
```
ClementWalter, line 68: we load the json file two times?

README.md
Eikix: it's mainly preference, i'm fine with adding just 1 arg

README.md
```diff
@@ -129,6 +129,44 @@ The script will load the ZKPI data for the specified block, convert it to the
...
+valid Ethereum node endpoint for verification purposes.
+
+If the state transition verification fails, the script will perform detailed
+diagnostics by comparing account and storage values against the expected values
+from the RPC node to help identify the source of the discrepancy.
+
+Requirements:
+
+- Block must be post-Cancun fork (block number ≥ 19426587)
+- ZKPI data must be available at `data/1/inputs/<BLOCK_NUMBER>.json`. To
```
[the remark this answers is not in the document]
Eikix: ah thats not true anymore

cairo/scripts/zkpi_to_stf.py
ClementWalter: I don't see the benefit of this if there is no call to a zkpi generate somewhere. In the current situation, the script needs a file, and I think that it's the only thing required, so there is no need to make it 2 variables

README.md
Eikix: but I prefer two args -> it's simpler to call

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,559 @@
...
+            for tx in block["transaction"]
+        )
+
+        block = Block(
+            header=load.json_to_header(block["header"]),
+            transactions=tuple(
+                (
+                    tx
+                    if isinstance(tx, LegacyTransaction)
+                    else Bytes(hex_to_bytes(encode_transaction(tx).hex()))
```
ClementWalter, line 100: bytes to hex to bytes ?

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,243 @@
+import logging
+from typing import Dict, Optional
+
+from ethereum.cancun.fork_types import EMPTY_ACCOUNT, Account, Address
+from ethereum.cancun.state import State, get_account, get_storage
+from ethereum_types.bytes import Bytes32
+from ethereum_types.numeric import U256
+
+# Set up logger
+logger = logging.getLogger()
```
ClementWalter, line 10: missing logger config

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,243 @@
+import logging
+from typing import Dict, Optional
+
+from ethereum.cancun.fork_types import EMPTY_ACCOUNT, Account, Address
+from ethereum.cancun.state import State, get_account, get_storage
+from ethereum_types.bytes import Bytes32
+from ethereum_types.numeric import U256
+
+# Set up logger
```
ClementWalter, line 9:
```suggestion
```

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,243 @@
...
+    A class that contains the differences between two states.
+    """
+
+    account_diffs: Dict[Address, AccountDiff]
+
+    def __init__(self):
+        self.account_diffs: Dict[Address, AccountDiff] = {}
+
+    @classmethod
+    def from_pre_post(cls, pre_state: State, post_state: State) -> "StateDiff":
```
ClementWalter, line 38:
```suggestion
    @staticmethod
    def from_pre_post(pre_state: State, post_state: State) -> "StateDiff":
```

ClementWalter: why moving the implementation out this function ?

python/mpt/src/mpt/state_diff.py
Eikix: i thought logger config is only written once in the 'main' of a script?

cairo/scripts/zkpi_to_stf.py
Eikix: fixed

python/mpt/src/mpt/state_diff.py
ClementWalter: checked with grok and it's not necessary to redo the config indeed, however at least `getLogger(__name__)` to have the module name in the message

python/mpt/src/mpt/state_diff.py
ClementWalter: I don't think that any such check belongs to a state diff function. this is rather a post processing check, that may or may not be used depending on the state's block

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,242 @@
...
+        Returns
+        -------
+        StateDiff
+            A StateDiff object containing:
+            - 'account_diffs': Dict[Address, AccountDiff] - Accounts that were added, modified or deleted with their storage changes
+        """
+        return compute_state_diff(pre_state, post_state)
+
+
+def compute_state_diff(pre_state: State, post_state: State) -> StateDiff:
```
ClementWalter, line 57: remove all logs related to fork version

python/mpt/src/mpt/state_diff.py
```diff
@@ -0,0 +1,242 @@
...
+    logger.debug(f"Total account diffs: {len(diff.account_diffs)}")
+
+    return diff
+
+
+def compute_account_diff(
+    pre_state: State,
+    post_state: State,
+    address: Address,
+) -> Optional[AccountDiff]:
```
ClementWalter, line 128: get the account and not the state as arg

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+                continue
+
+            rlp_account = RLPAccount(*rlp.decode(account))
+
+            if rlp_account.code_hash == EMPTY_CODE_HASH:
+                code = b""
+            else:
+                code = self.codes.get(rlp_account.code_hash, None)
+                # TODO: This is a hack to get the code for codes not present in the StateTries object
+                # This is due to accessing code only through EXTCODEHASH opcode
```
ClementWalter, line 151: due to the fact that the Account class in python doesn't match the account node structure: the class contains the full code and not only the code hash

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+        """
+
+        # Step 1: Recursively explore the state trie to get to the leaves
+        _main_trie: Trie[Address, Optional[Account]] = Trie(
+            secured=True, default=None, _data=defaultdict(lambda: None)
+        )
+        _storage_tries: Dict[Address, Trie[Bytes32, U256]] = dict()
+
+        logger.debug("Starting to convert StateTries to State")
+        for address in self.access_list.keys():
```
ClementWalter, line 134: why do we loop from the access_list and not from the root node?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+
+
+class ExclusionProof(Exception):
+    """Exception raised when a key is proven not to exist in the trie."""
+
+    pass
+
+
+@dataclass
+class RLPAccount:
```
ClementWalter, line 40:
```suggestion
class AccountNode:
```

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+        )
+
+    @classmethod
+    def from_json(cls, path: str) -> "StateTries":
+        logger.debug(f"Loading MPT from JSON file: {path}")
+        with open(path, "r") as f:
+            data = json.load(f)
+
+        nodes = {
+            keccak256(hex_to_bytes(node)): (hex_to_bytes(node))
```
ClementWalter, line 223:
```suggestion
            keccak256(hex_to_bytes(node)): hex_to_bytes(node)
```

python/mpt/src/mpt/__init__.py
Eikix: it's lighter to iterate over the access_list 
we could also traverse the trie

python/mpt/src/mpt/state_diff.py
Eikix: then we need the account & the post_state.storage_tries[address] in the args

python/mpt/src/mpt/state_diff.py
Eikix: do you mean the log elated to EIP6780? 

```python
logger.error(
            f"Warning: Since EIP-6780, we should not see any accounts marked for deletion at the level of the block - found {len(deletions)}"
        )
``` 
?

Eikix:
broke the tests, will investigate why tomorrow

error is most likely linked to https://github.com/kkrt-labs/keth/blob/bfb82e1ca43c53d719161352adb42ae244c8d98c/python/mpt/src/mpt/__init__.py#L837 and

cairo/scripts/zkpi_to_stf.py
Eikix: then perhaps the interesting twist is that it is running the block statelessly, eels itself is also an impl. detail

`run_block_stateless` 

?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1565 @@
...
+    @classmethod
+    def create_empty(cls) -> "StateTries":
+        return cls(
+            nodes={},
+            codes={},
+            access_list={},
+            state_root=EMPTY_TRIE_ROOT_HASH,
+        )
+
+    def __init__(
```
Eikix, line 108: perhaps if StateTries is a `@dataclass` you don't need such an init fn

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1565 @@
...
+        """
+        Convert the StateTries to a State object.
+
+        Returns
+        -------
+        State
+            A State object representing the Ethereum state
+        """
+
+        # Step 1: Recursively explore the state trie to get to the leaves
```
Eikix, line 134:
```suggestion
```

python/mpt/src/mpt/__init__.py
Eikix: btw why are we using _data=defaultdict?
isnt it redundant with the default field of Trie?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1565 @@
...
+            The path to look up
+        root_hash : Hash32, optional
+            The root hash to start from, defaults to the trie's state_root
+
+        Returns
+        -------
+        Optional[Bytes]
+            The value at the path, or None if not found
+        """
+        is_state_access = False
```
Eikix, line 314: nit: metadata meant for debugging, not sure this is needed

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1539 @@
...
+from ethereum_rlp import rlp
+from ethereum_types.bytes import Bytes, Bytes32
+from ethereum_types.numeric import U256, Uint
+
+from mpt.state_diff import StateDiff
+
+logger = logging.getLogger("mpt")
+
+
+class ExclusionProof(Exception):
```
Eikix, line 33: nit: should this really be an exception?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1539 @@
...
+                logger.debug("Path matches leaf key, returning value")
+                return node.value
+
+            raise ExclusionProof(
+                f"Path does not match leaf key: {nibble_path_to_hex(nibble_path)} != {nibble_path_to_hex(node.rest_of_key)}"
+            )
+
+        raise ValueError(f"Unknown node type: {type(node)}")
+
+    def _decode_node(self, node_data: Bytes) -> InternalNode:
```
Eikix, line 477: nit: move the impl of this function inside the body of the function

Eikix: and remove it from the class, since it doesn't use `self` ever

README.md
ClementWalter: yagni

python/mpt/src/mpt/state_diff.py
ClementWalter: yes, this is a Trie lib, focused on trie management and diff. These diff may or may not be consistent with EIPs, but it's another topic

python/mpt/src/mpt/__init__.py
ClementWalter: I doubt, it's the same as right below for codes

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+        Example of inserting 'abc':123 into a trie:
+        ```
+        Before:                After:
+        [Root]                [Root]
+        |                     |
+        [Branch]              [Branch]
+        |                   /    \
+        'd'               'ab'    'd'
+        |                 |      |
+        456              'c':123  456
```
ClementWalter, line 900: shouldn't it be `d:456` in the right and side?

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1553 @@
...
+        root_hash : Hash32, optional
+            The root hash to start from, defaults to the trie's state_root
+
+        Returns
+        -------
+        Hash32
+            The new root hash after insertion/update
+        """
+        if root_hash is None:
+            root_hash = self.state_root
```
ClementWalter, line 918: why this?

python/mpt/src/mpt/__init__.py
Eikix: yeah these cursor generated examples aren't accurate imo, deleting

python/mpt/src/mpt/__init__.py
Eikix: some kind of fn overload where if the root_hash isn't provided as argument in the function, then it's a state access

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,554 @@
...
+        # protocol only requires the last 255
+        chain.blocks = chain.blocks[-255:]
+
+
+def main():
+    logging.basicConfig(
+        level=logging.DEBUG,
+        format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+    )
+    logging.getLogger().setLevel(logging.DEBUG)
```
ClementWalter, line 530: at the module level

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,554 @@
...
+        List of storage keys to include in the proof
+
+    Returns
+    -------
+    dict
+        The storage proof data or None if there was an error
+    """
+    import os
+
+    import requests
```
ClementWalter, line 412: import at top level

cairo/scripts/zkpi_to_stf.py
```diff
@@ -0,0 +1,554 @@
...
+                print(
+                    f"  ✅ Storage key {key_hex}: Both state and RPC have zero/empty value"
+                )
+
+    print(
+        f"\nStorage check complete: {mismatches} mismatches found out of {len(proof['storageProof'])} keys"
+    )
+
+
+def get_storage_proof(address, block_number, storage_keys):
```
ClementWalter, line 392: eth_get_proof

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1464 @@
...
+            access_list={},
+            state_root=EMPTY_TRIE_ROOT_HASH,
+        )
+
+    def __init__(
+        self,
+        nodes: Mapping[Bytes32, Bytes],
+        codes: Mapping[Bytes32, Bytes],
+        access_list: Mapping[Address, Optional[List[Bytes32]]],
+        state_root: Hash32,
```
ClementWalter, line 113: `Hash32` VS `Bytes32`

python/mpt/src/mpt/__init__.py
```diff
@@ -0,0 +1,1464 @@
...
+    ):
+        self.nodes = nodes
+        self.codes = codes
+        self.access_list = access_list
+        self.state_root = state_root
```
ClementWalter, line 118: in init, should verify that state_root is in nodes

python/mpt/tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+)
+
+
+@pytest.fixture
+def test_file_path():
+    return f"{TEST_PATH}/{SUB_PATH}/{FILE_NAME}"
+
+
+@pytest.fixture
+def mpt_from_json(test_file_path):
```
ClementWalter, line 33: scope="session"

python/mpt/tests/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+    return StateTries.from_json(test_file_path)
+
+
+@pytest.fixture(scope="module")
```
ClementWalter, line 37:
```suggestion
```

Eikix:
UPDATE: How to move forward on this PR?

Problems & Challenges:
- too many new LoCs
- code quality not up to codebase standard
- logic that does not serve our codebase directly

Recommended solution:
- Break down the PR in smaller PRs that directly serve a purpose for keth.

Next actionable:
- Open a PR with an eth rpc minimal module with methods for eth_getProof, eth_getCode
- Open a PR with a way to convert a zkpi json file to a EthereumTries object and a method to flatten the tries to obtain a EELS state object.

PR #972, opened 2025-03-12
refactor: remove bool struct from G1Point

### LINKED ISSUE #971: refactor: replace bool in G1Point by felt
`G1Point__eq__` is using the `bool` for no apparent reason (only function, add friction), remove it.

zmalatrax:
Closes #971


enitrat:
This is part of our "strong type system", instead of using a felt (that could be any value) we use a boolean wrapper - this is consistent with all other functions that return either True or False

zmalatrax:
ahhh, okok
Then I'm closing it

PR #1005, opened 2025-03-14
feat: BNF12

### LINKED ISSUE #977: BNF12
# What 
BNF12 is necessary to implement the ec_pairing.
Those class/struct need to be implemented in cairo 

# How

Create a file `src/ethereum/crypto/altbn128.cairo` with a struct representing a field element of `BNF12` along with `w`, `i_plus_9` and frob coeffs: 

```
struct BNF12 {
    c0: felt,
    c1: felt, 
    ...
    c12: felt
}

func FROBENIUS_COEFFICIENTS() {
    return <BNF12.calculate_frobenius_coefficients> // no need to implement the function, calculate those coeff in advance
}

// It will not really be used in our case because we will define constants instead of recomputing values
func BNF12_W() {
 return <BNF12((0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0))>
}


// Used in twist function
func BNF12_W_2() {
  return <BNF12.w**2>
}

// Used in twist function
func BNF12_W_3() {
  return <BNF12.w**3>
}

// Used in bnf2_to_bnf12
func BNF12_i_plus_9() {
   return <BNF12.w**6>
}

```

obatirou:
Close #977

### DIFF
--- a/cairo/tests/utils/serde.py
+++ b/cairo/tests/utils/serde.py
@@ -433,12 +433,10 @@ class Serde(SerdeProtocol):
             if value is None:
                 return None
 
-            coeffs = []
-            for i in range(12):
-                low = value[f"c{i}"]["low"]
-                high = value[f"c{i}"]["high"]
-                coeff_value = uint256_to_int(low, high)
-                coeffs.append(coeff_value)
+            coeffs = [
+                uint256_to_int(value[f"c{i}"]["low"], value[f"c{i}"]["high"])
+                for i in range(12)
+            ]
 
             return BNF12(tuple(coeffs))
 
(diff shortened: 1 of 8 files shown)


obatirou:
Seems to be impacted by [#1006](https://github.com/kkrt-labs/keth/issues/1006)

cairo/tests/utils/serde.py
```diff
@@ -425,6 +427,21 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
         if python_cls in (Bytes0, Bytes1, Bytes4, Bytes8, Bytes20):
             return python_cls(value.to_bytes(python_cls.LENGTH, "little"))
 
+        if python_cls == BNF12:
+            # BNF12 is represented as a struct with 12 Uint256
+            # We need to extract these values and create a BNF12 object
+            if value is None:
+                return None
+
+            coeffs = []
```
enitrat, line 436:
```suggestion
            coeffs = [uint256_to_int(value[f"c{i}"]["low"], value[f"c{i}"]["high"]) for i in range(12)]
```

enitrat (approved):

PR #1022, opened 2025-03-18
refacto: change BNF12 coefficients type from Uint256 to UInt384

### LINKED ISSUE #1018: refacto: use UInt384 for coefficients of BNF12 instead of UInt256
# Why
We chose to use UInt256 for coefficients of `BNF12`, as it fits on 256 bits, hence better memory use.
However, as we use ModBuiltin circuits, it requires to cast all coefficients to UInt384 to use them, then cast them back to UInt256 to return a BNF12 ; all these casts end up using more cells than directly using UInt384 for BNF12 coefficients. Plus it bloats the BNF12 methods with that much casts..

# What
Replace the coefficients type from UInt256 to UInt384, and corresponding tests accordingly.

zmalatrax:
Closes #1018 

Frobenius coefficients obtained by running `int_to_uint384` on each coefficients of each element of EELS's `BNF12.FROBENIUS_COEFFICIENTS`


cairo/tests/utils/serde.py
obatirou, line 1: You will need to change args_gen also

obatirou (approved):

PR #1031, opened 2025-03-18
feat: implement bnf12_scalar_mul

### LINKED ISSUE #997: scalar_mul
# What 
Implement scalar_mul operation for BNF12 field element

# How 
Replicate exec specs
 
https://github.com/ethereum/execution-specs/blob/1a587803e3e698407d204888b02342393f8b4fe5/src/ethereum/crypto/finite_field.py#L301

Use `mul` from `mod_ops_compiled.cairo`

zmalatrax:
Closes #997 

It also reorders `bnf12_sub` which was declared after the `BNP12` struct

### DIFF
--- a/cairo/ethereum/crypto/alt_bn128.cairo
+++ b/cairo/ethereum/crypto/alt_bn128.cairo
@@ -1,7 +1,9 @@
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.cairo_builtins import UInt384, ModBuiltin
+from starkware.cairo.lang.compiler.lib.registers import get_fp_and_pc
 
-from cairo_ec.circuits.mod_ops_compiled import add, sub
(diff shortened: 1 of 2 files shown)


enitrat (approved):
plz move to  bnf12.cairo and bnp12.cairo files later on

obatirou:
> > plz move to bnf12.cairo and bnp12.cairo files later on
> 
> @enitrat `bnp12.cairo` will use `bnf12.cairo` (i.e. `from bnf12 import ...`), and `alt_bn128.cairo` will use both `bnf12.cairo` and `bnp12.cairo`
> 
> --> There will be cyclic dependencies, no? (cc. @obatirou)

Yes as discussed that is why we need to keep everything in the same file for now.
We might be able to port some functions to utils

PR #1038, opened 2025-03-19
refactor: use U384 for BNF12

### LINKED ISSUE #1033: dev: use U384 for BNF12
# What
U384 should be used instead of UInt384 to match the type system and avoid specific case.

# How
Modify altbn128.cairo, arg_gen and serde

obatirou:
Close #1033

### DIFF
--- a/cairo/ethereum/crypto/alt_bn128.cairo
+++ b/cairo/ethereum/crypto/alt_bn128.cairo
@@ -9,18 +9,18 @@ from cairo_ec.curve.alt_bn128 import alt_bn128
 // BNF12 represents a field element in the BNF12 extension field
 // This is a 12-degree extension of the base field used in alt_bn128 curve
 struct BNF12Struct {
-    c0: UInt384,
-    c1: UInt384,
-    c2: UInt384,
-    c3: UInt384,
-    c4: UInt384,
-    c5: UInt384,
-    c6: UInt384,
-    c7: UInt384,
-    c8: UInt384,
-    c9: UInt384,
-    c10: UInt384,
-    c11: UInt384,
+    c0: U384,
+    c1: U384,
+    c2: U384,
+    c3: U384,
+    c4: U384,
+    c5: U384,
+    c6: U384,
+    c7: U384,
+    c8: U384,
+    c9: U384,
+    c10: U384,
+    c11: U384,
 }
 
 struct BNF12 {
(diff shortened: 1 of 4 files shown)


cairo/ethereum/crypto/alt_bn128.cairo
```diff
@@ -45,235 +45,340 @@ func FROBENIUS_COEFFICIENTS() -> TupleBNF12 {
...
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
-            UInt384(0, 0, 0, 0),
+            U384(new UInt384(1, 0, 0, 0)),
+            U384(new UInt384(0, 0, 0, 0)),
```
enitrat, line 49: for all U384_ZERO values we should have a label, like we do with `U256_ZERO`, and use `let (u384_zero) = get_label_location(U256_ZERO);` then i think we can do

`U384(cast(u384_zero, UInt384*))`

cairo/tests/utils/serde.py
```diff
@@ -427,28 +426,16 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
...
+            coeffs = [int(value[f"c{i}"]) for i in range(12)]
 
             return BNF12(tuple(coeffs))
 
         if python_cls == BNP12:
             # BNP12 is represented as a struct with two BNF12 values (x and y)
-            if value is None:
-                return None
-            # Create a new BNP12 point with these coordinates
             return BNP12(value["x"], value["y"])
```
enitrat, line 439: you should be able to remove this entirely, as this is supposed to be caught by the fallback case

cairo/tests/utils/serde.py
```diff
@@ -427,28 +426,16 @@ def serialize_type(self, path: Tuple[str, ...], ptr) -> Any:
...
-                    value[f"c{i}"]["d0"],
-                    value[f"c{i}"]["d1"],
-                    value[f"c{i}"]["d2"],
-                    value[f"c{i}"]["d3"],
-                )
-                for i in range(12)
-            ]
+            # In python, BNF12 is a tuple of 12 int but in cairo it's a struct with 12 U384
+            # Cast the U384 to int to be able to serialize
+            coeffs = [int(value[f"c{i}"]) for i in range(12)]
```
enitrat, line 433: isnt it just value.values()

enitrat: in which case this could simply be

```
        if python_cls == BNF12:
            value = value.values()
```

and then let the `signature(python_cls.__init__).bind(None, *value)` take over

obatirou: You need to cast to int as you receive a dict of U384 but BNF12 accept a tuple of int

enitrat: but it doesnt really matter

enitrat (approved):

PR #1054, opened 2025-03-19
feat: convert ethereum MPT to EELS state

### LINKED ISSUE #1028: feat: EthereumTries to EELS State
## Why

we need to load a pre state starting from zkpi

## What

a to_state method that generates a State object

## How

1/ trie traversal 
a/ when reaching a leaf, look up path == keccak(address) to get address
b/ get storage trie root hash from decoded account 
c/ storage trie traversal
d/ when reaching a leaf node, look up path == keccak(storage_key) to get storage_key

### If Help needed
break down issue in two PRs

Eikix:
- a ethereum tries function called `to_state` that traverses the ethereum tries to construct a state object
- a test that attempts to run EELS' `state_transition` on this newly constructed state object

closes #1028

### DIFF
--- a/python/mpt/src/mpt/ethereum_tries.py
+++ b/python/mpt/src/mpt/ethereum_tries.py
@@ -1,8 +1,9 @@
 import json
 import logging
 from dataclasses import dataclass
+from functools import partial
 from pathlib import Path
-from typing import Any, Dict, Mapping, Optional
+from typing import Any, Callable, Dict, Mapping, Optional
 
 from ethereum.cancun.fork_types import Address
 from ethereum.cancun.state import State, set_account, set_storage
@@ -116,7 +117,10 @@ class EthereumTries:
         )
 
     def traverse_trie_and_process_leaf(
-        self, node: InternalNode, current_path: Bytes, process_leaf: callable, **kwargs
+        self,
+        node: InternalNode,
+        current_path: Bytes,
+        process_leaf: Callable,
     ) -> None:
         """
         Recursive trie traversal function with a callback for each leaf node.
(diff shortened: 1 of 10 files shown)


python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -71,3 +108,181 @@ def from_data(data: Dict[str, Any]):
...
+                self.traverse_trie(next_node, current_path, process_leaf, **kwargs)
+                return
+            # if subnode is less than 32 bytes, it's an embedded (RLP-encoded) node
+            if len(node.subnode) < 32:
+                next_node = decode_node(node.subnode)
+                self.traverse_trie(next_node, current_path, process_leaf, **kwargs)
+                return
+
+            if len(node.subnode) > 32:
+                raise ValueError(f"Invalid subnode length: {len(node.subnode)}")
```
enitrat, line 193:
```suggestion
            if len(node.subnode) > 32:
                raise ValueError(f"Invalid subnode length: {len(node.subnode)}")
                
            # subnode is a hash, so we need to resolve it
            next_node = self.nodes.get(node.subnode) if len(node.subnode) == 32 else decode_node(node.subnode)
            if not next_node:
                # If the subnode is not found, we assume this path
                # is not needed for block execution
                return
            return self.traverse_trie(next_node, current_path, process_leaf, **kwargs)
```

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -71,3 +108,181 @@ def from_data(data: Dict[str, Any]):
...
+        value = rlp.decode(node.value)
+        set_storage(
+            state, account_address, storage_key, U256(int.from_bytes(value, "big"))
+        )
+
+    def resolve(self, node: InternalNode, current_path: Bytes, state: State) -> None:
+        logger.debug(f"Resolving node at path 0x{current_path.hex()}")
+        return self.traverse_trie(
+            node, current_path, self.process_account_leaf, state=state
+        )
```
enitrat, line 268: looks useless, could be inlined

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -71,3 +108,181 @@ def from_data(data: Dict[str, Any]):
...
+            logger.debug(
+                f"Storage root node not found for account 0x{address.hex()}, skipping"
+            )
+            return
+        logger.debug(
+            f"Storage root node found for 0x{address.hex()}, opening storage trie"
+        )
+        self.resolve_storage(storage_root_node, b"", state, address)
+
+    def process_storage_leaf(
```
enitrat, line 243:
```suggestion
    def set_storage_from_leaf(
```

nit: function name should indicate mutation of state

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -71,3 +108,181 @@ def from_data(data: Dict[str, Any]):
...
+        if isinstance(node, LeafNode):
+            logger.debug(
+                f"Traversing leaf node with current path 0x{nibble_list_to_bytes(current_path + node.rest_of_key).hex()}"
+            )
+            full_path = nibble_list_to_bytes(current_path + node.rest_of_key)
+            return process_leaf(node, full_path, **kwargs)
+
+        return
+
+    def process_account_leaf(
```
enitrat, line 204:
```suggestion
    def set_account_from_leaf(
```

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -71,3 +108,181 @@ def from_data(data: Dict[str, Any]):
...
+        if isinstance(node, BranchNode):
+            logger.debug(
+                f"Traversing branch node with current path 0x{nibble_list_to_bytes(current_path).hex()}"
+            )
+            for i in range(16):
+                nibble = bytes([i])
+                subnode = node.subnodes[i]
+                # We skip empty nodes
+                if not subnode:
+                    continue
```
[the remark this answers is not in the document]
Eikix: > for subnode in subnodes ?

we need the nibble index, we could use enumerate

Issue #1013, opened 2025-03-17
feat: Create a Python function to compute the diff between two EthereumTries


enitrat:
#### Description:  
Develop a Python reference implementation that traverses an initial Merkle Patricia Trie (`M`) and a final Merkle Patricia Trie (`M'`) to compute the state diff. The function will identify added, deleted, and modified key-value pairs, producing a structured output that can be used to verify the correctness of the State Transition Function (STF). This implementation is inspired by an existing Go implementation (`compareTrie`), adapting its recursive traversal logic to Python while focusing on a simplified, clear diff output, to be then implemented in Cairo.

> Note: the main purpose of this function is to serve as a reference implementation for diff-testing in Cairo. Any implementation choice __MUST__ take this into account.

#### Acceptance Criteria:  
1. The `__sub__`  implementation for two `EthereumTries`
```py
class EthereumTries:
    nodes: Mapping[Hash32, InternalNode]
    codes: Mapping[Hash32, Bytes]
    address_preimages: Mapping[Hash32, Address]
    storage_key_preimages: Mapping[Hash32, Bytes32]
    state_root: Hash32
```

```py

@dataclass
class StateDiff:
    """
    Contains all information that is preserved between transactions.
    """

    _main_trie: Dict[Address, Tuple[Optional[Account], Optional[Account]]]
    _storage_tries: Dict[Address, Tuple[Dict[Bytes32, U256], Dict[Bytes32, U256]]]

```

2. The function outputs a`StateDiff`
4. For each key in `M` and `M'`:   - If the key is different in `M'` compared to `M`, (either modified, added or deleted); add it to the diff
4. The implementation handles MPT node types recursively. The Go implementation handles it this way
   - `nil`: Empty node (no value).
   - `valueNode`: Leaf node with a value.
   - `shortNode`: Extension node
   - `fullNode`: Branch node with 16 children (nibbles 0-f).
5. Edge cases are handled:
   - Empty `M` or `M'` (e.g., `nil` root).
   - Identical `M` and `M'` (empty diff).
6. Unit tests verify:
   - Adding a new key produces a diff
   - Removing a key produces a diff
   - Changing a value produces a diff
   - No changes produce an empty diff

> Note: In https://github.com/kkrt-labs/keth/pull/962 we have a StateDiff object that we could use as the reference output format.

## Reference implementation
https://github.com/kkrt-labs/go-ethereum/pull/1

gabito1451:
[...]

PR #1056, opened 2025-03-19
feat: trie diff

enitrat:
closes [#1013](https://github.com/kkrt-labs/keth/issues/1013)


python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+        r_name = r_node.__class__.__name__ if r_node else "None"
+
+        match l_name, r_name:
+            case "None", "None":
+                # No change
+                pass
+
+            case "None", "LeafNode":
+                # new leaf
+                preimage = nibble_path_to_bytes(path + r_node.rest_of_key)
```
Eikix, line 85: are you sure this is the entire path?

imagine you are half way thru your trie, your path is 0x123456

if your full path to a leaf is 0x123456abcde
the leaf' rest_of_key will be 0xabcde

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+            case "ExtensionNode", "ExtensionNode":
+                l_node: ExtensionNode = cast(ExtensionNode, l_node)
+                r_node: ExtensionNode = cast(ExtensionNode, r_node)
+
+                # Equal keys -> Look for diffs in childrens
+                if l_node.key_segment == r_node.key_segment:
+                    self._compute_diff(l_node.subnode, r_node.subnode, path + l_node.key_segment)
+
+                # Right is prefix of left -> Look for diffs in left sub-tree
+                elif l_node.key_segment.startswith(r_node.key_segment):
```
Eikix, line 139: why do we stop exploring the right trie?
in my understanding, if it is a prefix of left, then it means a branch node was inserted at the divergence point between l and r, which suggests there is a diff inside of it too

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+                        # Look for diffs in other branches
+                        self._compute_diff(l_node.subnode, r_node.subnodes[i], path + nibble)
+
+            case "BranchNode", "None":
+                # Look for diffs in all branches of the left sub-tree
+                l_node: BranchNode = cast(BranchNode, l_node)
+                for i in range(0, 16):
+                    self._compute_diff(l_node.subnodes[i], None, path + i.to_bytes(1, "big"))
+
+            case "BranchNode", "LeafNode":
```
Eikix, line 174: why not?

branch with two branches, both branches leading to a leaf
if you delete one branch, then you have to perform a branch node reduction into a single leaf

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+                raise ValueError("BranchNode -> LeafNode should not happen if comparing two valid trie roots")
+
+            case "BranchNode", "ExtensionNode":
+                l_node: BranchNode = cast(BranchNode, l_node)
+                r_node: ExtensionNode = cast(ExtensionNode, r_node)
```
Eikix, line 179: haunted by the cast keyword, no matter where we go

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+                    else:
+                        # Look for diffs in other branches
+                        self._compute_diff(l_node, None, path + nibble)
+
+            case "BranchNode", "BranchNode":
+                # Look for diffs in all branches of the right sub-tree
+                l_node = cast(BranchNode, l_node)
+                r_node = cast(BranchNode, r_node)
+                for i in range(0, 16):
+                    l_hash = l_node.subnodes[i]
```
Eikix, line 196: iirc, there can exist embedded nodes here, i.e. there might exist i such that l_node.subnodes[i] := rlp.encode(node) and not digest(node)

if len(rlp.encode(node)) < 32

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+        r_root = tries.post_state_root
+
+        diff._compute_diff(l_root, r_root, Bytes())
+        return diff
+
+    def _compute_diff(self, left: Optional[Hash32], right: Optional[Hash32], path: Bytes):
+        if left == right:
+            return
+
+        l_node = self._nodes.get(left) if left else None
```
Eikix, line 72: grok tells me you can try:
```python
def process_node(node):
    match node:
        case BranchNode(subnodes, value) if len(subnodes) > 10:
            return "Large BranchNode"
        case BranchNode(subnodes, value):
            return f"Small BranchNode with {len(subnodes)} subnodes"
        case ExtensionNode(key_segment, _) if key_segment.startswith(b"key"):
            return "ExtensionNode with 'key' prefix"
        case ExtensionNode(_, subnode):
            return f"ExtensionNode with subnode {subnode}"
        case _:
            return "Unknown node type"
```

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,201 @@
...
+
+            case "LeafNode", "ExtensionNode":
+                raise ValueError("LeafNode -> ExtensionNode should not happen if comparing two valid trie roots")
+
+            case "ExtensionNode", "None":
+                # Look for diffs in the left sub-tree
+                l_node: ExtensionNode = cast(ExtensionNode, l_node)
+                self._compute_diff(l_node.subnode, None, path + l_node.key_segment)
+
+            case "ExtensionNode", "LeafNode":
```
Eikix, line 127: left: ext -> branch -> leaf A & leaf B
right: deletion of leaf A, then branch converted into ext, ext to leaf converted into unique leaf

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                        state_diff._storage_tries[address] = {}
+                    state_diff._storage_tries[address][key] = tuple((pre, post))
+
+            state_diff._main_trie[address] = tuple((pre_account, post_account))
+
+        return state_diff
+
+    @classmethod
+    def from_tries(cls, tries: EthereumTrieTransitionDB) -> "StateDiff":
+        # Merge all mappings of hash -> node to have a single DB to fetch nodes and resolve addresses from
```
Eikix, line 100: outdated i think

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+        self,
+        left: Optional[Hash32 | Bytes],
+        right: Optional[Hash32 | Bytes],
+        path: Bytes,
+        process_leaf_diff: Callable,
+    ):
+        if left == right:
+            return
+
+        def resolve(node: Hash32 | Bytes | None | InternalNode) -> InternalNode | None:
```
Eikix, line 121: i don't mind nesting fns sometimes but 3 levels of nesting is too much

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+            if isinstance(node, InternalNode):
+                return node
+            if not node or node == b"":
```
Eikix, line 124: `not b""` casts it to truthy

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                return None
+            if len(node) == 32:
```
Eikix, line 126: what is the behaviour of calling len on an Internal Node?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+
+        # Use direct class pattern matching
+        match (l_node, r_node):
+            case (None, None):
+                # No change
+                pass
+
+            case (None, LeafNode()):
+                # new leaf
+                preimage = nibble_path_to_bytes(path + r_node.rest_of_key)
```
Eikix, line 141: that's not a preimage, that's the 'post image'

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                    )
+
+            case (LeafNode(), None):
+                # deleted leaf (should not happen post-cancun)
+                preimage = nibble_path_to_bytes(path + l_node.rest_of_key)
+                process_leaf_diff(preimage, l_node, None)
+
+            case (LeafNode(), LeafNode()):
+                if l_node.value != r_node.value:
+                    preimage = nibble_path_to_bytes(path + l_node.rest_of_key)
```
Eikix, line 167: should we check an invariant here, that l_node rest of key == r_node rest of key?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                    process_leaf_diff(preimage, l_node, r_node)
+
+            case (LeafNode(), BranchNode()):
+                # Look for diffs in all branches of the right sub-tree
+                for i in range(0, 16):
+                    if (
+                        i != l_node.rest_of_key[0]
+                        or not self._nodes[r_node.subnodes[i]].value == l_node.value
```
Eikix, line 175: in case some subnodes are digests that aren't meant to be opened / resolved, this dict access might fail

Eikix: moreover, `value` attribute is only available for branch and leaf nodes, not extension nodes

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                            None,
+                            r_node.subnodes[i],
+                            path + bytes([i]),
+                            process_leaf_diff,
+                        )
+
+            case (LeafNode(), ExtensionNode()):
+                # Explore the extension node's subtree for any new leaves, comparing it to the old
+                # leaf with the same key
+                l_node = LeafNode(
```
Eikix, line 187: here we assume that len(r_node.key_segment) < len(l_node.rest_of_key), can we add an invariant?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+            case (ExtensionNode(), None):
+                # Look for diffs in the left sub-tree
+                self._compute_diff(
+                    l_node.subnode, None, path + l_node.key_segment, process_leaf_diff
+                )
+
+            case (ExtensionNode(), LeafNode()):
+                # The extension node was deleted and replaced by a leaf - meaning that down the line of the extension node, in a branch, we deleted some nodes.
+                # Explore the extension node's subtree for any deleted nodes, comparing it to the new leaf
+                r_node = LeafNode(
```
Eikix, line 203: here we assume that len(r_node.rest_of_key) > len(l_node.key_segment), can we add an invariant?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                    self._compute_diff(
+                        l_node.subnode,
+                        r_node.subnode,
+                        path + l_node.key_segment,
+                        process_leaf_diff,
+                    )
+                # Right is prefix of left
+                elif l_node.key_segment.startswith(r_node.key_segment):
+                    # Compare the right node's value with the left node shortened by right key
+                    l_node.key_segment = l_node.key_segment[len(r_node.key_segment) :]
```
Eikix, line 222: here we mutate the l_node, but above we reassign in Extension vs. Leaf, we reassign the leaf node to a new instnace of the leafnode object, should we choose one way?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -0,0 +1,388 @@
...
+                            process_leaf_diff,
+                        )
+
+            case (BranchNode(), None):
+                # Look for diffs in all branches of the left sub-tree
+                for i in range(0, 16):
+                    self._compute_diff(
+                        l_node.subnodes[i],
+                        None,
+                        path + i.to_bytes(1, "big"),
```
Eikix, line 282:
```suggestion
                        path + bytes([i])
```
nit btw, if you prefere i.to_bytes, ok for me

Eikix:
Tests pass! yay
Also I noticed that for the test data we have, we pass tests even if we delete certain parts of the code, should we try to pass 10 blocks first locally?

python/mpt/src/mpt/trie_diff.py
enitrat: cannot happen because of condition 1

python/mpt/src/mpt/trie_diff.py
enitrat: yeah idk why but some nodes are frozen and cant be mutated while others can

python/mpt/src/mpt/trie_diff.py
enitrat: the slice operation would fail otherwise, thus it seems un-necessary

python/mpt/src/mpt/trie_diff.py
```diff
@@ -153,30 +144,31 @@ def resolve(node: Hash32 | Bytes | None | InternalNode) -> InternalNode | None:
...
-                preimage = nibble_path_to_bytes(path + l_node.rest_of_key)
-                process_leaf_diff(preimage, l_node, None)
+                full_path = nibble_path_to_bytes(path + l_node.rest_of_key)
+                process_leaf_diff(full_path, l_node, None)
 
             case (LeafNode(), LeafNode()):
                 if l_node.value != r_node.value:
-                    preimage = nibble_path_to_bytes(path + l_node.rest_of_key)
-                    process_leaf_diff(preimage, l_node, r_node)
+                    assert (
```
Eikix, line 157: makes me wonder if this case is actually possible, in case of a deletion and an insertion inside the same branch node, at the same nibble

pre state - leaf that is a branch subnode at nibble 0, rest of key is abcde
post state - leaf that is a branch subnode at nibble 0, rest of key is deade

python/mpt/src/mpt/trie_diff.py
```diff
@@ -386,3 +378,15 @@ def _process_storage_diff(
             if address not in self._storage_tries:
                 self._storage_tries[address] = {}
             self._storage_tries[address][key] = tuple((left_decoded, right_decoded))
+
+
+def resolve(
```
Eikix, line 389: i dont think resolve is used with internal node as argument

enitrat: it is

Eikix (approved):
gg!!!

PR #1057, opened 2025-03-20
feat: BNF2 and BNF2 addition

### LINKED ISSUE #1042: BNF2
Add a struct to represent the BNF2 field (quadratic extension of BNF)

Something like
```cairo
struct BNF2 {
  a: U384,
  b: U384
}

### LINKED ISSUE #1053: feat: bnf2_add
Implement addition between two BNF2 elements

EELS specs
https://github.com/ethereum/execution-specs/blob/9c58cc8553ec3a59e732e81d5044c35aa480fbbb/src/ethereum/crypto/finite_field.py#L226-L236

### PR DESCRIPTION
**Pull request by obatirou**
Close #1042 
Close #1053

### DIFF
--- a/cairo/tests/test_serde.py
+++ b/cairo/tests/test_serde.py
@@ -30,7 +30,7 @@ from ethereum.cancun.vm.exceptions import (
 )
 from ethereum.cancun.vm.gas import ExtendMemory, MessageCallGas
 from ethereum.cancun.vm.interpreter import MessageCallOutput
-from ethereum.crypto.alt_bn128 import BNF12, BNP12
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: cairo/tests/test_serde.py [hunk 1]
```diff
@@ -295,6 +292,7 @@ def test_type(
             BNF12,
             Tuple[BNF12, ...],
             BNP12,
+            BNF2,
```
zmalatrax, line 295: Do we care about any order here ?
It seems to simply be appended every time a new class is added so I guess not 🤔

zmalatrax (approved):

PR #1078, opened 2025-03-21
feat: add storage root to account

### LINKED ISSUE #1074: refactor: modify `account_has_storage` to use storage root checks and add `storage_root` to Account type
### Description

The function `account_has_storage` in `cairo/ethereum/cancun/state.cairo` is currently used to verify non-collision during deployment transactions (`CREATE`/`CREATE2`) by checking if an account has storage. Geth uses the state root for these checks, but our implementation relies on `get_storage_keys_for_address`, which is unsound and inefficient.

Moreover, because we're working with Prover Inputs, which represent a partial state that is accessed during execution, the storage_tries we use might not contain other values than those accessed during the transactions, which would make `has_storage` return `False` even if the account has storage.

An important thing to note is that, since the Cancun fork, a newly created account has its nonce set to 1 and some code, and if no code is associated, it implies no storage exists. Thus, either an account has a storage root BEFORE the transaction (legacy, pre spurious dragon) OR it has a nonce or code.

Thus, we should align with Geth's approach by leveraging the storage root, which would solve our issue of unsoundness.

Additionally, to fully support this, we need to extend the `Account` type to include a `storage_root` field (representing the root of the storage trie), which is currently missing. This requires updates to both the Cairo `Account` struct in `fork_types.cairo` and the Python `Account` type in `args_gen.py`, along with proper import patching in `conftest.py`.

---

### Acceptance Criteria

1. **Modify `account_has_storage`:**
   - Update `account_has_storage` in `cairo/ethereum/cancun/state.cairo` to check the storage root instead of relying on `get_storage_keys_for_address`.
   - Use the `storage_roots` field of the account to retrieve the storage root.
   - Return `false` if the storage root is equal to `EMPTY_ROOT` indicating no prior storage.
   - Remove the unsound `get_storage_keys_for_address` call (noted with `TODO: this is unsound` in the current code).

2. **Extend `Account` Type in Cairo:**
   - In `cairo/ethereum/cancun/fork_types.cairo`, add a `storage_root: Hash32` field to the `AccountStruct` definition.
   - Update `EMPTY_ACCOUNT()` to initialize `storage_root` to the hash of empty storage  root `0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421`
   - Adjust `Account__eq__` to compare the `storage_root` field. 

3. **Extend `Account` Type in Python:**
   - In `cairo/tests/utils/args_gen.py`, overload the `Account` class from `ethereum.cancun.fork_types` to include a `storage_root: Hash32` field.
   - Ensure serialization in `_gen_arg` handles the new `storage_root` field correctly.

4. **Patch Imports:**
   - In `cairo/tests/conftest.py`, patch the import of `ethereum.cancun.fork_types.Account` to use the overloaded `Account` class with `storage_root`.
   - Verify that all dependent modules use the updated `Account` type.

5. **Testing:**
   - Add a test case to verify that `account_has_storage` returns `false` for an account with an empty storage root.
   - Adapt any required tests.

enitrat:
Closes #1074


cairo/tests/conftest.py
```diff
@@ -79,13 +79,37 @@ def pytest_configure(config):
...
+    # See explanation below. Lots of EELS modules import `Account` and `EMPTY_ACCOUNT` from `ethereum.cancun.fork_types`.
+    # I think these modules get loaded before this patch is applied. Thus we must replace them manually.
+    setattr(ethereum.cancun.trie, "Account", Account)
+    setattr(ethereum.cancun.state, "Account", Account)
+    setattr(ethereum.cancun.state, "EMPTY_ACCOUNT", EMPTY_ACCOUNT)
+    setattr(ethereum.cancun.fork_types, "EMPTY_ACCOUNT", EMPTY_ACCOUNT)
+    setattr(ethereum.cancun.vm.instructions.environment, "EMPTY_ACCOUNT", EMPTY_ACCOUNT)
+
+    ethereum.cancun.trie.Node = Node
+    setattr(tests.utils.args_gen, "Node", Node)
```
Eikix, line 111: why do we need to replace args_gen's Node by itself?

enitrat:
```suggestion
```

cairo/tests/utils/strategies.py
```diff
@@ -44,9 +51,18 @@
 from starkware.cairo.lang.cairo_constants import DEFAULT_PRIME
 
 from cairo_ec.curve import AltBn128
-from tests.utils.args_gen import (
+
+# Note: I have noticed that even if we path the imports in conftests.py, because hypothesis runs before these patches are applied,
```
Eikix, line 55:
```suggestion
# Note: I have noticed that even if we patch the imports in conftests.py, because hypothesis runs before these patches are applied,
```

cairo/tests/utils/strategies.py
```diff
@@ -447,49 +470,60 @@ def bounded_u256_strategy(min_value: int = 0, max_value: int = 2**256 - 1):
...
-state = st.lists(address, max_size=MAX_ADDRESS_SET_SIZE, unique=True).flatmap(
-    lambda addresses: st.builds(
-        State,
-        _main_trie=st.builds(
+
+@st.composite
+def state_strategy(draw):
+    addresses = draw(st.lists(address, max_size=MAX_ADDRESS_SET_SIZE, unique=True))
+
+    _main_trie = draw(
```
Eikix, line 478: i think you can build this trie after the storage trie is built so that you don't need to update accounts

cairo/tests/utils/strategies.py
```diff
@@ -65,7 +81,8 @@
 uint64 = st.integers(min_value=0, max_value=2**64 - 1).map(U64)
 uint = uint64.map(Uint)
 uint128 = st.integers(min_value=0, max_value=2**128 - 1)
-felt = st.integers(min_value=0, max_value=DEFAULT_PRIME - 1)
+felt = st.integers(min_value=-DEFAULT_PRIME // 2, max_value=DEFAULT_PRIME // 2)
```
Eikix, line 84: this felt is always equal to default prime // 2, is it intentional?

cairo/tests/utils/strategies.py
```diff
@@ -447,49 +471,71 @@ def bounded_u256_strategy(min_value: int = 0, max_value: int = 2**256 - 1):
...
+                        if address in _storage_tries.keys()
+                        else account_strategy
+                    )
+                    for address in addresses
+                }
+            ).map(lambda x: defaultdict(lambda: None, x)),
+        )
+    )
+
+    # Update the accounts in the state_trie with the correct storage_root
```
Eikix, line 514: you dont need this anymore

cairo/tests/utils/strategies.py
enitrat: minus

Eikix (approved):
lgtm

PR #1097, opened 2025-03-25
feat: bnp2 add

### LINKED ISSUE #1063: add
# What 
Implement addition for BNP2 (necessary for mul_by)

# How 
https://github.com/ethereum/execution-specs/blob/9c58cc8553ec3a59e732e81d5044c35aa480fbbb/src/ethereum/crypto/elliptic_curve.py#L129

obatirou:
Closes #1063 
Note: https://github.com/kkrt-labs/keth/pull/1098 was merged in the current branch instead of main

### DIFF
--- a/cairo/ethereum/crypto/alt_bn128.cairo
+++ b/cairo/ethereum/crypto/alt_bn128.cairo
@@ -270,6 +270,75 @@ func bnp2_double{range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr
     return result;
 }
 
+// Implementation of scalar multiplication for BNP2
+// Uses the double-and-add algorithm
+func bnp2_mul_by{
+    range_check_ptr, range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: ModBuiltin*
+}(p: BNP2, n: U384) -> BNP2 {
+    alloc_locals;
+    let n_is_zero = U384_is_zero(n);
+    if (n_is_zero != 0) {
+        let res = bnp2_point_at_infinity();
+        return res;
+    }
+
+    let bnf2_zero = BNF2_ZERO();
+    let x_is_zero = BNF2__eq__(p.value.x, bnf2_zero);
+    let y_is_zero = BNF2__eq__(p.value.y, bnf2_zero);
+    let p_is_infinity = x_is_zero * y_is_zero;
+    if (p_is_infinity != 0) {
+        return p;
+    }
+
+    // Extract the bits of n
+    let (bits_ptr, bits_len) = get_u384_bits_little(n);
+
+    // Initialize result as the point at infinity
+    let result = bnp2_point_at_infinity();
+
+    // Implement the double-and-add algorithm
+    return bnp2_mul_by_bits(p, bits_ptr, bits_len, 0, result);
+}
+
+func bnp2_mul_by_bits{
+    range_check_ptr, range_check96_ptr: felt*, add_mod_ptr: ModBuiltin*, mul_mod_ptr: ModBuiltin*
+}(p: BNP2, bits_ptr: felt*, bits_len: felt, current_bit: felt, result: BNP2) -> BNP2 {
+    alloc_locals;
+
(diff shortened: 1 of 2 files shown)


cairo/ethereum/crypto/alt_bn128.cairo
```diff
@@ -171,6 +171,61 @@ func bnp2_point_at_infinity() -> BNP2 {
...
+
+    // Calculate x_r = λ^2 - p.x - q.x
+    let lambda_squared = bnf2_mul(lambda, lambda);
+    let x_sum = bnf2_add(p.value.x, q.value.x);
+    let x_r = bnf2_sub(lambda_squared, x_sum);
+
+    // Calculate y_r = λ(p.x - x_r) - p.y
+    let x_diff_r = bnf2_sub(p.value.x, x_r);
+    let lambda_times_x_diff = bnf2_mul(lambda, x_diff_r);
+    let y_r = bnf2_sub(lambda_times_x_diff, p.value.y);
```
enitrat, line 222: should this part not be an ec_ops circuit?

obatirou: functions calls are not supported in circuits, here calls to bnf2_sub, div and mul are necessary

cairo/ethereum/crypto/alt_bn128.cairo
```diff
@@ -171,6 +171,61 @@ func bnp2_point_at_infinity() -> BNP2 {
...
+        return p;
+    }
+
+    let x_equal = BNF2__eq__(p.value.x, q.value.x);
+    if (x_equal != 0) {
+        let y_equal = BNF2__eq__(p.value.y, q.value.y);
+        if (y_equal != 0) {
+            return bnp2_double(p);
+        }
+        let res = bnp2_point_at_infinity();
```
zmalatrax, line 200: Why point at infinity ?
Here we only know that `p.x == q.x` and `p.y != q.y`, not that `p.y == -q.y`

We need an additional check to return point at infinity
There is the function `uint384_is_neg_mod_p` which checks if two Uint384 values are opposite. (used in the [`ec_add` function](https://github.com/kkrt-labs/keth/blob/236155cc51c166225afa84a77ef86e730148bafc/python/cairo-ec/src/cairo_ec/ec_ops.cairo#L117))

cairo/tests/ethereum/crypto/test_alt_bn128.py
```diff
@@ -85,6 +85,11 @@ def test_bnp2_point_at_infinity(self, cairo_run):
         def test_bnp2_double(self, cairo_run, p: BNP2):
             assert cairo_run("bnp2_double", p) == p.double()
 
+        @given(p=..., q=...)
+        def test_bnp2_add(self, cairo_run, p: BNP2, q: BNP2):
```
zmalatrax, line 89: Add tests to test all the branches of bnp2_add (p or q is zero, p = -q, p = q and standard case)

zmalatrax (changes requested):

cairo/ethereum/crypto/alt_bn128.cairo
obatirou: Isn't it implicit? 
If x == x', and x !=0, there are possible 2 points on the curve (x,y) and (x,-y)
Unless I miss something?

cairo/ethereum/crypto/alt_bn128.cairo
enitrat: yes they would need to be inlined

cairo/ethereum/crypto/alt_bn128.cairo
obatirou: As seen IRL:
on EC, if x == x', there are 2 possible points: (x, y) or (x, -y). But (x, y) + (x, -y) = point at infinity

cairo/tests/ethereum/crypto/test_alt_bn128.py
obatirou: added p = -q only one not covered

zmalatrax (approved):

PR #1116, opened 2025-03-26
feat: felt252_to_bits

### LINKED ISSUE #953: dev: leverage non-determinism in `extract_limb_bits`
Move `extract_limb_bits` to `maths.cairo` in `cairo_core`. Rename it to `felt252_to_bits`. 

Using a hint, we can ask the prover to directly provide the bit decomposition of limb as an array of bits. The Cairo code then only needs to:

-Verify that the provided bit array is valid (i.e., contains only 0s and 1s).
-Reconstruct the original limb from the bit array and ensure it matches the input.

This approach eliminates the recursive calls and reduces the operation to a single pass for verification, significantly lowering the number of steps.

# How It Works

**Non-Deterministic Hint:** 
Inside the `%{%}` block, the prover computes the binary representation of limb and writes the bits directly into bits_ptr. It also sets len to the number of bits.
This is done efficiently in Python & Rust.

**Verification:**
The Cairo code:
- iterates over the provided bits, verifying each is either 0 or 1 using range checks.
- reconstructs the original limb by computing sum(bit[i] * 2^i) and checks that it equals the input limb.

The optimized version is still an O(n) verification steps (iteration on all bits), but these are simpler operations (multiplications and additions) and avoid recursion and heavy function calls. The heavy lifting is offloaded to the prover.

The logic is very similar to `felt252_to_bytes_be`

n-eq:
Closes #953


python/cairo-addons/src/cairo_addons/hints/maths.py
```diff
@@ -37,3 +37,19 @@ def is_positive_hint(ids: VmConsts):
...
+
+    # Convert the limb to binary representation and remove '0b' prefix
+    binary = bin(limb)[2:]
+
+    # Write each bit to memory
+    for i, bit in enumerate(binary):
+        segments.memory[bits_ptr + current_len + i] = int(bit)
+
+    # Update the current_len with the new length
+    ids.current_len = current_len + len(binary)
```
enitrat, line 56: - rename limb -> value (agnostic - it's just a felt value)
- current_len should not be required
- inlined intermediate bits_ptr variable
- used `load_data` to load a sequence of data at a given pointer
```suggestion
    # Convert the limb to binary representation and remove '0b' prefix
    # and write it to memory
    binary = bin(ids.value)[2:]
    segments.load_data(ids.bits_ptr, binary)
```

enitrat: note: update my suggestion sinppet to shift the value until it has the desired bit length (see similar byte function).

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,57 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
 
     return bit_length;
 }
+
+func felt252_to_bits{range_check_ptr}(limb: felt, bits_ptr: felt*, current_len: felt) -> felt {
```
enitrat, line 540:
```suggestion
func felt252_to_bits{range_check_ptr}(value: felt, len: felt, dst: felt*) -> felt {
```

to keep consistency with `felt252_to_bytes` names - update internal usage of values

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,57 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
...
+    alloc_locals;
+    if (limb == 0) {
+        return current_len;
+    }
+    with_attr error_message("felt252_to_bits: limb must be < 252") {
```
enitrat, line 545:
```suggestion
    with_attr error_message("felt252_to_bits: len must be < 252") {
```

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,57 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
...
+    jmp end if is_done != 0;
+
+    // Get the current bit using modulo 2
+    let (_, bit) = unsigned_div_rem(limb, 2);
+    assert bits_ptr[current_len] = bit;
+
+    // all the generated felt elements should be either 0 or 1
+    let bit_minus_one = bit - 1;
+    let product = bit * bit_minus_one;
+    let is_valid_bit = is_zero(product);
```
enitrat, line 574: // # 1. Validate that all intermediate bits are 0 or 1
this can be simply translated into `assert bit * bit = bit`

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,57 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
...
+    let range_check_ptr = [ap - 3];
+    let current_len = [ap - 2];
+    let limb = [ap - 1];
+
+    let is_done = is_zero(limb);
+    jmp end if is_done != 0;
+
+    // Get the current bit using modulo 2
+    let (_, bit) = unsigned_div_rem(limb, 2);
+    assert bits_ptr[current_len] = bit;
```
enitrat, line 569: this is the right verification, however, if we do a division of `value/2` to get the current bit, this defeats the purpose of this function - as we want to rely on multiplications only, that are much cheaper than divisions.

This is why we have an accumulator and verify that sum(bit[i] << i) = original_value

enitrat:
good for the function itself, now, we need to
1. Write the hint in Rust
2. Integrate it properly

python/cairo-core/tests/src/cairo_core/test_maths.py
```diff
@@ -182,3 +182,12 @@ def test_felt252_to_bytes_be_should_panic_on_wrong_output_noncanonical(
     def test_felt252_bit_length(self, cairo_run, value):
         res = cairo_run("felt252_bit_length", value=value)
         assert res == value.bit_length()
+
+    @given(
+        value=st.integers(min_value=150, max_value=200),
+        len=st.integers(min_value=10, max_value=11),
```
n-eq, line 188:
```suggestion
        len=st.integers(min_value=10, max_value=50),
```

cairo/ethereum/utils/numeric.cairo
```diff
@@ -493,66 +499,43 @@ func U384_to_be_bytes{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
...
+    }
+    let d2_len = felt252_to_bits_rev(num.value.d2, 96, bits_ptr + total_bits_len);
+    tempvar total_bits_len = total_bits_len + d2_len;
 
-    // Continue with the remaining bits
-    return extract_limb_bits(q, bits_ptr, current_len + 1);
+    // Process limb3 (d3)
+    let bits_len = felt252_bit_length(num.value.d3);
+    let d3_len = felt252_to_bits_rev(num.value.d3, bits_len, bits_ptr + total_bits_len);
+    return (bits_ptr, total_bits_len + d3_len);
```
enitrat, line 540: missing increment of total_bits_len here, no?

crates/cairo-addons/src/vm/hint_definitions/maths.rs
```diff
@@ -171,3 +172,39 @@ fn felt252_bit_and(num_x: Felt252, num_y: Felt252) -> Result<Felt252, HintError>
...
+                    let bit = if value_biguint.bit(i as u64) { BigUint::from(1u32) } else { BigUint::from(0u32) };
+                    Felt252::from(bit).into()
+                })
+                .collect();
+            bits.extend(vec![Felt252::from(0u32).into(); len - bits.len()]);
+
+            // Write the bits to memory
+            vm.segments.load_data(dst_ptr, &bits)?;
+            Ok(())
+
```
n-eq, line 207:
```suggestion
```

python/cairo-core/tests/src/cairo_core/test_maths.py
```diff
@@ -182,3 +182,19 @@ def test_felt252_to_bytes_be_should_panic_on_wrong_output_noncanonical(
...
         res = cairo_run("felt252_bit_length", value=value)
         assert res == value.bit_length()
+
+    @given(
+        value=st.integers(min_value=2, max_value=2**248 - 1),
+        len=st.integers(min_value=0, max_value=50),
+    )
+    def test_felt252_to_bits_rev(self, cairo_run, value, len):
+        # when len is 0, we expect an empty list, otherwise if len is too small,
+        # we need to increase it to the bit length of the value + 1
```
enitrat, line 192: why so? what if I want to get, let's say, the 20 first bits of a number `n`?

n-eq: I just reread the issue description and I'm not sure we should allow a `len` argument (since the last steps verifies the decomposition corresponds to the initial value)

enitrat: yes, we can use a system of mask

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,61 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
 
     return bit_length;
 }
+
+// @notice Converts a felt252 to a bit array, little-endian, and outputs to `dst`.
+// @dev Can only convert up to 251 bits included.
+// @returns the length of the bit array.
+func felt252_to_bits_rev{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
```
Eikix, line 543: why do we call it rev instead of felt252_to_le_bits?

enitrat: endianness is the order in which bytes within a word of digital data are stored, thus i don't think it applies to bits

.github/workflows/python_tests.yml
```diff
@@ -40,7 +40,10 @@ jobs:
           uv run pytest -n ${{ inputs.parallelism }} --durations=0 -v -s --log-cli-level=DEBUG --no-skip-cached-tests ${{ inputs.pytest-add-params }}
 
       - name: Upload coverage
-        if: inputs.coverage-flags != ''
+        if:
```
enitrat, line 43: what is this for?

n-eq: nothing related, I just migrated to another Mac and committed everything that was wip

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -536,3 +536,69 @@ func felt252_bit_length{range_check_ptr}(value: felt) -> felt {
...
+    // Check if the bit is valid (0 or 1)
+    let bit = output[current_len];
+    with_attr error_message("felt252_to_bits_rev: bits must be 0 or 1") {
+        assert bit * bit = bit;
+    }
+
+    let pow = pow2(current_len);
+    tempvar shifted_bit = bit * pow;
+    tempvar current_len = current_len + 1;
+    tempvar acc = acc + shifted_bit;
```
feltroidprime, line 587: can avoid access to pow = pow2(len) by multiplying by 2 recursively instead : 

word = b0 + b1 *2 + b2 *2 ^2 + b3 * b^3
word = b0 + 2 * ( b1 + 2 * (b2 + 2 * ( b3)))

enitrat:
A few comments, recap, and what's left to do to get this in `done` state, with an decrease of 50% of steps used
image
(from ~1e4 to 5e3)

1.  Have the hint (`felt252_to_bits_rev`) calculate the actual `bits_used` for the value (masked by `len`).
2.  The hint writes *only* the first `bits_used` little-endian bits to the `dst` pointer.
3.  The hint makes `bits_used` available to the Cairo code.
4.  The Cairo code uses `memset` to zero out the remaining memory from `dst + bits_used` up to `len`.
5.  The Cairo code verifies the result by looping only up to `bits_used`, reconstructing the value using little-endian logic, and comparing it to the original `value` masked by `len`.

**Specific File Feedback:**

1.  **`python/cairo-core/src/cairo_core/maths.cairo` (`felt252_to_bits_rev`)**
```
    // Hint populates `bits_used` with the number of bits used to represent the value
    // and writes the bits to `dst` in reversed order.
    let output = &dst[0];
    local bits_used: felt;
    %{ felt252_to_bits_rev %}

    // We know it takes `bits_used` bits to represent the value
    // the rest is filled with zeroes.
    memset(output + bits_used, 0, len - bits_used);

    tempvar pow_i = 1;
    tempvar current_len = 0;
    tempvar acc = 0;
 ```
Here with `memset` we're ensured only zeroes are written between [bits_used, len]. Thus if we compute `sum(bit_i * 2^i) for i in range(0, bits_used)` we should get our original value.
Also, instead of using `pow2` and make a function call per iteration, we can simply multiply by 2 the offset on each loop, which should be considerably cheaper.


```
    loop:
    let pow_i = [ap - 3];
    let current_len = [ap - 2];
    let acc = [ap - 1];
    let len = [fp - 4];

    // loop stop condition: current_len == bits_used (we iterated `bits_used` times)
    // because then sum{i=0..bits_used-1} 2^i * bit[i] = original value
    let is_done = is_zero(current_len - bits_used);
    jmp end if is_done != 0;
```

The loop condition changes, as we know we've verified everything once current_len == bits_used, the rest are of the bits are zeroed out.

For the jump preparation:
```
    // Accumulate value: acc = acc + bit * (2^i)
    tempvar shifted_bit = bit * pow_i;
    tempvar next_pow_i = pow_i * 2;
    tempvar current_len = current_len + 1;
    tempvar acc = acc + shifted_bit;

    static_assert next_pow_i == [ap - 3];
    static_assert current_len == [ap - 2];
    static_assert acc == [ap - 1];
    jmp loop;
```


The pythoon hint would look like this:

```
@register_hint
def felt252_to_bits_rev(ids: VmConsts, segments: MemorySegmentManager):
    """
    Hint to write the `len` least significant bits of `value`
    to the memory segment starting at `dst` in reversed order.
    """
    value = ids.value
    length = ids.len
    dst_ptr = ids.dst

    if length != 0:
        # Ensure we only work with the bits relevant to the requested length
        # Python's integers handle large numbers automatically
        mask = (1 << length) - 1
        value_masked = value & mask
        bits_used = value_masked.bit_length()

        # Generate the 'length' bits in reversed order
        bits = [int(bit) for bit in bin(value_masked)[2:].zfill(length)[::-1]]

        # Load the generated bits into the specified memory segment
        ids.bits_used = min(bits_used, length)
        segments.load_data(dst_ptr, bits)
    else:
        ids.bits_used = 0
```

and  for rust:

```rs
pub fn felt252_to_bits_rev() -> Hint {
    Hint::new(
        String::from("felt252_to_bits_rev"),
        |vm: &mut VirtualMachine,
         _exec_scopes: &mut ExecutionScopes,
         ids_data: &HashMap<String, HintReference>,
         ap_tracking: &ApTracking,
         _constants: &HashMap<String, Felt252>|
         -> Result<(), HintError> {
            // Get input values from Cairo
            let value = get_integer_from_var_name("value", vm, ids_data, ap_tracking)?;
            let len = get_integer_from_var_name("len", vm, ids_data, ap_tracking)?;
            let dst_ptr = get_ptr_from_var_name("dst", vm, ids_data, ap_tracking)?;

            let len_usize: usize =
                len.try_into().map_err(|_| MathError::Felt252ToUsizeConversion(Box::new(len)))?;

            // Handle length == 0 case separately
            if len_usize == 0 {
                insert_value_from_var_name("bits_used", Felt252::ZERO, vm, ids_data, ap_tracking)?;
                // No bits to write for len 0, dst_ptr remains untouched
                return Ok(());
            }

            let value_biguint = value.to_biguint();

            // Ensure we only work with the bits relevant to the requested length
            // Create mask: (1 << length) - 1
            let mask = (BigUint::from(1u64) << len_usize) - BigUint::from(1u64);
            let value_masked = value_biguint & mask;

            // Calculate bits_used based on the masked value's actual bit length
            let bits_used = value_masked.bits() as usize;
            let bits_used_to_assign = std::cmp::min(bits_used, len_usize);
            insert_value_from_var_name(
                "bits_used",
                Felt252::from(bits_used_to_assign),
                vm,
                ids_data,
                ap_tracking,
            )?;

            // Generate the 'bits_used' bits in reversed order (LSB first in the vec)
            let bits: Vec<MaybeRelocatable> = (0..bits_used)
                .map(|i| {
                    // Check the i-th bit of the masked value
                    Felt252::from(value_masked.bit(i as u64)).into()
                })
                .collect();

            // Write the bits to memory starting at dst_ptr
            vm.segments.load_data(dst_ptr, &bits)?;
            Ok(())
        },
    )
}
```

enitrat (dismissed):

enitrat:
test fails probably because of a rebase conflict with main (a new function was introduced that doesnt take BitwiseBuiltin as arg)

enitrat (dismissed):

PR #1120, opened 2025-03-26
feat: startswith

### LINKED ISSUE #1119: feat: Bytes__startswith__
Bytes__startswith__

Eikix:
closes #1119

### DIFF
--- a/cairo/ethereum/utils/bytes.cairo
+++ b/cairo/ethereum/utils/bytes.cairo
@@ -13,6 +13,7 @@ from ethereum_types.bytes import (
 from ethereum_types.numeric import bool
 from starkware.cairo.common.alloc import alloc
 from starkware.cairo.common.math import assert_not_equal, split_int
+from starkware.cairo.common.math_cmp import is_le
 from starkware.cairo.common.memcpy import memcpy
 from starkware.cairo.common.memset import memset
 from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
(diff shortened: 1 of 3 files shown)


cairo/ethereum/utils/bytes.cairo
```diff
@@ -71,6 +72,18 @@ func Bytes__eq__(_self: Bytes, other: Bytes) -> bool {
     return res;
 }
 
+func Bytes__startswith__{range_check_ptr}(self: Bytes, prefix: Bytes) -> bool {
+    alloc_locals;
+    let is_prefix = is_le(prefix.value.len, self.value.len);
+    if (is_prefix == 0) {
+        tempvar res = bool(0);
+        return res;
+    }
```
enitrat, line 81:
```suggestion
    let len_smaller = is_le(prefix.value.len, self.value.len);
    if (len_smaller == 0) {
        tempvar res = bool(0);
        return res;
    }
```

enitrat: (the len not being smaller doesn't mean its a prefix)

cairo/ethereum/utils/bytes.cairo
```diff
@@ -71,6 +72,18 @@ func Bytes__eq__(_self: Bytes, other: Bytes) -> bool {
...
+    tempvar self = Bytes(new BytesStruct(data=self.value.data, len=prefix.value.len));
```
enitrat, line 82:
```suggestion
    # Check whether self.value.data[0:len(prefix)] == prefix
    tempvar self_slice = Bytes(new BytesStruct(data=self.value.data, len=prefix.value.len));
```

PR #1137, opened 2025-03-27
feat: storage diff func

### LINKED ISSUE #1130: feat:  storage diff computation in `compute_diff` function
complete #1124

Eikix:
closes #1130


python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -31,6 +32,8 @@
     "c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470"
 )
 
+0x56E81F171BCC55A6FF8345E692C0F86E5B48E01B996CADC001622FB5E363B421
```
enitrat, line 35: coucou

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -358,7 +361,7 @@ def from_data(cls, data: Dict[str, Any]) -> "EthereumTrieTransitionDB":
             raise ValueError(f"Post state root not found in nodes: {post_state_root}")
 
         instance = cls(
-            nodes={**pre_trie.nodes, **post_nodes},
+            nodes=defaultdict(lambda: None, {**pre_trie.nodes, **post_nodes}),
```
enitrat, line 364: no

python/mpt/src/mpt/ethereum_tries.py
Eikix: turns out you can just type stuff inside a codebase?

PR #1142, opened 2025-03-28
fix: add checks on len in Bytes_to_Bytes32

enitrat:


cairo/ethereum/utils/bytes.cairo
```diff
@@ -26,6 +26,7 @@ from cairo_core.maths import unsigned_div_rem, felt252_to_bytes_le, felt252_to_b
 from cairo_core.comparison import is_zero
 from ethereum.utils.numeric import min
 from legacy.utils.bytes import bytes_to_felt_le
+from cairo_core.control_flow import raise
```
enitrat, line 29:
```suggestion
```

PR #1143, opened 2025-03-28
feat: support serialization of enitre segments

enitrat:
Supports serialization of entire segments in Serde. For example, you can now return a `MyStruct*`, pointer to the start of a segment of size `N` and it will be serialized as `[N[0], N[1]...]`.

This also fixes a small issue when using `%{serialize(ids.variable)%}` inside a hint, as the we couldn't properly resolve the identifiers because we didn't know which cairo file we were running.


enitrat:
closing this, it would be nice to have, but let's keep our strong types for now.

PR #1159, opened 2025-03-31
feat: check branch node

### LINKED ISSUE #1157: feat: check branch node length in compute_diff
check branch node length in compute_diff is at least 2 
i.e. at minimum 2 subnodes are non null

Eikix:
closes #1157


python/mpt/src/mpt/utils.cairo
```diff
@@ -123,3 +124,50 @@ func deserialize_to_internal_node{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}
...
+// @dev Raises an error if the branch node is invalid.
+// @param node The branch node to check.
+func check_branch_node(node: BranchNode) {
+    alloc_locals;
+
+    local first_non_null_branch;
+    local seconds_non_null_branch;
+    let subnodes_ptr = cast(node.value.subnodes.value, felt*);
+    %{
+        node = serialize(ids.node);
```
enitrat, line 139:
```suggestion
```

no serialization plz. use my proposal unless it cannot be used.

python/mpt/src/mpt/utils.cairo
```diff
@@ -123,3 +124,50 @@ func deserialize_to_internal_node{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}
...
+        non_null_branches = [idx for idx, subnode in enumerate(node.subnodes) if (subnode is not None and subnode != b"")]
```
Eikix, line 140: here since we check difference wrt `None` and empty bytes, writing directly memory accesses is a bit ugly cc @enitrat do you agree?

python/mpt/src/mpt/utils.cairo
```diff
@@ -123,3 +124,50 @@ func deserialize_to_internal_node{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}
...
+    }
+
+    let (buffer) = alloc();
+    tempvar empty_bytes = Bytes(new BytesStruct(data=buffer, len=0));
+
+    let invalid = Bytes__eq__(x.value.bytes, empty_bytes);
+
+    if (invalid.value != 0) {
+        raise('ValueError');
+    }
```
enitrat, line 157: use `assert_non_zero` on the pointer value. once asserted it's not zero, use `assert_non_zero` on the length of the bytes (as per the method described in the issue).

python/mpt/tests/src/mpt/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+    @given(branch_node=...)
+    def test_check_branch_node(self, cairo_run_py, branch_node: BranchNode):
+        try:
+            cairo_run_py("check_branch_node", branch_node)
+        except ValueError as cairo_error:
+            with strict_raises(type(cairo_error)):
+                check_branch_node(branch_node)
+            return
+
+        check_branch_node(branch_node)
```
enitrat, line 23: can run with RustVM if you get rid of serialize

python/mpt/src/mpt/utils.cairo
enitrat: i prefer the old way. no use of `serialize`

python/mpt/src/mpt/utils.cairo
Eikix: assert_not_zero from sw does not work on pointers due to their call to `assert_integer`

```
// Verifies that value != 0. The proof will fail otherwise.
func assert_not_zero(value) {
    %{
        from starkware.cairo.common.math_utils import assert_integer
        assert_integer(ids.value)
        assert ids.value % PRIME != 0, f'assert_not_zero failed: {ids.value} = 0.'
    %}
    if (value == 0) {
        // If value == 0, add an unsatisfiable requirement.
        value = 1;
    }

    return ();
}
```

enitrat (commented):
👍 lmk if you need help on the rust-hint side.

python/mpt/tests/src/mpt/test_utils.py
```diff
@@ -0,0 +1,23 @@
...
+from mpt.utils import check_branch_node, nibble_list_to_bytes
+
+
+class TestUtils:
+    @given(bytes=...)
+    def test_nibble_list_to_bytes(self, bytes: Bytes):
+        assert nibble_list_to_bytes(bytes_to_nibble_list(bytes)) == bytes
+
+    @given(branch_node=...)
+    def test_check_branch_node(self, cairo_run, branch_node: BranchNode):
```
enitrat, line 15: are you generating invalid branches with the strategy?

crates/cairo-addons/src/vm/hint_definitions/mpt.rs
```diff
@@ -0,0 +1,122 @@
...
+                Ok(())
+            } else {
+                Err(HintError::CustomHint(Box::from(format!(
+                    "Could not find two non-null subnodes. Found only {}",
+                    non_null_indices.len()
+                ))))
+            }
+        },
+    )
+}
```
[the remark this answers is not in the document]
enitrat: best to avoid lots of levels of nesting

python/mpt/tests/src/mpt/test_utils.py
Eikix: not yet

enitrat (commented):
it seems that codecov says the raise cases are not covered

python/mpt/src/mpt/utils.cairo
```diff
@@ -123,3 +124,54 @@ func deserialize_to_internal_node{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}
...
+    // Check that the first subnode is not None and not empty
+    tempvar x = Extended(cast(subnodes_ptr[first_non_null_index], ExtendedEnum*));
+
+    if (cast(x.value, felt) == 0) {
+        raise('ValueError');
+    }
+    // Case 1: subnode is a digest
+    if (cast(x.value.bytes.value, felt) != 0) {
+        if (x.value.bytes.value.len == 0) {
+            raise('ValueError');
```
enitrat, line 150: looks like only this case is covered by the test

Eikix: i think it's just the number of examples

Eikix: ah no, it's that we can't serialize empty arrays in this context

Eikix: example:

```python
def invalid_branch_node_strategy():
    return st.builds(
        BranchNode,
        subnodes=st.one_of(
            st.integers(min_value=0, max_value=2**256 - 1).map(
                lambda x: x.to_bytes(32, "little")
            ),
            st.from_type(LeafNode).map(lambda x: encode_internal_node(x)),
            st.from_type(ExtensionNode).map(lambda x: encode_internal_node(x)),
            st.none(),
            st.just([]),
        ).map(lambda subnode: tuple([b""] * 13 + [] + [] + [subnode])),
        value=st.just(b""),
    )
```

Just results in a tuple of subnodes of len < 16 which won't cause us to compare an item to an empty list.

enitrat: ok yeah i get it 👍

enitrat (approved):

python/mpt/src/mpt/utils.cairo
Eikix: Trying with [[]] instead of []

enitrat: now this one is no longer covered 😢

enitrat: `map(lambda subnode: tuple([b""] + [[]] + [b""] * 13)),`

I think this is what you want to have as strat to cover things properly? simply two null nodes, each one of different type? 😅

python/mpt/tests/src/mpt/test_utils.py
```diff
@@ -0,0 +1,65 @@
...
+            st.just(list()),
+        ).map(
+            lambda subnode: (
+                list([b""] * 13 + [[]] + [[]] + [subnode])
+                if isinstance(subnode, bytes)
+                else list([[]] * 13 + [b""] + [b""] + [subnode])
+            )
+        ),
+        value=st.just(b""),
+    )
```
enitrat, line 35:
```suggestion
def invalid_branch_node_strategy():
    return st.builds(
        BranchNode,
        subnodes=st.lists(
            st.one_of(st.just([b""]), st.just([[]])),
            min_size=16,
            max_size=16,
            unique=False
        ),
        value=st.just(b""),
    )
```

Is this not what we want eventually?

Eikix: but if all of the subnodes are zero length, then the assertions will fail when we assert x len != 0?

python/mpt/tests/src/mpt/test_utils.py
```diff
@@ -0,0 +1,50 @@
...
+            cairo_run("check_branch_node", branch_node)
+        except ValueError as cairo_error:
+            with strict_raises(type(cairo_error)):
+                check_branch_node(branch_node)
+            return
+
+        check_branch_node(branch_node)
+
+    @given(invalid_branch_node=invalid_branch_node_strategy())
+    def test_check_branch_node_invalid(
```
enitrat, line 40: also this is a bad pattern! you have a reference function in python, you should be generating an input that can be either correct or incorrect, and then have a single test function!

enitrat:
@Eikix i pushed the required changes to merge this PR, should be good now

enitrat (approved):

PR #1166, opened 2025-03-31
dev: finish felt252_to_bits

enitrat:


python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -2,13 +2,14 @@ from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
...
 from starkware.cairo.common.uint256 import Uint256
+from starkware.cairo.common.memset import memset
 from cairo_core.comparison import is_zero
 
 // @dev Inlined version of unsigned_div_rem
 // Returns q and r such that:
 //  0 <= q < rc_bound, 0 <= r < div and value = q * div + r.
 //
-// Assumption: 0 < div <= PRIME / rc_bound.
+// Assumption: 0 < div <= ME / rc_bound.
```
enitrat, line 12:
```suggestion
// Assumption: 0 < div <= PRIME / rc_bound.
```

python/cairo-core/src/cairo_core/maths.cairo
```diff
@@ -554,50 +555,57 @@ func felt252_to_bits_rev{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
         let range_check_ptr = range_check_ptr + 2;
     }
 
+    // Hint populates `bits_used` with the number of bits used to represent the value
+    // and writes the bits to `dst` in reversed order.
     let output = &dst[0];
-    // last non-zero bit
-    tempvar last_one: felt;
+    local bits_used: felt;
```
n-eq, line 561: why `local` instead of `tempvar`?

enitrat: it's a constant that we want to access at each iteration of the loop, without having to update it before jumping

PR #1196, opened 2025-04-01
try/catch for missing file in RLP ef tests

ClementWalter:


n-eq (approved):

ClementWalter:
@enitrat it's just if you never run pytest and the first time you do it, you only do it with a filter `-k`, the git pull fixture may not be triggered and so pytest raises with missing file

PR #1202, opened 2025-04-02
feat: support accessing scope constants in rust hints

### LINKED ISSUE #1191: feat: make constant values accessible when running python hints in RustVM
`constants` are not part of `ids_data` and thus are not accessible through the `ids` object in the Rust VM. Implement this.

It should be simple, the only problem is that the `constants` variable passed to the hint are not local to the hint execution (visible in the same scope) but are the ones of the entire program. We must ensure that we only put accessible the constants in scope.

This will require some changes upstream in the cairo-vm repo. The current strat envisioned is to
- modify `serialize/deserialize` of the program to load the `destination` of an alias, so that we can resolve imported constants
- make the hint accessible_scopes accessible in the HintData

enitrat:
Closes #1191

Uses the newly added hint data in the cairo vm to resolve the imported constants ("alias") and the local constant, making them available in the `ids` object.

### DIFF
--- a/cairo/tests/test_hint_runner.cairo
+++ b/cairo/tests/test_hint_runner.cairo
@@ -1,7 +1,6 @@
 from ethereum_types.numeric import U256, U256Struct
 from starkware.cairo.common.alloc import alloc
 
-from cairo_ec.curve.alt_bn128 import alt_bn128
 from ethereum.exceptions import ValueError
 
 func test__ap_accessible() {
--- a/cairo/tests/test_hint_runner.cairo
+++ b/cairo/tests/test_hint_runner.cairo
@@ -1,6 +1,9 @@
 from ethereum_types.numeric import U256, U256Struct
 from starkware.cairo.common.alloc import alloc
 
+from cairo_ec.curve.alt_bn128 import alt_bn128
+from ethereum.exceptions import ValueError
+
 func test__ap_accessible() {
     tempvar x = 100;
     %{ assert memory[ap-1] == 100 %}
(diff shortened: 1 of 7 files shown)


cairo/tests/test_hint_runner.cairo
```diff
@@ -1,6 +1,9 @@
 from ethereum_types.numeric import U256, U256Struct
 from starkware.cairo.common.alloc import alloc
 
+from cairo_ec.curve.alt_bn128 import alt_bn128
```
enitrat, line 4:
```suggestion
```

crates/cairo-addons/src/vm/vm_consts.rs
```diff
@@ -674,11 +675,48 @@ pub fn create_vm_consts_dict(
...
+        if hint_accessible_scopes.iter().any(|scope| module_path == *scope) {
+            py_ids_dict
+                .borrow_mut(py)
+                .items
+                .insert(const_name.to_string(), value.to_biguint().into_bound_py_any(py)?.into());
+            continue;
+        }
+
+        // Check if constant is accessible through an alias in any accessible scope
+        for scope in hint_accessible_scopes {
```
Eikix, line 703: is this case covered by the tests in this PR? it looks like only the directly accessible case is tested

ClementWalter (approved):
🙈

crates/cairo-addons/src/vm/vm_consts.rs
enitrat: It is tested

PR #1212, opened 2025-04-02
Put same known first party everywhere and update file tree

ClementWalter:
image
image


zmalatrax:
There is `import os` in `cairo_ec/tests/curve/test_alt_bn128.py` though ([link](https://github.com/kkrt-labs/keth/blob/b21b793a13508f9dd147aa7c72dac479093036c6/python/cairo-ec/tests/curve/test_alt_bn128.py#L1-L6))

enitrat (approved):

zmalatrax (approved):

Issue #1127, opened 2025-03-27
feat: add `bnf12_multiplicative_inverse` Rust hint


zmalatrax:
# What

If needed, and with more time available, we should solve this and implement the Rust hint for `bnf12_multiplicative_inverse`.

# Context

The current implementation of `bnf12_div` #999  relies on the `bnf12_multiplicative_inverse` hint, which is similar to the one used in `bnf2_div` #1065 

However, due to disrepancies between the two inverses computed with EELS and arkworks, without identifying/resolving the root cause in reasonable times we chose to continue development solely with the Python hint (EELS).

Potential root cause is the coefficients ordering to instantiate objects of the `Fq12` tower construction, different from the EELS one. One example toward this is the `halo2-lib` which doesn't encode the 12 coefficients "sequentially" (see [this](https://github.com/axiom-crypto/halo2-lib/blob/dd04ade06e205d63ea48aee616608ad62c5a7641/halo2-ecc/src/fields/fp12.rs#L225-L241))

Details on a test case are available [here](https://github.com/kkrt-labs/keth/pull/1103#issuecomment-2754986113)

A [comment by Paul Miller](https://github.com/paulmillr/noble-curves/blob/fa50a114a8241982b4e8c2fa40e3bcc77c3766ce/src/bn254.ts#L9-L30) on the state of alt_bn128 pairing libraries (7 months old):

> There are huge compatibility issues in the ecosystem:
>
> 1. Different libraries call it in different ways: "bn254", "bn256", "alt_bn128", "bn128".
> 2. libff has bn128, but it's a different curve with different G2:
   https://github.com/scipr-lab/libff/blob/a44f482e18b8ac04d034c193bd9d7df7817ad73f/libff/algebra/curves/bn128/bn128_init.cpp#L166-L169
> 3. halo2curves bn256 is also incompatible and returns different outputs
>
> The goal of our implementation is to support "Ethereum" variant of the curve, because it at least has specs:
>
> - EIP196 (https://eips.ethereum.org/EIPS/eip-196) describes bn254 ECADD and ECMUL opcodes for EVM
> - EIP197 (https://eips.ethereum.org/EIPS/eip-197) describes bn254 pairings
> - It's hard: EIPs don't have proper tests. EIP-197 returns boolean output instead of Fp12
> - The existing implementations are bad. Some are deprecated:
>     - https://github.com/paritytech/bn (old version)
[...]

PR #1221, opened 2025-04-02
feat: add rustvm hint for bn12-mul

### LINKED ISSUE #1220: feat: add `bnf12_multiplicative_inverse` rust hint
it's currently missing. we ideally want these hints to run on the rust vm.

I see that currently, if the python multiplicative_inverse uses an input too big, it will fail with an Overflow error. I think we should filter out inputs too big, so as not to overflow on the rust vm for "nothing"?

enitrat:
@zmalatrax this fails because even if, in rust, `b*b_inv` = 1, the verification fails in cairo. know why?

Closes #1220.

### DIFF
--- a/cairo/ethereum/crypto/alt_bn128.cairo
+++ b/cairo/ethereum/crypto/alt_bn128.cairo
@@ -1138,10 +1138,14 @@ func bnf12_div{
     local b_inv: BNF12;
 
     %{ bnf12_multiplicative_inverse %}
+    %{ breakpoint() %}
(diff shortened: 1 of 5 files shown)


obatirou:
It is related to the discussion on slack https://kkrtlabs.slack.com/archives/C07K44XHW0J/p1743073532149839 and resumed in the issue here https://github.com/kkrt-labs/keth/issues/1127#issue-2952984578

enitrat:
ah great, i missed that one.

PR #1224, opened 2025-04-03
feat: use zkpig newest inputs in prove_block and test_fork

### LINKED ISSUE #1009: feat: load state from state tries in `prove_block`
## WHY

we need to prove the final state root for a given block. For that we will load the initial State using the MPT Libraries.

## WHAT

Load the prover inputs into the cairo memory as a State object

## HOW

- Modify the `prove_block` script to call `StateTries.from_json`, passing the ZK-PI generated into `data/1/inputs` generated with `zkpig generate <BLOCK_NUMBER>`.
- Call `.to_state()`on this object to get the EELS state object
- Change the `state=convert_defaultdict(load.json_to_state(fixture["pre"])),` to use this new state.

Eikix:
replace eels json by zkpi inputs in prove_block and test_fork

closes #1009


README.md
```diff
@@ -143,7 +136,7 @@ Requirements:
 - ZKPI data must be available as a JSON file
 - Compiled Cairo program must exist at the specified path (you can run
   `cairo-compile --proof_mode cairo/ethereum/cancun/main.cairo --cairo_path=cairo --no_debug_info --output build/main_compiled.json`
-  for that)
+  for that or alternatively `uv run compile_keth`
```
enitrat, line 139: just `uv run compile_keth`

cairo/scripts/prove_block.py
```diff
@@ -114,6 +122,75 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
+def load_pre_state(data: Dict[str, Any]) -> State:
+    """Load a trie fixture from a JSON file."""
+    try:
+        fixture = EthereumTrieTransitionDB.from_data(data)
+        state = fixture.to_pre_state()
+    except Exception:
```
enitrat, line 130: comment some context pls

cairo/scripts/prove_block.py
```diff
@@ -174,16 +241,17 @@ def load_zkpi_fixture(zkpi_path: Path) -> Dict[str, Any]:
...
     # Create blockchain
     chain = BlockChain(
         blocks=blocks,
-        state=prepare_state(load.json_to_state(fixture["pre"])),
-        chain_id=U64(fixture["chainId"]),
+        state=prepare_state(load_pre_state(prover_inputs)),
+        chain_id=U64(prover_inputs["chainConfig"]["chainId"]),
     )
 
+    logger.info("Temporary fix: Computing partial state root")
```
enitrat, line 254: are these logger useful?

enitrat (commented):
can we delete zkpi_to_eels ?

cairo/tests/ethereum/cancun/test_fork.py
```diff
@@ -821,8 +787,10 @@ def test_apply_body(self, cairo_run, data, withdrawals: Tuple[Withdrawal, ...]):
 
     @pytest.mark.parametrize(
         "zkpi_path",
-        list(Path("data/1/eels").glob("*.json")),
-        ids=[x.stem for x in Path("data/1/eels").glob("*.json")],
+        [
+            Path("test_data/21688509.json"),
+            *[Path(p) for p in glob("data/inputs/1/*.json")],
```
Eikix, line 792: this makes the test very long, maybe not such a good idea

enitrat: for sure, we should select blocks in test_data only imo. data/** should not be commited

enitrat (dismissed):

PR #1247, opened 2025-04-04
feat: sort_storage_diff

Eikix:


python/mpt/src/mpt/utils.cairo
```diff
@@ -276,3 +277,10 @@ func sort_account_diff{range_check_ptr}(account_diff: AccountDiff) -> AccountDif
     let sorted_account_diff = AccountDiff(sorted_diff_struct_ptr);
     return sorted_account_diff;
 }
+
+func sort_storage_diff{range_check_ptr}(storage_diff: StorageDiff) -> StorageDiff {
+    let casted_storage_diff = AccountDiff(cast(storage_diff.value, AccountDiffStruct*));
+    let diff = sort_account_diff(casted_storage_diff);
+    let res = StorageDiff(cast(diff.value, StorageDiffStruct*));
+    return res;
+}
```
enitrat, line 286: maybe create a generic function `sort_diff`, a struct `GenericDiff`, cast `AccountDiff` to `GenericDiff` and cast `StorageDiff` to `GenericDiff`, so as to avoid assumptions on the types inside `sort_account_diff` ?

enitrat: yagni?

Eikix: "you are not going to need it"

enitrat (approved):

python/mpt/src/mpt/utils.cairo
Eikix: on top of the comment you suggested?

PR #1250, opened 2025-04-04
fix: correctly finalize state at end of block

### LINKED ISSUE #1243: bug: `state_root`, `storage_roots` do not rebind State with squashed segments
in `state_root` and `storage_roots`, we squash the main_trie and storage_tries, however, the `State` struct is not rebinded with these squashed segments. fix it.

could've been caught by https://github.com/kkrt-labs/keth/issues/535, which should be a priority next sprint

enitrat:
Closes #1243

### DIFF
--- a/cairo/ethereum/cancun/state.cairo
+++ b/cairo/ethereum/cancun/state.cairo
@@ -1080,6 +1080,8 @@ func empty_transient_storage{range_check_ptr}() -> TransientStorage {
     return transient_storage;
 }
 
+// @notice Computes the storage roots of all the addresses in the state
+// @dev The input state must've been squashed for unique keys.
 func storage_roots{
     range_check_ptr,
     bitwise_ptr: BitwiseBuiltin*,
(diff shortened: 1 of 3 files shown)


cairo/ethereum/cancun/state.cairo
```diff
@@ -1403,3 +1381,67 @@ func mapping_address_bytes32_write{range_check_ptr, mapping: MappingAddressBytes
...
+    let main_trie_start = cast(main_trie.value._data.value.dict_ptr_start, DictAccess*);
+    let main_trie_end = cast(main_trie.value._data.value.dict_ptr, DictAccess*);
+
+    let (squashed_main_trie_start, squashed_main_trie_end) = default_dict_finalize(
+        main_trie_start, main_trie_end, cast(main_trie.value.default.value, felt)
+    );
+
+    tempvar squashed_main_trie = TrieAddressOptionalAccount(
+        new TrieAddressOptionalAccountStruct(
+            secured=bool(1),
```
Eikix, line 1398: why do we hardcode values here but below in the storage tries we use the storage_tries attributes `.secured` and `.default`

Eikix (dismissed):
one nit but ready to merge technically

PR #1260, opened 2025-04-07
feat: hash state and trie diff segments

### LINKED ISSUE #1241: feat: hash the state and storage diffs in STF program
**Goal:** Output cryptographic commitments (hashes) representing the state diff (changes to accounts and storage) generated during a block's execution, instead of only the final state root.

**Current Process & Challenge:**
- The `state_transition` function (`cairo/ethereum/cancun/fork.cairo:1358`) calls `apply_body` (`fork.cairo:958`) to execute transactions.
- `apply_body` currently calculates the final `state_root` by calling `state_root` (`cairo/ethereum/cancun/state.cairo:1282`).
- Inside `state_root`, the `main_trie` (account data) is finalized using `default_dict_finalize` (`state.cairo:1293`).
- Inside `storage_roots` (`state.cairo:1090`, called by `state_root`), the `storage_tries` are finalized using `default_dict_finalize` (`state.cairo:1111`).
- **Problem:** These finalized dictionaries are not returned back to the top-level `state_transition` function, so we cannot use them to hash the state diff.

> Note: Ideally, we would like to get rid of `state_transition` altogether and only rely on the diffs. However, we want to make this change incrementally, so we will keep the `state_transition` function and add the diffs inside it.

**Proposed Solution:**
1.  **Track Changes:** Modify `state_root` to also return the state, storage diff segments.
2.  **Pass Diffs:** Modify `apply_body` (`fork.cairo:958`) to return them.
3.  **Process Diffs:** In `state_transition` (`fork.cairo:1358`) call a function dedicated to, hash the raw diffs recursively, iterate over the diffs and hashing them together.
4.  **Output:** Return the commitments on state, storage diffs from `state_transition`.
5.  **Test**: Add tests to ensure that the diffs are correctly tracked and hashed.

Strategy to process the diffs:
- Start with an initial state 0
- For each entry in the diff:
   - cast the `dictaccess` to the corresponding `AddressAccountDiffEntry` or `StorageDiffEntry`
  - hash the entry using either `poseidon_account_diff` or `poseidon_storage_diff`
  - hash(prev_state, hash(entry)) together

**Task List:**
- [x] Modify `state_root` to take as input the squashed state instead of squashing it in the function.
- [x] Modify `storage_roots` function to take as input the squashed state instead of squashing it in the function
- [x] Update main.cairo to compute hash commitments on the state object returned from `state_transition`
- [x] Add tests to verify that the state / storage is properly committed

Eikix:
closes #1241 

## Summary of Changes

This PR implements cryptographic commitments for state diffs (account and storage changes) in the Ethereum Cancun fork, addressing the issue of outputting these hashes alongside the final state root. Below are the key changes:

### `main.cairo`
- **Added State Diff Commitments**: Extended the `main` function to compute and output hashes for account and storage diffs.
  - Imported `hash_state_account_diff` and `hash_state_storage_diff` from `mpt.hash_diff`.
  - Computed commitments using the state object and added them to the output pointer (increased size from 6 to 8).
- **Purpose**: Outputs diff hashes for verification, fulfilling the issue's goal.

### `args_gen.py`
- **Adjusted Hashing Logic**:
  - Updated `hash_args` with an optional `with_storage_root` parameter to exclude `storage_root` when needed.
  - Modified `hash_poseidon` to skip the new `storage_root` for `new_value`, as it’s not computable from partial changes.
- **Purpose**: Aligns hashing with partial state diff requirements.

### `hash_diff.cairo`
- **Enhanced Diff Hashing**:
  - Updated `poseidon_account_diff` to hash only the pre-state `storage_root` (buffer size reduced from 15 to 13).
  - Added recursive hashing functions:
    - `hash_account_diff_segment` and `hash_storage_diff_segment`: Hash diff segments, returning 0 for empty lists.
    - `hash_state_account_diff` and `hash_state_storage_diff`: Hash entire state diffs by casting dictionary pointers.
  - Used a Merkle-like approach with Poseidon hashing for iterative diff processing.
- **Purpose**: Implements the proposed recursive diff hashing strategy.

### `test_hash_diff.py`
- **Added Tests**:
  - `test_hash_account_diff_segment`: Verifies account diff hashing, including empty cases.
  - `test_hash_storage_diff_segment`: Verifies storage diff hashing.
- **Purpose**: Ensures correctness of the new hashing functions.

### Additional Notes
- Updated imports in `hash_diff.cairo` for compatibility with state and diff types.
- Assumes `state_transition` provides a state object with diff data (not shown in this diff).
- TODOs remain for future work (e.g., empty segment handling, MPT verification).

This PR incrementally adds diff commitments while retaining `state_transition`, completing all tasked changes (modifying `state_root`, `storage_roots`, updating `main.cairo`, and adding tests—some changes may be in other files).


python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -47,3 +59,97 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+
+func hash_account_diff_segment{poseidon_ptr: PoseidonBuiltin*}(account_diff: AccountDiff) -> felt {
+    alloc_locals;
+    let len = account_diff.value.len;
+    if (len == 0) {
+        // TODO: Do we return 0 as the hash of an empty segment?
+        // We don't want to raise here as a block could have 0 account diffs.
+        return 0;
+    }
+    let accumulator = poseidon_account_diff(account_diff.value.data[0]);
```
enitrat, line 71: shouldn't accumulator start at 0 ?

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -47,3 +59,97 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+    accumulator: felt, start_ptr: AddressAccountDiffEntry*, i: felt, len: felt
+) -> felt {
+    if (i == len) {
+        return accumulator;
+    }
+    let next_hash = poseidon_account_diff(start_ptr[i]);
+
+    let (buffer) = alloc();
+    assert buffer[0] = accumulator;
+    assert buffer[1] = next_hash;
```
enitrat, line 86: at `i=0` here you hash diff[0] with itself right?

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -47,3 +59,97 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+    }
+    let next_hash = poseidon_storage_diff(start_ptr[i]);
+
+    let (buffer) = alloc();
+    assert buffer[0] = accumulator;
+    assert buffer[1] = next_hash;
+    let (accumulator) = poseidon_hash_many(2, buffer);
+
+    return _hash_storage_diff_inner(accumulator, start_ptr, i + 1, len);
+}
```
enitrat, line 134: can we make this generic by passing a `codeoffset` and doing `call abs hash_util` instead?

python/mpt/src/mpt/hash_diff.cairo
Eikix: no since i start at i = 1

python/mpt/src/mpt/hash_diff.cairo
Eikix: why would we hash 0?

Eikix: I propose to define it as a recurrent sequence: 
u_0 = poseidon_account_diff(account_diff.value.data[0]);
u_n+1=poseidon(u_n, poseidon_account_diff(account_diff.value.data[n+1]))

enitrat: we don't hash 0
initial_acc is 0
then you start the loop at i=0

Eikix: but then the inner function has a different behavior for i == 0?
since it can't use the recursive construct "poseidon_hash_many(acc, new)"

enitrat, line 71:
```suggestion
    let accumulator = 0
```

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -47,3 +59,97 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+func hash_account_diff_segment{poseidon_ptr: PoseidonBuiltin*}(account_diff: AccountDiff) -> felt {
+    alloc_locals;
+    let len = account_diff.value.len;
+    if (len == 0) {
+        // TODO: Do we return 0 as the hash of an empty segment?
+        // We don't want to raise here as a block could have 0 account diffs.
+        return 0;
+    }
+    let accumulator = poseidon_account_diff(account_diff.value.data[0]);
+    let final_hash = _hash_account_diff_inner(accumulator, account_diff.value.data, 1, len);
```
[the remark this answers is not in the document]
enitrat: yes? what would be the problem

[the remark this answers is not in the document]
enitrat: the first state is the empty, zero state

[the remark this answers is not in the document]
enitrat: also I just realized, I think it's sub-optimal to hash on each (prev, new) on each iteration, since we can hash as many values as we want: 
It's _probably_ better to just accumulate a big buffer of values to hash (all felts) and then do `poseidon_hash_many(buffer_len, buffer)`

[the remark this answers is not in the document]
Eikix: So instead of hashing recursively, we hash all the entries, and hash once the hash of all entries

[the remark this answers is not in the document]
enitrat: yes, I think it alignes better with poseidon's design

Eikix:
Seen with @enitrat who is picking up the PR at current commit

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -53,3 +74,172 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+// @notice Hashes a segment of account diffs into a single commitment
+// @dev All entries are hashed in order and accumulated in a buffer
+//      The buffer is then hashed to produce the final commitment
+// @param account_diff The account diff struct containing all diff entries
+// @return The hash commitment of all account diffs, or 0 if empty
+func hash_account_diff_segment{poseidon_ptr: PoseidonBuiltin*}(account_diff: AccountDiff) -> felt {
+    alloc_locals;
+    let len = account_diff.value.len;
+    if (len == 0) {
+        // TODO: Do we return 0 as the hash of an empty segment?
```
enitrat, line 89: I think it's fine tbh

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -53,3 +74,172 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
+// @dev Recursively processes each account diff entry and stores its hash
+// @param buffer The buffer to store the accumulated hashes
+// @param account_diff The account diff struct containing all entries
+// @param i The current index being processed
+// @return The number of hashes accumulated (buffer length)
+func _accumulate_diff_hashes{poseidon_ptr: PoseidonBuiltin*}(
+    buffer: felt*, account_diff: AccountDiff, i: felt
+) -> felt {
+    if (i == account_diff.value.len) {
+        return i;
```
Eikix, line 109: nit: is it useful to return `i` since we coerce it to be equal to account_diff.value.len?

Eikix (commented):
nice! gg! approved (i can't approve)

python/mpt/src/mpt/hash_diff.cairo
enitrat: I left it as a "good practice" of returning the numbers of elements iterated over when constructing a memory segment, but it's not very useful

python/mpt/src/mpt/hash_diff.cairo
enitrat, line 89:
```suggestion
```

enitrat (approved):

PR #1262, opened 2025-04-07
fix: handle point at infinity case in bnp init and blsp init

### LINKED ISSUE #1252: cov: handle point at infinity in BLSP and BLSP2 strategy
# What
Current strategy for `BLSP` and `BLSP2` does not include the point at infinity.

# How
Add with `Z1` and `Z2` from py_ecc
Update assertions in serde/args_gen

### LINKED ISSUE #1257: fix: handle point at infinity in bnp_init and blsp_init
# What

Current implem of `bnp_init` and `blsp_init` does not handle the point at infinity, when they should.

# How

Add point at infinity check.

zmalatrax:
Closes #1257
Closes #1252

### DIFF
--- a/cairo/tests/ethereum/crypto/test_bls12_381.py
+++ b/cairo/tests/ethereum/crypto/test_bls12_381.py
@@ -163,20 +163,14 @@ segments.load_data(ids.b_inv.address_, [blsf2_struct_ptr])
 
         @given(p=...)
         def test_blsp_double(self, cairo_run, p: Optimized_Point3D[BLSF]):
-            expected = double(p)
-            if is_inf(expected):
-                expected = Z1
-            else:
-                expected = normalize1(expected)
+            p_double = double(p)
+            expected = Z1 if is_inf(p_double) else normalize1(p_double)
             assert cairo_run("blsp_double", p) == expected
 
         @given(p=..., n=...)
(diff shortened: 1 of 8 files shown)


cairo/tests/ethereum/crypto/test_bls12_381.py
```diff
@@ -197,15 +212,25 @@ def test_blsp2_add_negated_y(self, cairo_run, p: Optimized_Point3D[BLSF2]):
...
+            assert cairo_run("blsp2_add", p, q) == expected
 
         @given(p=...)
         def test_blsp2_double(self, cairo_run, p: Optimized_Point3D[BLSF2]):
-            assert cairo_run("blsp2_double", p) == normalize1(double(p))
+            expected = double(p)
+            if is_inf(expected):
+                expected = Z2
+            else:
+                expected = normalize1(expected)
```
obatirou, line 226: Is this necessary?
shouldn't this be done in serde in some way?
```suggestion
            p_double = double(p)
            expected = Z2 if is_inf(p_double) else normalize1(p_double)
```

zmalatrax: We're making modifs on the `py_ecc` expected result, so it won't be handled by args_gen/serde, right?
(args_gen does `p`in python --> `p` in Cairo, and serde does `result` in Cairo --> `result` in python ; to compare `result` and `expected`)

obatirou (approved):

Issue #1029, opened 2025-03-18
feat: modify main.cairo entrypoint to incorporate the new trie diff logic


Eikix:
## Why

we want to prove the STF

## What

- [ ] add trie diff logic in main.cairo (new inputs + call `compute_diff` between the two tries)
- [ ] hash the returned two segments (account diff + storage diff) 
- [ ] assert this hash is equal to the hash commitment of the post state

mimisavage:
I’d like to help with this.

PR #1263, opened 2025-04-07
feat: add state and trie logic to main.cairo

Eikix:
addresses #1029


python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -443,33 +450,34 @@ func _process_storage_diff{
...
-        let left_decoded = U256_from_rlp(left.value.value.value.bytes);
+        let left_u256 = U256_from_rlp(left.value.value.value.bytes);
         tempvar range_check_ptr = range_check_ptr;
         tempvar bitwise_ptr = bitwise_ptr;
         tempvar poseidon_ptr = poseidon_ptr;
         tempvar keccak_ptr = keccak_ptr;
     } else {
-        tempvar left = left;
+        // If the left doesn't have a value, it means the storage key was deleted, the new value is 0.
+        tempvar left_u256 = U256(cast(u256_zero, U256Struct*));
```
enitrat, line 460: based on how I understand things, the state diff will also have a 0 there, so I am surprised this is even needed (note to self). 

There might be an issue with how the `ids` object is built, leading to a slight mishap in how this value was represented in the debug data...

enitrat:
#1313

PR #1264, opened 2025-04-07
feat: check extension nodes well formed

### LINKED ISSUE #1255: feat: extension nodes must have non null key_segment and subnode
extension nodes must have non null key_segment and subnode

enitrat:
Closes #1255

### DIFF
--- a/python/mpt/src/mpt/utils.cairo
+++ b/python/mpt/src/mpt/utils.cairo
@@ -224,6 +224,37 @@ func check_leaf_node(path: Bytes, node: LeafNode) {
     }
 }
 
+func check_extension_node(node: ExtensionNode) {
+    alloc_locals;
+
+    let key_segment = node.value.key_segment;
+    let subnode = node.value.subnode;
+
+    if (key_segment.value.len == 0) {
+        raise('ValueError');
+    }
+
+    let bytes_variant = subnode.value.bytes.value;
+    if (cast(bytes_variant, felt) != 0) {
+        if (bytes_variant.len == 0) {
+            raise('ValueError');
+        }
+        return ();
+    }
+
+    let sequence_variant = subnode.value.sequence.value;
+    if (cast(sequence_variant, felt) != 0) {
+        if (sequence_variant.len == 0) {
+            raise('ValueError');
+        }
+        return ();
+    }
+
+    with_attr error_message("ValueError: Unsupported extension node value variant") {
+        jmp raise.raise_label;
+    }
+}
+
 // @notice Sorts an AccountDiff struct in descending order based on the key.
 // This function implies that the original AccountDiff struct does not contain any duplicate keys.
 // @dev The sorted segment is returned from the hint.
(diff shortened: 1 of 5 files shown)


python/mpt/src/mpt/utils.cairo
```diff
@@ -189,8 +189,72 @@ func check_branch_node(node: BranchNode) {
...
+        }
+        return ();
+    }
+
+    with_attr error_message("ValueError: Unsupported leaf value variant") {
+        jmp raise.raise_label;
+    }
+}
+
+func check_extension_node(node: ExtensionNode) {
```
Eikix, line 227: nice:)

Eikix: i'm realizing as we speak that also you're not allowed to have an extension node point to another extension node, not sure how we can implement that

enitrat: this would need to be a check made when running `compute_diff`: you'd need to pass the parent nodes, and ensure the resolved child is not an extension

Eikix: https://github.com/kkrt-labs/keth/issues/1269

python/mpt/src/mpt/utils.py
```diff
@@ -97,3 +97,16 @@ def check_leaf_node(path: Bytes, node: LeafNode) -> None:
...
+def check_extension_node(node: ExtensionNode) -> None:
+    """
+    Check that an extension node is valid
+     - Extension nodes must have a non-zero key segment
+     - Extension nodes must have a non-zero subnode
+    """
+    if len(node.key_segment) == 0:
+        raise ValueError("Invalid extension node, expected a non-zero key segment")
+
+    if len(node.subnode) == 0:
```
Eikix, line 111: what is subnode is none, can you call len on None?

python/mpt/tests/src/mpt/test_utils.py
```diff
@@ -112,6 +118,30 @@ def leaf_node_could_be_invalid_strategy(draw):
             return path, LeafNode(rest_of_key=rest_of_key, value=value)
 
 
+@st.composite
+def extension_node_could_be_invalid_strategy(draw):
```
Eikix, line 122: invalid subnode could be:
None
Empty List (empty embedded node)
Empty Bytes (i.e. empty hash32 digest)

python/mpt/src/mpt/utils.py
enitrat: can a subnode be None?

Eikix: ah no you're right, thanks

Eikix:
```
Extended: TypeAlias = Union[
    Sequence["Extended"], bytearray, bytes, Uint, FixedUnsigned, str, bool, RLP
]

```

Eikix (approved):
kgtm

PR #1266, opened 2025-04-07
feat: check branch node value

### LINKED ISSUE #1254: feat: raise/panic if value in a branch node is set
## Why

paths in Eth MPTs are keccak256, which are 32 bytes long, which means a branch node cannot have a value, only leaf nodes.

## What
raise/panic if value in a branch node is set

enitrat:
Closes #1254

### DIFF
--- a/python/mpt/tests/src/mpt/test_utils.py
+++ b/python/mpt/tests/src/mpt/test_utils.py
@@ -54,13 +54,19 @@ def branch_node_could_be_invalid_strategy(draw):
     1. Non-empty value.
     2. Less than two non-null subnodes.
     """
-    test_cases = st.sampled_from(["non_empty_value", "less_than_two_subnodes", "ok"])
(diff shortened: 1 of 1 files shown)


python/mpt/src/mpt/utils.py
```diff
@@ -75,6 +75,9 @@ def check_branch_node(node: BranchNode) -> None:
     """
     Check that a branch node is valid.
     """
+    if not isinstance(node.value, bytes) and len(node.value) == 0:
```
Eikix, line 78:
```suggestion
    if not isinstance(node.value, bytes) or (isinstance(node.value, bytes) and len(node.value) > 0):
```

enitrat:
ah yes indeed

Eikix (approved):
lgtm, though we don't test `value != bytes variant` (as pointed out by codecov)

PR #1270, opened 2025-04-08
fix: unit mismatch in leaf node path length check

0xsimka:
the check was adding nibbles to bytes, which led to incorrect validation
updated it to `nibbles_len + path_len * 2 == 64` so both are in nibbles now.


Eikix:
Hi, `path` in that instance is already expressed as a nibble list, there is no need to change it.

PR #1276, opened 2025-04-08
hash 0 if storage is deleted or created

### LINKED ISSUE #1272: bug: prove_block script broken
Problem:
prove_block script is broken due to `main.cairo::state_transition` asserting on state root equality with the block, while zkpi_to_eels 'patch' of state root (partial state root injected as real state root) was deleted

Eikix:
closes #1272

### DIFF
--- a/python/mpt/src/mpt/hash_diff.cairo
+++ b/python/mpt/src/mpt/hash_diff.cairo
@@ -65,11 +65,25 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
 
     assert buffer[0] = diff.value.key.value;
     // Inline the prev storage value (U256)
-    assert buffer[1] = diff.value.prev_value.value.low;
-    assert buffer[2] = diff.value.prev_value.value.high;
+    if (cast(diff.value.prev_value.value, felt) == 0) {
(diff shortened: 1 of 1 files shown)


python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -65,11 +65,25 @@ func poseidon_storage_diff{poseidon_ptr: PoseidonBuiltin*}(diff: StorageDiffEntr
...
-    assert buffer[2] = diff.value.prev_value.value.high;
+    if (cast(diff.value.prev_value.value, felt) == 0) {
+        // Note: in case of State Diffs, writing the default U256(0) value deletes the key, writing a null pointer instead.
+        // In that case, we hash (0, 0) as the value.
+        assert buffer[1] = 0;
+        assert buffer[2] = 0;
+    } else {
+        assert buffer[1] = diff.value.prev_value.value.low;
+        assert buffer[2] = diff.value.prev_value.value.high;
+    }
```
enitrat, line 76:
```suggestion
    assert buffer[1] = diff.value.prev_value.value.low;
    assert buffer[2] = diff.value.prev_value.value.high;
```

this shouldn't ever happen because the dict is a default_dict with de
fault_value U256(0), so if the value is not present before, the `prev` is `U256(0)`

n-eq:
works on my side ✅

PR #1280, opened 2025-04-08
feat: hash preimages in `node_store_get`

### LINKED ISSUE #1131: feat [soundness]: re-hash nodes gotten from calls to `node_store_get` to check that the hash map is sound
## Why

Our program takes as public input a node store, that stores all MPT nodes in the partial state.
As we traverse the MPT, when we query a node for a specific hash, we need to be sure the node is addressed by hash, i.e. that hash(rlp(node)) == digest of parent node

## What

Whenever in compute_diff or resolve, we 'resolve' a node in the node store, hash it and verify that the bytes match the digest used to lookup the node store

Eikix:
closes #1131


cairo/ethereum_rlp/rlp.cairo
```diff
@@ -222,6 +222,48 @@ namespace ExtendedImpl {
         );
         return extended;
     }
+
+    func from_simple(simple: Simple) -> Extended {
```
Eikix, line 226: a bit unlucky here: 
rlp decode returns a `Simple`, but all the types in the trie.py/cairo (branch nodes, subnodes etc.) are `Extended`

enitrat, line 226: doc

python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -1792,9 +1797,22 @@ func node_store_get{poseidon_ptr: PoseidonBuiltin*, node_store: NodeStore}(
...
-    // Cast the result to an OptionalInternalNode and return
-    tempvar res = OptionalInternalNode(cast(pointer, InternalNodeEnum*));
-    return res;
+
+    // Cast the result to a Bytes, hash it to check invariant and RLP-decode it.
+    tempvar encoded_node = Bytes(cast(pointer, BytesStruct*));
+    if (encoded_node.value.len == 0) {
+        let res = OptionalInternalNode(cast(0, InternalNodeEnum*));
+        return res;
+    }
```
enitrat, line 1806: no hash to do in this case? in this case it means that the RLP encoding of the node is `b""`: in which cases does this happen?

enitrat (approved):
okpm

cairo/ethereum_rlp/rlp.cairo
```diff
@@ -222,6 +222,52 @@ namespace ExtendedImpl {
         );
         return extended;
     }
+
+    // @notice Converts a Simple to an Extended. This is needed because `decode` returns a Simple but `trie.cairo` types are Extended.
+    // @dev in Python, this conversion is trivial since Simple is included in Extended.
+    //      in Cairo, we need to check the type of the Simple and convert it to the correct Extended variant.
+    //      Perhaps we should use either only Extended or only Simple in the RLP encoding library.
+    func from_simple(simple: Simple) -> Extended {
```
enitrat, line 230: also, we can't test this because it's in a namespace?

cairo/ethereum_rlp/rlp.cairo
```diff
@@ -222,6 +222,52 @@ namespace ExtendedImpl {
         );
         return extended;
     }
+
+    // @notice Converts a Simple to an Extended. This is needed because `decode` returns a Simple but `trie.cairo` types are Extended.
+    // @dev in Python, this conversion is trivial since Simple is included in Extended.
+    //      in Cairo, we need to check the type of the Simple and convert it to the correct Extended variant.
+    //      Perhaps we should use either only Extended or only Simple in the RLP encoding library.
```
enitrat, line 229:
```suggestion
```

python/mpt/src/mpt/trie_diff.cairo
enitrat: ah ok, I think this should be done prior to casting the pointer then:

```suggestion
    if (pointer == 0) {
        let res = OptionalInternalNode(cast(0, InternalNodeEnum*));
        return res;
    }
```

enitrat: and giving a second thought, when are we querying nodes that are not in the node store?

cairo/ethereum_rlp/rlp.cairo
Eikix: Mhm I think since it's imported in another file, we can test it, let me try

Eikix: Ah no it doesn't work

python/mpt/tests/src/mpt/test_trie_diff.py
```diff
@@ -402,7 +395,13 @@ def test_resolve(self, cairo_run, node_store: Mapping[Hash32, InternalNode], dat
         keys.append(keccak256("non_existing".encode()))
 
         for key in keys:
-            _, cairo_result = cairo_run("resolve", small_store, node=bytes(key))
+            try:
+                _, cairo_result = cairo_run("resolve", small_store, node=bytes(key))
+            except Exception:
```
enitrat, line 400: can we catch an error message here?

python/mpt/tests/src/mpt/test_trie_diff.py
```diff
@@ -169,8 +163,13 @@ def test_node_store_get(self, cairo_run, node_store, data):
         keys.append(keccak256("non_existing".encode()))
 
         for key in keys:
-            _, result = cairo_run("node_store_get", small_store, key)
-            assert result == small_store.get(key)
+            try:
+                _, result = cairo_run("node_store_get", small_store, key)
+            except Exception:
```
enitrat, line 168: which exception?

python/mpt/tests/src/mpt/test_trie_diff.py
```diff
@@ -402,7 +395,14 @@ def test_resolve(self, cairo_run, node_store: Mapping[Hash32, InternalNode], dat
         keys.append(keccak256("non_existing".encode()))
 
         for key in keys:
-            _, cairo_result = cairo_run("resolve", small_store, node=bytes(key))
+            try:
+                _, cairo_result = cairo_run("resolve", small_store, node=bytes(key))
+            except Exception as cairo_error:
+                assert "Dict Error: No value found for key" in str(cairo_error)
+                # We can't use strict_raises here because the error is a RuntimeError in Cairo
```
enitrat, line 402: is there a better way?

python/mpt/tests/src/mpt/test_trie_diff.py
enitrat:
```suggestion
            except Exception as cairo_error:
                assert "Dict Error: No value found for key" in str(cairo_error)
```

(i guess)

enitrat (commented):
lgtm

PR #1283, opened 2025-04-08
fix: sort state diff keys in ascending order

### LINKED ISSUE #1282: bug: sort_segments is sorting in the wrong order
Unlike perviously believed, the squashed dict segment is sorted in _ascending_ order, not _descending_

enitrat:
Closes #1282


Eikix (approved):
lgtm

PR #1284, opened 2025-04-08
ci: use shared rust-cache key

enitrat:
As per https://github.com/Swatinem/rust-cache the `shared-key` option overrides the default job-based key, so that rust caches can be reused among jobs. This should significantly reduce the size of rust caches, which should be divided by 3 (ef-tests, python_unit, rust_unit)


Eikix (approved):
lgtm

PR #1286, opened 2025-04-09
feat: decompress g2

### LINKED ISSUE #1180: feat: decompress_g2
# What

Implement the `decompress_g2(G2Compressed) -> G2Uncompressed` function.

# How

[py_ecc specs](https://github.com/ethereum/py_ecc/blob/36f5ef87ef0d8f5647af66ad8273fb059656fc8a/py_ecc/bls/point_compression.py#L175-L226)

### LINKED ISSUE #1182: feat: modular_squareroot_in_FQ2
# What 
Implement the "square root" (quadratic residue) of a point in Fq2

# How 

Make use of non-determinism

[py_ecc specs](https://github.com/ethereum/py_ecc/blob/36f5ef87ef0d8f5647af66ad8273fb059656fc8a/py_ecc/bls/point_compression.py#L123)

zmalatrax:
Closes #1180 
Closes #1182 

The `decompress_g2_hint` calls `decompress_G2` from py_ecc, which computes the modular square root of $`x^3 + b`$ in $`\mathbb{F}_q^2`$ and chooses the "right" `y` among the two possibilities.

This specific part (modular sqrt + choosing y) could be inlined, but previous checks are not expensive and provides additional checks on the input values of the hint.

### DIFF
--- a/cairo/ethereum/crypto/bls12_381.cairo
+++ b/cairo/ethereum/crypto/bls12_381.cairo
@@ -427,6 +427,19 @@ func blsp_mul_by_bits{
     return blsp_mul_by_bits(doubled_p, bits_ptr, bits_len, current_bit + 1, new_result);
 }
 
+// G2Compressed is the Cairo equivalent of Tuple[int, int] from py_ecc.
+// Where each int is 48 bytes long.
+struct G2CompressedStruct {
+    c0: U384,
+    c1: U384,
+}
+
+struct G2Compressed {
+    value: G2CompressedStruct*,
+}
+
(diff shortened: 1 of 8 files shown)


obatirou:
After careful review, it seems this function is not necessary. 
It is used by only by `signature_to_G2` which itself is used on a constant.
https://github.com/ethereum/execution-specs/blob/2926c00c1d130e5d0641d278012d42267f3feaf3/src/ethereum/crypto/kzg.py#L173

Hence, a constant will be used directly in cairo code

PR #1287, opened 2025-04-09
dev: allow small decrease in coverage on ethereum_rlp

### LINKED ISSUE #1198: kaizen: "100% green CI in valid in PRs"
## Why

A good CI should be green when a PR is completed:
- feature is done
- tests are implemented so that coverage is 100%

## Problem 

When a PR is done, codecov coverage is often red.
We ignore it most of the time, creating a bad behaviour: red CI should almost never be ignored.

## What

Investigate during a kaizen how we can prevent codecov CI to fail due to indirect changes.

Eikix:
closes #1198 

strategy choice (very lean strategy, can be improved later):
- allow for slight decrease in coverage for ethereum_rlp package (indirect decrease) as I've noticed that over the last 5 PRs, the CI fails from an indirect decrease in coverage of ethereum_rlp package due to fuzzing framework being random

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -10,6 +10,7 @@ coverage:
       ethereum_rlp:
         paths:
           - cairo/ethereum_rlp
+        threshold: 1.5%
       ethereum_types:
         paths:
           - cairo/ethereum_types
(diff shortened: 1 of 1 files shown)


Eikix:
> can you also add this https://docs.codecov.com/docs/codecovyml-reference#codecovnotifywait_for_ci

yes good idea

enitrat (approved):

PR #1292, opened 2025-04-09
Compile keth implement hints

ClementWalter:


enitrat (approved):
lgtm but ideally should we use the `get_cairo_program`  from testing/compiler.py to avoid duplication?

ClementWalter:
not sure we want to have the pickle cache system for this "prod" compilation

enitrat:
do we know whether it's because of the hint execution itself, or building the hint context beforehand ?

PR #1294, opened 2025-04-09
add a check on extension nodes

### LINKED ISSUE #1269: feat: check extension node subnode is never an extension node
## Why

As we traverse our inputted MPT, we want to make sure it is well formed.

## What

An extension node must not have as subnode another extension node.

## How
from @enitrat 
> this would need to be a check made when running compute_diff: you'd need to pass the parent nodes, and ensure the resolved child is not an extension

but open to other proposal

Eikix:
closes #1269 

this proposed PR deals with minimal flags `left_parent_is_extension` and `right_parent_is_extension`, but is quite verbose inside business logic code.
we could instead pass full parent left and right and in `check_extension_node` calls pass the corresponding parent node in order to assert whether both node and parent_node are extension

PS: i also noticed we don't call the python function `check_extension_node` in `trie_diff.py`

### DIFF
--- a/python/mpt/src/mpt/trie_diff.cairo
+++ b/python/mpt/src/mpt/trie_diff.cairo
@@ -397,7 +397,12 @@ func _process_account_diff{
     tempvar new_address = OptionalAddress(new address.value);
 
     _compute_diff(
-        left=left_storage_root, right=right_storage_root, path=new_path, account_address=new_address
+        left=left_storage_root,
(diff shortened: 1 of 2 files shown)


enitrat:
> we could instead pass full parent left and right and in check_extension_node calls pass the corresponding parent node in order to assert whether both node and parent_node are extension

Sounds like a better idea; all checks should be done in the same place

PR #1298, opened 2025-04-09
fix: ids object construction on casted types

### LINKED ISSUE #1279: bug: pyo3 raise
To reproduce, checkout to https://github.com/kkrt-labs/keth/pull/1271 and 
` uv run --reinstall pytest -k ecpairing_inputs_d22g0v0_Cancun -vvvvvvvv -s -log-cli-level="TRACE"`

Image

### LINKED ISSUE #1297: bug: ids object fails de-referencing 0 pointers
```
        %{
            if ids.err.value == 0:
                logger.trace_cairo(f"OpEnd")
        %}
```

running this fails if err.value == 0 (as it's not a pointer, i guess?)

enitrat:
Closes #1297
Might close #1279

### DIFF
--- a/crates/cairo-addons/src/vm/vm_consts.rs
+++ b/crates/cairo-addons/src/vm/vm_consts.rs
@@ -294,9 +294,7 @@ impl PyVmConst {
                     let var = CairoVar {
                         name: format!("{}.{}", parent_name, member_name),
                         value: Some(value.clone()),
-                        address: Some(
-                            value.get_relocatable().expect("Pointer value is not relocatable"),
-                        ),
+                        address: self.get_address()?,
                         var_type: member_type,
                     };
                     let py_pointee = PyVmConst { var, vm: self.vm, identifiers: self.identifiers };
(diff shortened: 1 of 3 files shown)


crates/cairo-addons/src/vm/vm_consts.rs
```diff
@@ -322,15 +320,16 @@ impl PyVmConst {
...
-
-                Ok(Some(addr))
+                // Note: if the variable is NOT a pointer (e.g. we cast something to a pointer),
+                // like tempvar my_pointer_struct = MyPointerStruct(cast(0, felt*));
+                // then the value is 0 and we should return `self.var.address`
+                let pointer_value = self.var.value.as_ref();
+
+                match pointer_value {
+                    Some(MaybeRelocatable::RelocatableValue(rel)) => Ok(Some(*rel)),
+                    Some(MaybeRelocatable::Int(_)) => Ok(self.var.address),
```
Eikix, line 330: are we saying that here, we expect `Some(MaybeRelocatable::Int(0))`?

should we use only Some(MaybeRelocatable::Int(0)) and fail otherwise?

enitrat: Technically it could be any non pointer value not just 0

enitrat: it's not based on what "we" want it's based on VM rules, you can cast any felt to a felt*

PR #1303, opened 2025-04-10
feat: bytes_to_kzg_proof

### LINKED ISSUE #1174: feat: bytes_to_kzg_proof
# What

Implement `bytes_to_kzg_proof(Bytes48) -> KZGProof`

# How

Requires #1169 

[specs](https://github.com/ethereum/consensus-specs/blob/86fb82b221474cc89387fa6436806507b3849d88/specs/deneb/polynomial-commitments.md#bytes_to_kzg_proof), [EELS specs](https://github.com/ethereum/execution-specs/blob/2926c00c1d130e5d0641d278012d42267f3feaf3/src/ethereum/crypto/kzg.py#L116-L121)

obatirou:
Closes #1174

### DIFF
--- a/cairo/ethereum/crypto/kzg.cairo
+++ b/cairo/ethereum/crypto/kzg.cairo
@@ -27,8 +27,6 @@ from ethereum.utils.numeric import (
     Bytes32_from_be_bytes,
     U384Struct,
 )
-from cairo_ec.circuits.ec_ops_compiled import assert_on_curve
-from cairo_ec.curve.bls12_381 import bls12_381
 from ethereum.crypto.bls12_381 import (
     BLSF,
     BLSFStruct,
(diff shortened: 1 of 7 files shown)


zmalatrax (dismissed):

cairo/ethereum/crypto/kzg.cairo
```diff
@@ -348,3 +349,17 @@ func bytes_to_kzg_commitment{
...
     return (KZGCommitment(b.value), bool(1));
 }
+
+func bytes_to_kzg_proof{
+    range_check_ptr,
+    bitwise_ptr: BitwiseBuiltin*,
+    range_check96_ptr: felt*,
+    add_mod_ptr: ModBuiltin*,
+    mul_mod_ptr: ModBuiltin*,
+}(b: Bytes48) -> (KZGProof, bool) {
```
enitrat, line 359: do you absolutely need to return a boolean flag for success instead of returning an Exception type ?

enitrat: If I understand well in Python this is an AssertionError right?

obatirou: Indeed it could be done with an Exception type

enitrat: I would rather have an exception type for consistency with the codebase: we can use `EthereumException` from `cairo/ethereum/exceptions.cairo`, add a new `const AssertionError = 'AssertionError'`, and return an `EthereumException*`

enitrat: so that in pytest we can use the `try... except... with strict_raises(type(error))` pattern and have the `cairo_run` call return only the non-exception type

cairo/tests/utils/args_gen.py
```diff
@@ -647,6 +647,19 @@ def to_transient_storage(self) -> TransientStorage:
...
+    lambda x: inspect.isclass(x) and issubclass(x, Exception),
+)
+
+builtins_exception_mappings = {
+    (
+        "builtins",
+        f"{name}",
+    ): cls
+    for name, cls in builtins_exception_classes
+}
```
enitrat, line 661:
```suggestion
// TODO: builtins exceptions are defined in the ethereum.exceptions module for now.
// Refactor if required
builtins_exception_mappings = {
    (
        "ethereum",
        "exceptions",
        f"{name}",
    ): cls
    for name, cls in builtins_exception_classes
}
```

no impact yet, however i dont want to risk surprises later, and dont think its important enough to make a new package specifically for builtin errors yet

enitrat (approved):

Issue #1302, opened 2025-04-10
feat: use `prestate` instead of computed state from zkpi MPTs in `prove_block`


enitrat:
# Why

Currently, in `prove_block` and `test_fork`'s `zkpi_fixture`, we were using `EthereumTries.to_state` to get the `PreState` that will be used for the transaction.

This is not a correct approach, because from a MPT perspective, accounts that will be created in this transaction and accounts that empty, interacted with, but not mutated will be absent from the MPT.

As we want to avoid using a defaultdict for our representation of the State in Cairo, we should ensure that all accounts accessed inside a transaction are part of the initial state.

As such, we should use the `preState` field from the ZKPI Json input to load the initial state.

## How

- A class `PreState` with a static method `from_data` that returns an `EELS` State
- A class ZKPI that composes 

```
    transition_db: EthereumTrieTransitionDB
    state_diff: StateDiff
    pre_state: State
``` 

Wherever we use EthereumTries::to_state, replace with usage of `pre_state `.

Make sure that the `PreState::from_data` function does not create defaultdicts for the main_trie and storage_tries. 

Then, because the state will be created __without__ defaultdicts, we will need to modify the Cairo code, in `state.cairo`, finalize, line 1392: we need to call `dict_squash` instead of `dict_finalize`.

This will not change the behavior for cases where the state is a defaultdict (such as our unit tests), and will allow us to continue running `main` with non-defaultdict states.

PR #1307, opened 2025-04-10
feat: use zkpi prestate and non-default state dicts

enitrat:
Implements #1302

### DIFF
--- a/cairo/tests/ef_tests/helpers/load_state_tests.py
+++ b/cairo/tests/ef_tests/helpers/load_state_tests.py
@@ -31,9 +31,11 @@ def prepare_state_and_code_hashes(
     for address in state._main_trie._data:
         account = trie_get(state._main_trie, address)
         if not account:
-            raise ValueError("Account is None")
-        account_code_hash = account.code_hash if account else EMPTY_BYTES_HASH
-        account_code = account.code if account else b""
+            account_code_hash = EMPTY_BYTES_HASH
+            account_code = b""
+        else:
+            account_code_hash = account.code_hash
+            account_code = account.code
         code_hash_int = int.from_bytes(account_code_hash, "little")
         code_hash_low = code_hash_int & 2**128 - 1
         code_hash_high = code_hash_int >> 128
--- a/cairo/tests/ef_tests/helpers/load_state_tests.py
+++ b/cairo/tests/ef_tests/helpers/load_state_tests.py
@@ -9,7 +9,7 @@ from typing import Any, Dict, Generator, Tuple, Union
 import pytest
 from _pytest.mark.structures import ParameterSet
 from ethereum.cancun.fork import state_transition
-from ethereum.cancun.fork_types import Account
+from ethereum.cancun.fork_types import EMPTY_ACCOUNT, Account
 from ethereum.cancun.state import State
 from ethereum.cancun.trie import root as compute_root
 from ethereum.cancun.trie import trie_get, trie_set
@@ -20,7 +20,10 @@ from ethereum_rlp import rlp
 from ethereum_rlp.exceptions import RLPException
 from ethereum_spec_tools.evm_tools.loaders.fixture_loader import Load
 from ethereum_types.bytes import Bytes
-from ethereum_types.numeric import U64, U256
+from ethereum_types.numeric import U64
+
+from mpt.ethereum_tries import EMPTY_BYTES_HASH
+from utils.fixture_loader import LoadKethFixture
 
 
 def prepare_state_and_code_hashes(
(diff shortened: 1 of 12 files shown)


Eikix (commented):
first question to understand

cairo/tests/ef_tests/helpers/load_state_tests.py
```diff
@@ -20,39 +20,49 @@
...
+            )
+        )
+
+        trie_set(state._main_trie, address, updated_account)
 
     for snap in state._snapshots:
         for address in snap._storage_tries:
             snap._storage_tries[address]._data = defaultdict(
-                lambda: defaultdict(lambda: U256(0)), snap._storage_tries[address]._data
+                lambda: snap._storage_tries[address]._data
```
Eikix, line 64: For each address in the snapshot storage tries, 
Replace that address' storage trie's dictionary by a default dictionary that returns the dictionary at that address in case of key miss?

Sounds recursive / wrong

Eikix: The dimension of snap._storage_tries[address]._data is Mapping[Bytes32, U256]

cairo/tests/utils/args_gen.py
```diff
@@ -1400,12 +1400,15 @@ def generate_trie_arg(
...
-        default_value = dict_manager.trackers[
-            dict_ptr.segment_index
-        ].data.default_factory()
+    if isinstance(arg._data, defaultdict):
+        if isinstance(dict_manager, DictManager):
+            default_value = dict_manager.trackers[
+                dict_ptr.segment_index
+            ].data.default_factory()
+        else:
+            default_value = dict_manager.get_default_value(dict_ptr.segment_index)
```
Eikix, line 1409: i don't understand when dict_manager isn't of class `DictManager`

cairo/tests/utils/args_gen.py
```diff
@@ -1400,12 +1400,15 @@ def generate_trie_arg(
...
     else:
-        default_value = dict_manager.get_default_value(dict_ptr.segment_index)
+        default_value = _gen_arg(dict_manager, segments, type(arg.default), arg.default)
```
Eikix, line 1411: what does this do?
if the dictionary in the trie is not a default dict, then the default value is the trie's default value attribute?

why would the trie's default value be different from the underlying default dict's default?

cairo/tests/utils/args_gen.py
enitrat: when it's a `RustDictManager`

cairo/tests/utils/args_gen.py
enitrat: 1. if the dictionary in the trie is a default dict, then the trie's default value is the defaultdict's default
2. if the dict is not a defaultdict, gen_arg a default_value for the trie

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -345,15 +346,94 @@ def from_data(cls, data: Dict[str, Any]) -> "EthereumTrieTransitionDB":
...
-        return self.to_state()
 
-    def to_post_state(self) -> State:
-        """Convert the post-state trie to a State object."""
-        state = State()
-        root_node = self.nodes[self.post_state_root]
-        self.traverse_trie_and_process_leaf(
-            root_node, b"", partial(self.set_account_from_leaf, state=state)
+@dataclass
+class PreState:
```
Eikix, line 351: arguably a nit but we don't need a dataclass for a single static method, wdyt?

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -44,14 +44,21 @@ func poseidon_account_diff{poseidon_ptr: PoseidonBuiltin*}(diff: AddressAccountD
...
+        assert buffer[offset] = diff.value.new_value.value.nonce.value;
+        assert buffer[offset + 1] = diff.value.new_value.value.balance.value.low;
+        assert buffer[offset + 2] = diff.value.new_value.value.balance.value.high;
+        assert buffer[offset + 3] = diff.value.new_value.value.code_hash.value.low;
+        assert buffer[offset + 4] = diff.value.new_value.value.code_hash.value.high;
+        let (account_diff_hash) = poseidon_hash_many(offset + 5, buffer);
+        return account_diff_hash;
+    }
+
+    let (account_diff_hash) = poseidon_hash_many(offset, buffer);
```
Eikix, line 61: where is the else branch?

python/mpt/src/mpt/ethereum_tries.py
enitrat: true

python/mpt/src/mpt/hash_diff.cairo
enitrat: it's an early return

cairo/tests/utils/args_gen.py
Eikix: why would we allow the defaultdict's default to not match the trie's default?

cairo/tests/ef_tests/helpers/load_state_tests.py
```diff
@@ -2,57 +2,44 @@
...
         account = trie_get(state._main_trie, address)
-        account = Account(
-            balance=account.balance,
-            nonce=account.nonce,
-            code=account.code,
-            storage_root=storage_root,
-            code_hash=account.code_hash,
-        )
-        code_hash_int = int.from_bytes(account.code_hash, "little")
+        if not account:
```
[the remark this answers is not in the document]
Eikix: nevermind, it seems that in PreState.from_data you set accounts to None (for args_gen purposes)

cairo/tests/utils/args_gen.py
enitrat: we don't, what we do is that if the dict is not a defaultdict, then we need to set a value for the trie's default field

cairo/tests/ef_tests/helpers/load_state_tests.py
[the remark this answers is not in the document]
enitrat: yeah 😅

Eikix:
tests fail on account on `hashdict_read` keys not found

cairo/utils/fixture_loader.py
```diff
@@ -20,12 +21,29 @@ class LoadKethFixture(Load):
     A superclass of the `Load` class that loads EELS-format fixtures for Keth
     """
 
-    def json_to_state(self, raw: Any) -> Any:
+    def json_to_state(self, raw: Dict[str, Any]) -> Any:
```
Eikix, line 24: are you importing this specific json_to_state in other parts of the code?

enitrat: everywhere iirc

PR #1310, opened 2025-04-10
feat: skip entries with same values when computing commitments

enitrat:
summary:

In trie diff, we have the luxury of choosing the diffs we write. as such,

if the prev and new storage values are the same -> we don't write the diff
if the prev and new acct are the same except for the storage root -> we don't write the diff
In the state diff, we don't have the luxury of choosing the diffs, they are inherited from the dict structure of cairo, so instead of not writing diffs, we skip them in the commitment phase:

in the storage diffs, if a prev and new storage values are the same (note that we replace null ptr by U256::ZERO there) -> we don't commit to that storage diff
in the acct diffs, if a prev and new acct values are the same (except for storage root) -> we don't commit to that account diff
One interesting caveat of our acct comparison is that:

we try to compare pointers to accts, if they are the same, we can stop the comparison there
if we can't say whether the pointers can be compared (allegedly according to prover not in the same segments) then we access their values are do a costly comparison


cairo/ethereum/cancun/main.cairo
```diff
@@ -51,10 +64,60 @@ func main{
...
+    local right: OptionalUnionInternalNodeExtended;
+    %{
+        from ethereum.crypto.hash import Hash32
+        from ethereum.cancun.fork_types import Address
+        from ethereum_types.bytes import Bytes32, Bytes
+        from ethereum.cancun.trie import InternalNode
+        from typing import Mapping, Optional, Union
+        from ethereum_rlp import Extended
+
+        ids.node_store = gen_arg(Mapping[Hash32, InternalNode], program_input["node_store"])
```
Eikix, line 81: smol nudge to rebase on NodeStore being a `Mapping[Hash32, Bytes]`

cairo/ethereum/cancun/main.cairo
```diff
@@ -51,10 +64,60 @@ func main{
...
+        right=right,
+    );
+
     let state = chain.value.state;
-    let account_diff_commitment = hash_state_account_diff(state);
-    let storage_diff_commitment = hash_state_storage_diff(state);
+    let account_diff = sort_account_diff(account_diff);
+    let storage_diff = sort_storage_diff(storage_diff);
+
+    tempvar state = state;
```
Eikix, line 103: i assume this is linked to wip stuff?

cairo/tests/utils/args_gen.py
```diff
@@ -1242,21 +1251,6 @@ def _gen_arg(
         ]
 
         segments.load_data(struct_ptr, data)
-
-        if arg_type_origin is Trie:
```
Eikix, line 1246: why are we deleting this?

enitrat: moved above to `generate_trie_arg`

cairo/ethereum/cancun/main.cairo
```diff
@@ -31,43 +46,57 @@ func main{
...
     ].value.header;
     let pre_state_root = parent_header.value.state_root;
-    let post_state_root = block.value.header.value.state_root;
-
     state_transition{chain=chain}(block);
 
-    let block_hash = keccak256_header(block.value.header);
+    // # Compute the diff between the pre and post STF MPTs to produce trie diffs.
+    let pre_state_root_bytes = Bytes32_to_Bytes(pre_state_root);
+    let pre_state_root_node = OptionalUnionInternalNodeExtendedImpl.from_bytes(
```
Eikix, line 67: nice way to link the two

cairo/ethereum/cancun/main.cairo
```diff
@@ -31,43 +46,57 @@ func main{
...
-    assert [output_ptr + 5] = block_hash.value.high;
-    // we commit to the state diffs by hashing the storage and account diffs collectively.
-    // TODO: in this program or another, assert that the account commitments
-    // generated by the state transition match the ones computed between the pre and post MPTs.
-    assert [output_ptr + 6] = account_diff_commitment;
-    assert [output_ptr + 7] = storage_diff_commitment;
+    assert [output_ptr + 2] = state_account_diff_commitment;
+    assert [output_ptr + 3] = state_storage_diff_commitment;
+    assert [output_ptr + 4] = trie_account_diff_commitment;
+    assert [output_ptr + 5] = trie_storage_diff_commitment;
```
Eikix, line 98: curious, why do we output them?

Eikix (approved):
lgtm

Eikix:
If I understand the current flow:
In trie diff, we have the luxury of choosing the diffs we write. as such,
- if the prev and new storage values are the same -> we don't write the diff
- if the prev and new acct are the same except for the storage root -> we don't write the diff

In the state diff, we don't have the luxury of choosing the diffs, they are inherited from the dict structure of cairo, so instead of not writing diffs, we skip them in the commitment phase:
- in the storage diffs, if a prev and new storage values are the same (note that we replace null ptr by U256::ZERO there) -> we don't commit to that storage diff
- in the acct diffs, if a prev and new acct values are the same (except for storage root) -> we don't commit to that account diff

enitrat:
yes exactly!

PR #1311, opened 2025-04-11
Feat/ext subnode not extension

### LINKED ISSUE #1269: feat: check extension node subnode is never an extension node

Eikix:
closes #1269 

- fixed a bug where the behaviour between `_compute_left_branch_node_diff_on_right_extension_node` didnt behave like `_compute_left_extension_node_diff_on_right_branch_node`
- fixed a bug where `check_extension_node` wasn't used in the python version of trie_diff

One thought: how can we test the `check_X` in the flow of compute_diff? we don't have well constructed MPTs that have invalid nodes

### DIFF
--- a/python/mpt/src/mpt/trie_diff.cairo
+++ b/python/mpt/src/mpt/trie_diff.cairo
@@ -398,7 +398,12 @@ func _process_account_diff{
     tempvar new_address = OptionalAddress(new address.value);
 
     _compute_diff(
-        left=left_storage_root, right=right_storage_root, path=new_path, account_address=new_address
+        left=left_storage_root,
+        right=right_storage_root,
+        parent_left=OptionalInternalNode(cast(0, InternalNodeEnum*)),
+        parent_right=OptionalInternalNode(cast(0, InternalNodeEnum*)),
+        path=new_path,
+        account_address=new_address,
     );
     return ();
 }
@@ -539,7 +544,14 @@ func compute_diff_entrypoint{
         storage_key_preimages=storage_key_preimages,
         main_trie_end=main_trie_end,
         storage_tries_end=storage_tries_end,
-    }(left=left, right=right, path=path, account_address=account_address);
+    }(
+        left=left,
+        right=right,
+        parent_left=OptionalInternalNode(cast(0, InternalNodeEnum*)),
+        parent_right=OptionalInternalNode(cast(0, InternalNodeEnum*)),
+        path=path,
+        account_address=account_address,
+    );
 
     tempvar account_diff = AccountDiff(
         new AccountDiffStruct(data=main_trie_start, len=main_trie_end - main_trie_start)
@@ -579,6 +591,8 @@ func _compute_diff{
 }(
     left: OptionalUnionInternalNodeExtended,
     right: OptionalUnionInternalNodeExtended,
+    parent_left: OptionalInternalNode,
+    parent_right: OptionalInternalNode,
     path: Bytes,
     account_address: OptionalAddress,
 ) -> () {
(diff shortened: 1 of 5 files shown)


python/mpt/src/mpt/trie_diff.py
```diff
@@ -228,7 +254,9 @@ def _compute_diff(
                             None,
                             r_node.subnodes[i],
                             path + bytes([i]),
-                            process_leaf_diff,
+                            left_parent=None,
```
Eikix, line 257: I haven't made up my mind whether `None` 👆 subnode has None parent or `left_parent`
I think `None`

python/mpt/src/mpt/utils.cairo
```diff
@@ -242,9 +243,15 @@ func check_leaf_node(path: Bytes, node: LeafNode) {
     }
 }
 
-func check_extension_node(node: ExtensionNode) {
+func check_extension_node(node: ExtensionNode, parent_node: OptionalInternalNode) {
```
enitrat, line 246: doc

python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -563,6 +575,8 @@ func compute_diff_entrypoint{
 // @implicit storage_tries_end Passed down to record storage diffs.
 // @param left The node (or reference) from the previous state's trie.
 // @param right The node (or reference) from the current state's trie.
+// @param parent_left The parent of the left node.
+// @param parent_right The parent of the right node.
```
enitrat, line 579:
```suggestion
// @param parent_left The parent of the left node. Used to ensure the trie traversed is well-formed.
// @param parent_right The parent of the right node. Used to ensure the trie traversed is well-formed.
```

python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -1430,59 +1579,102 @@ func _compute_left_branch_on_right_extension_node{
     main_trie_end: AddressAccountDiffEntry*,
     storage_tries_end: StorageDiffEntry*,
 }(
-    left: BranchNode,
-    right: ExtensionNode,
+    subnodes: Subnodes,
```
enitrat, line 1582: why changing the signature here?

Eikix: to fit the symmetrical impl. of the same logic

enitrat: i'd revert that change unless strictly necessary because it breaks the pattern of all helper functions being left node vs right node

enitrat: also is there a check_branch_node missing here?

Eikix: > i'd revert that change unless strictly necessary because it breaks the pattern of all helper functions being left node vs right node

it's the current signature of `_compute_left_extension_node_diff_on_right_branch_node`

Eikix: i'm simply making _compute_left_branch_node_diff_on_right_extension_node a symmetrical helper

Eikix: > also is there a check_branch_node missing here?

i dont think so, since `_compute_left_branch_node_diff_on_right_extension_node` is called in `_left_is_branch_node` which performs the check

python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -1430,59 +1579,102 @@ func _compute_left_branch_on_right_extension_node{
...
-                    new BytesStruct(
-                        data=right.value.key_segment.value.data + 1,
-                        len=right.value.key_segment.value.len - 1,
+        // Fully consumed by this nibble: compare to the subnode
+        if (right_extension.value.key_segment.value.len == 1) {
+            let node_to_compare_ = OptionalUnionInternalNodeExtendedImpl.from_extended(
+                right_extension.value.subnode
+            );
+            tempvar current_parent_right = right;
+            tempvar node_to_compare = node_to_compare_;
```
enitrat, line 1623: what would happen if we didn't do this?

python/mpt/src/mpt/trie_diff.cairo
```diff
@@ -1704,14 +1922,15 @@ func _compute_left_extension_node_diff_on_right_branch_node{
...
-    let l_extension = left.value.extension;
+    let l_extension = left.value.extension_node;
     let first_nib = l_extension.value.key_segment.value.data[0];
     if (first_nib == index) {
         // Fully consumed by this nibble: compare to the subnode
         if (l_extension.value.key_segment.value.len == 1) {
             let node_to_compare_ = OptionalUnionInternalNodeExtendedImpl.from_extended(
                 l_extension.value.subnode
             );
+            tempvar current_parent_left = left;
```
enitrat, line 1933: now that I think of it it would be "cleaner" to define 

```
local node_to_compare: OptionalUnionInternalNodeExtended;
local current_parent_left: OptionalInternalNode;
```

and doing `assert node_to_compare = xxx` in each branch rather than having tempvar that we need to re-bind at the end

python/mpt/src/mpt/trie_diff.cairo
enitrat: right 👍

python/mpt/src/mpt/trie_diff.cairo
Eikix: if you mean:
- checking whether the length of the key segment is exactly one? if we didn't we could shorten an extension node to have a keysegment of length 0 and that might be a problem
- if the keysegment len >= 1, we need to compare the branch subnode to the shortened extension, if key segment len == 1 we must compare the branch subnode to the extension subnode

python/mpt/src/mpt/trie_diff.cairo
Eikix: Ok!

enitrat (approved):
lgtm but this is starting to be complex and there's a lot of cases we're not testing properly 🤔

PR #1313, opened 2025-04-11
feat: implement STF & Trie Diff commitments in `main`

### LINKED ISSUE #1029: feat: modify main.cairo entrypoint to incorporate the new trie diff logic

### LINKED ISSUE #1208: feat: output hash of trie / state diffs
## Why

In our program we need to assert State Diffs == Trie Diffs,
To do that, we can iterate over keys and assert equality or hash values and compare hashes.

## What

A sequential hash strategy over keys (which are sorted) of the following form:
```
acc_0 = poseidon2((address, prev_account_0, post_account_0))
acc_i = poseidon2(acc_i-1, (address, prev_account_i, post_account_i))
```

### LINKED ISSUE #1289: bug: prove-block panic with bloc 22223896
Image

[22223896.json](https://github.com/user-attachments/files/19664993/22223896.json)

### LINKED ISSUE #1301: feat: when hashing state diff commitment, skip non-diff entries
# What 

Given the State access dictionaries for _main_trie and _storage_tries, if a prev_value is equal to new_value, it must not be a part of the commitment

# Why

The State dict logs all accesses. A read-only access would produce (key, prev, new=prev), which is not a diff, and thus must not be part of the state diff commitment

# How

Inside `_accumulate_state_diff_hashes` we want to compare `prev_value` and `new_value` that are both either pointers (relocatable == felt*) or 0.

If they're equal, we skip the call to `poseidon_account_diff` and filling the buffer at this index.

Otherwise, we `poseidon_account_diff` and add the hash to the buffer.

Because we will skip some entries in the iteration, we need to return the final len of the buffer from this function. Refactor the arguments to instead return the `buffer` end pointer, increase the `buffer: felt*` pointer on each recursive call, so that the last call returns the end pointer.

Notes:

1. We can't directly compare pointers in cairo. If they're in the same segment, we can compare them, otherwise, we cant.

Some ideas on how to compare pointers.
```
// TODO: this is not the best approach, considering that the non-equal case is hard to prove,
// and thus, we should think of a better approach.
// @notice Tries to check if two pointers are equal. This can be trusted if the pointers are equal, but
// cannot be trusted if they are not equal. Thus, use carefully.
// @dev Returns 1 if lhs and rhs point to the same memory location, 0 otherwise
//      A typical `lhs == rhs` comparison will fail if both pointers don't have the same segment index.
//      We can ask the prover to provide whether they are equal, and simply verify this result.
// @param lhs The first pointer to compare
// @param rhs The second pointer to compare
// @return 1 if pointers are equal. If we cannot attest their equality, we return 0.
func is_ptr_equal(lhs: felt*, rhs: felt*) -> bool {
    alloc_locals;

    if (cast(lhs, felt) == 0) {
        if (cast(rhs, felt) == 0) {
            let res = bool(1);
            return res;
        }
        let res = bool(0);
        return res;
    }

    tempvar segment_equal;
    %{ ids.segment_equal = (ids.lhs.segment_index == ids.rhs.segment_index) %}
    jmp segments_are_equal if segment_equal != 0, ap++;

    // We can prove the equality, but not the inequality on different segments.
    let res = bool(0);
    return res;

    segments_are_equal:
    let res_ = is_zero(lhs - rhs);
    let res = bool(res_);
    return res;
}
```

Because this cannot be trusted in case the pointers are in different segments, we would need to do value comparison if these two pointers are in different segments.

### LINKED ISSUE #1302: feat: use `prestate` instead of computed state from zkpi MPTs in `prove_block`

enitrat:
Closes #1301, Closes #1302, Closes #1208, Closes #1289, Closes #1029

# AI-REPORT: Ethereum State Diff Comparison Logic PR (April 11, 2025)

## Overview

This PR refactors the Ethereum state transition prover in Cairo, replacing final
`state_root` validation with a diff-based approach. It verifies state
modifications (account/storage diffs) from the State Transition Function (STF)
against Merkle Patricia Trie (MPT) diffs, enhancing modularity and altering the
prover's input/output contract.

## Context

The system proves Ethereum block state transitions. Previously, it validated by
recomputing the post-state root and comparing it to the block header’s
`state_root`. This PR adopts an EELS-aligned strategy, validating STF-generated
diffs (account/storage changes) against MPT diffs computed from provided
`pre_state_root` and `post_state_root`, avoiding costly full state root
recalculation.

## Changes

1. **cairo/ethereum/cancun/main.cairo (Entrypoint)**

   - **Logic Shift**: Orchestrates diff comparison:
     - Runs STF (`state_transition`) to generate account/storage diffs.
     - Computes MPT diffs (`compute_diff_entrypoint`) using `pre_state_root`
       (parent block) and input `post_state_root`.
     - Sorts diffs (`sort_account_diff`, `sort_storage_diff`).
     - Computes Poseidon hash commitments for STF diffs
       (`hash_state_account_diff`, `hash_state_storage_diff`) and MPT diffs
       (`hash_account_diff_segment`, `hash_storage_diff_segment`).
     - Asserts commitment equality
       (`state_account_diff_commitment == trie_account_diff_commitment`,
       `state_storage_diff_commitment == trie_storage_diff_commitment`).
   - **I/O Changes**:
     - **Inputs**: Added `node_store`, `address_preimages`,
       `storage_key_preimages`, `post_state_root` via `%{ main_inputs %}` hint.
     - **Outputs**: Removed `post_state_root`, `block_hash`. Now outputs
       `pre_state_root`, STF/MPT diff commitments (6 felts, down from 8).
   - **Rationale**: Validates state changes incrementally, reducing STF
     complexity by offloading root hash computation.

2. **cairo/ethereum/cancun/fork.cairo (STF)**

   - **Change**: Removed `state_root` equality check
     (`output.value.state_root == block.value.header.value.state_root`).
   - **Comment**: Notes validation now uses diff comparison, per EELS.
   - **Rationale**: Avoids redundant root computation, relying on diff-based
     correctness.

3. **cairo/ethereum/cancun/fork_types.cairo (Data Structures)**

   - **New Function**: `account_eq_without_storage_root` compares
     `OptionalAccount` instances, ignoring `storage_root`.
   - **Update**: `Account__eq__` uses this helper.
   - **Rationale**: Separates account (nonce, balance, code hash) and storage
     diffs for independent validation, as storage changes are tracked
     separately.

4. **cairo/ethereum/cancun/state.cairo (State Management)**

   - **Change**: Replaced `default_dict_finalize` with `dict_squash` in
     `finalize_state` for `main_trie` and `storage_tries`.
   - **Rationale**: Preserves pre-execution values in state tries (not
     defaulting to 0), ensuring accurate diff computation.

5. **cairo/scripts/prove_block.py (Proving Script)**

   - **Changes**:
     - Uses `ZkPi.from_data` to load inputs, including `transition_db` (MPT
       data).
     - Supplies `node_store`, preimages, `post_state_root` to Cairo
       (`program_input`).
     - Replaces `prepare_state_and_code_hashes` with `map_code_hashes_to_code`
       (no more adaptation of the State object required: it must _always_ be
       made upfront).
     - Removed Python state root recalculation (`apply_body`), as
       `post_state_root` is now an input of the program.
     - Patches Python environment: adds `is_account_alive`, `EMPTY_ACCOUNT`.
   - **Rationale**: Aligns with diff-based validation, simplifies input
     preparation, ensures Python-Cairo consistency. We want the pre-state to be
     loaded from zkpi, which contains all touched accounts / storage slots.

6. **cairo/tests/ef_tests/cancun/test_state_transition.py (Tests)**

   - **Change**: Ignores `wrongStateRoot_Cancun` test, obsolete due to
     diff-based validation;
   - **Rationale**: Tests focusing on state root mismatches are irrelevant as we
     don't recompute a new state root, we don't need to test for it: our
     diff-approach is sufficient.

7. **mpt/hash_diff.cairo (Diff Hashing)**

   - **Change**: Skips unchanged diffs in `hash_state_account_diff` and
     `hash_state_storage_diff` using `account_eq_without_storage_root` and
     `U256__eq__`. Handles null pointers as `U256::ZERO` for storage diffs.
   - **Rationale**: excluding no-diff entries from the hash computation aligns
     STF and MPT diff commitments.

8. **mpt/trie_diff.cairo (Trie Diffs)**

   - **Change**: Skips account diffs if only `storage_root` differs
     (`account_eq_without_storage_root`). Skips storage diffs if values are
     equal (`U256__eq__`).
   - **Rationale**: We don't recompute the storage root in the STF. We ensure we
     have the correct result by emitting full storage diffs.

9. **Python and Test Updates**
   - **trie_diff.py**: Filters out unchanged storage diffs
     (`left_decoded == right_decoded`). Computes commitments for testing.
   - **test_main.py**: Verifies commitments match expected values.
   - **args_gen.py**: Patches `is_account_alive` to check that it's not the
     EMPTY_ACCOUNT (with storage_root and code_hash) instead of only checking
     nonce, balance, code.
   - **Rationale**: Enhances test coverage for diff-based logic, aligns Python
     logic with Cairo.

## Patterns

- **Diff-Based Validation**: Verifies correctness via STF vs. MPT diff
  comparison, not final state root.
- **Input-Driven Post-State**: `post_state_root` is input, not computed,
  shifting validation dependency.
- **Account vs. Storage Separation**: `account_eq_without_storage_root` isolates
  account and storage diffs for independent hashing.
- **Differential Diff Handling**:
  - **MPT Diffs**: Omits unchanged slots/accounts (ignoring `storage_root`).
  - **STF Diffs**: Cairo dictionaries capture all accesses; unchanged entries
    skipped during commitment (`hash_state_account_diff`,
    `hash_state_storage_diff`).
- **Optimized Comparison**: Uses pointer comparison first for accounts, falling
  back to field-by-field if inconclusive.
- **Python Patching**: Consistent `setattr` in `prove_block.py` aligns Python
  and Cairo behaviors.

The most important pattern to keep in mind is that **defaultdicts get serialized
as default_dict** meaning that we can query absent keys, while
**non-defaultdicts** in args_gen generate regular dicts. As such one must be
very careful when writing tests: if you don't know all keys queried during the
test, you **must** use defaultdicts.

## Side Effects and Risks

- **API Change**: `main` requires MPT context inputs, outputs diff commitments.
- **Input Dependency**: Relies on external `node_store`, preimages,
  `post_state_root`. Validation critical (e.g., well-formed tries, correct node
  mappings).
- **Testing Shift**: Obsoletes state root tests; needs new diff-focused tests.
- **Performance**: Avoids STF root hashing but adds MPT diff computation and
  Poseidon hashing costs. Requires benchmarking.
- **Complexity**: MPT diff generation adds complexity, though STF logic is
  simplified.

## Decisions

- **Why Diff Comparison?**
  - **Modularity**: Decouples STF from MPT root calculation.
  - **Efficiency**: Avoids costly MPT hashing in Cairo STF.
  - **Verifiability**: Enables incremental proofs for transaction chunks.
- **Why Ignore Storage Root?** Treats account and storage changes as separate
  streams for distinct hashing.
- **Why Remove Python Recalculation?** Validation uses input `post_state_root`
  for MPT diffs, not STF recomputation.


python/cairo-addons/src/cairo_addons/hints/os.py
```diff
@@ -41,3 +41,14 @@ def block_hashes(segments: MemorySegmentManager, ids: VmConsts):
     ids.block_hashes = segments.gen_arg(
         [random.randint(0, 2**128 - 1) for _ in range(256 * 2)]
     )
+
+
+@register_hint
+def get_code_from_hash(
```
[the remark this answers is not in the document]
enitrat: good idea 👍

python/mpt/src/mpt/ethereum_tries.py
```diff
@@ -352,19 +353,92 @@ def from_data(cls, data: Dict[str, Any]) -> "EthereumTrieTransitionDB":
...
+                continue
+
+            # Initialize the account
+            pre_balance = U256(int(account["balance"][2:], 16))
+            pre_nonce = Uint(int(account["nonce"][2:], 16))
+            pre_code_hash = Hash32.fromhex(account["codeHash"][2:])
+            pre_storage_hash = Hash32.fromhex(account["storageHash"][2:])
+            pre_code = Bytes.fromhex(account["code"][2:]) if "code" in account else None
+
+            # Note: If you want to use EELS with the ZKPi-pre-state, you need to explicitly set the code to an empty bytearray if the codehash is none,
```
Eikix, line 382: why are we code to None and not bytearray()?

python/mpt/src/mpt/hash_diff.cairo
```diff
@@ -44,39 +54,38 @@ func poseidon_account_diff{poseidon_ptr: PoseidonBuiltin*}(diff: AddressAccountD
...
 
-    // Inline the new Account
-    assert buffer[offset] = diff.value.new_value.value.nonce.value;
-    assert buffer[offset + 1] = diff.value.new_value.value.balance.value.low;
-    assert buffer[offset + 2] = diff.value.new_value.value.balance.value.high;
-    assert buffer[offset + 3] = diff.value.new_value.value.code_hash.value.low;
-    assert buffer[offset + 4] = diff.value.new_value.value.code_hash.value.high;
+    // TODO: This technically only happens in cases where the account was deleted,
+    // which can only happen in case of a same-transaction selfdestruct, meaning the account was
+    // originally Null. As such: once we skip computing hashes when there's no diff, this should be removed
```
Eikix, line 59: Should we remove it now that we skip?

Eikix:
How do we assert our work is "complete"?
should we use `zkpi::extra::stateDiffs` field, and hash it in a python test to assert whether our python and cairo impls. are correct?

python/mpt/src/mpt/ethereum_tries.py
enitrat: None would be a non-cached entry, whereas bytearray() is explicitly saying "this account has for code b''"

enitrat:
> should we use zkpi::extra::stateDiffs field, and hash it in a python test in the same way our cairo hashes to assert whether our cairo impl is correct?

We could add this as an assertion in `test_main`, but if we have the same trie diffs and storage diffs, i'd say it's ok?

Eikix:
> > should we use zkpi::extra::stateDiffs field, and hash it in a python test in the same way our cairo hashes to assert whether our cairo impl is correct?
> 
> We could add this as an assertion in `test_main`, but if we have the same trie diffs and storage diffs, i'd say it's ok?

I wonder if having the same trie diffs and state diffs is ok, probably it is 99%, i simply wonder whether it's enough,
maybe we'd be safer if we knew we match the geth output of state diffs?

because our state diffs originate from our cairo impl. so we can't say they are reference

docs/AI-REPORT.md
```diff
@@ -0,0 +1,171 @@
...
+### 1. `cairo/ethereum/cancun/main.cairo` (Entrypoint Logic)
+
+- **Core Logic Shift:** This file now orchestrates the diff comparison.
+  - Executes the STF (`state_transition`) to generate internal state diffs
+    (account and storage changes).
+  - Computes MPT diffs (`compute_diff_entrypoint`) using the `pre_state_root`
+    (from the parent block) and a provided `post_state_root` input. Requires new
+    inputs: `node_store`, `address_preimages`, `storage_key_preimages`.
+  - Sorts both sets of diffs (`sort_account_diff`, `sort_storage_diff`).
+  - Computes Pedersen hash commitments for both STF-generated diffs
```
Eikix, line 40: poseidon?

Eikix:
```suggestion
  - Computes Poseidon hash commitments for both STF-generated diffs
```

cairo/tests/utils/args_gen.py
```diff
@@ -394,6 +394,22 @@ def encode_account(raw_account_data: Account, storage_root: Bytes) -> Bytes:
     )
 
 
+# TODO PR in EELS?
```
Eikix, line 397: why PR EELS?

enitrat: instead of doing a patch we could simply PR EELS and avoid that patch

cairo/tests/ethereum/cancun/test_fork_types.py
```diff
@@ -5,13 +5,15 @@
...
 from mpt.ethereum_tries import EMPTY_BYTES_HASH, EMPTY_TRIE_HASH
+from tests.utils.strategies import account_strategy
 
 
 class TestForkTypes:
     def test_account_default(self, cairo_run):
         assert EMPTY_ACCOUNT == cairo_run("EMPTY_ACCOUNT")
 
-    @given(account_a=..., account_b=...)
+    # Note using ellipsis, as hypothesis loads the strategies before loading the EELS patches.
```
Eikix, line 15:
```suggestion
    # Not using ellipsis, as hypothesis loads the strategies before loading the EELS patches.
```

cairo/tests/utils/args_gen.py
Eikix: do you mean PR EELS to add storage root and codehash in account clasS?

python/mpt/src/mpt/trie_diff.py
```diff
@@ -97,16 +99,53 @@ def from_data(cls, data: Dict[str, Any]) -> "StateDiff":
...
+                storage_diffs.append(StorageDiffEntry(key_hashed, pre, post))
+        storage_diffs = sorted(storage_diffs, key=lambda x: x.key)
+
+        account_diff_hashes = [diff.hash_poseidon() for diff in account_diffs]
+        storage_diff_hashes = [diff.hash_poseidon() for diff in storage_diffs]
+
+        account_diff_commitment = poseidon_hash_many(account_diff_hashes)
+        storage_diff_commitment = poseidon_hash_many(storage_diff_hashes)
+
+        return account_diff_commitment, storage_diff_commitment
```
Eikix, line 147: this doesn't look like it matches the function signature

cairo/tests/utils/args_gen.py
enitrat: no just do account != EMPTY_ACCOUNT

Eikix (approved):
GG ✅↗🗿

PR #1317, opened 2025-04-11
dev: optimize Bytes__eq__

### LINKED ISSUE #1316: dev: optimize `Bytes__eq__` by replacing `is_zero` with assert
When we want to verify that two bytes are equal, the prover is supposed to tell us if there's a diff. Otherwise, all is supposed to be equal. As such, instead of computing `is_zero(self, other)` we can just `assert self = other` for the current byte, which should consume a lot more steps (avoids a function call). 

Empirically, this would decrease the step usage of this function of ~30% (4.3M -> 3.3M on sample block `22188088`

enitrat:
Empirically, this decreases the step usage of this function of ~30% (4.3M -> 3.3M) on sample block 22188088, from 520 to 400 per-call.

Closes #1316

### DIFF
--- a/cairo/ethereum/utils/bytes.cairo
+++ b/cairo/ethereum/utils/bytes.cairo
@@ -71,9 +71,7 @@ func Bytes__eq__(_self: Bytes, other: Bytes) -> bool {
 
     if (is_diff == 1) {
         // Assert that the bytes are different at the first different index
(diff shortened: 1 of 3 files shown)


Eikix:
Side note but doesn't look like my codecov PR worked

Eikix (approved):
lgtm

Issue #1225, opened 2025-04-03
dev: make `patch_hint` work on RustVM runs


enitrat:
# What 
Now that we have (almost) full compatibility of python hints execution inside the rust vm, modify `patch_hint` to work for runs on the RustVM runs.

Once this works, remove all `cairo_run_py` used along with `patch_hint`

# How

The signature of `patch_hint` today is

```
@contextmanager
def patch_hint(
    programs: List[Program], hint: str, new_hint: str, scope: Optional[str] = None
):
```

It' used in tests like for example:

```
    @given(address=st.sampled_from(list(PRE_COMPILED_CONTRACTS.keys())))
    @settings(verbosity=Verbosity.quiet)
    def test_precompile_table_lookup_hint_index_out_of_bounds(
        self, cairo_programs, cairo_run_py, address
    ):
        address_int = int.from_bytes(address, "little")
        with (
            patch_hint(
                cairo_programs,
                "precompile_index_from_address",
                f"ids.index = {len(PRE_COMPILED_CONTRACTS)*3 + 1}",
            ),
            cairo_error(message="precompile_table_lookup: index out of bounds"),
        ):
            cairo_run_py("precompile_table_lookup", address_int)
```

Where `cairo_programs` is a fixture that returns the list of compiled programs associated to this test file in their __Pythonic Object__ form.

```
@pytest.fixture(scope="module")
def cairo_programs(request) -> List[Program]:
    return request.session.cairo_programs[request.node.fspath]
```

If we want to make this compatible with RustVM runs, we need to patch these hints on the RustVM Program  Object

For this we already have a fixture called `rust_programs`

```
@pytest.fixture(scope="module")
def rust_programs(cairo_programs: List[Program], python_vm: bool) -> List[RustProgram]:
    if python_vm:
        return []

    return [
        RustProgram.from_bytes(
            json.dumps(cairo_program.Schema().dump(cairo_program)).encode()
        )
        for cairo_program in cairo_programs
    ]
```

which returns a binding to the Rust `Program` object defined in  `program.rs`.

What we should do is: 

- Inside the Pyo3 binding object for `RustProgram` defined in `program.rs`, create an entrypoint `program_with_patched_hint` that given an identifier `id`(the hint code), clones the program object, searches for this identifier in the program's hints list, and replaces its code with the patched code, and then returns this cloned, modified program.

Eikix:
idea: try to hotswap fixture rust_programs by patched_rust_programs

PR #1318, opened 2025-04-11
feat: add patch_hint support on rustvm

### LINKED ISSUE #447: enable patch_hint test util with the rust VM
## Why

We want to run everything on the rust vm as it's the real prod situation and it's faster. Currently, the `patch_hint` context manager only works with the python vm though

## What

Find a way to update hints for the rust vm at runtime

## How

tbd

Eikix:
closes https://github.com/kkrt-labs/keth/issues/1225

### DIFF
--- a/python/cairo-addons/src/cairo_addons/testing/hints.py
+++ b/python/cairo-addons/src/cairo_addons/testing/hints.py
@@ -1,5 +1,4 @@
 import contextlib
-import json
 import re
 from contextlib import contextmanager
 from importlib import import_module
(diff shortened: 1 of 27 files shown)


python/cairo-addons/src/cairo_addons/testing/hints.py
```diff
@@ -28,111 +28,109 @@ def _debug_info(pc):
...
+        new_hint: The new hint code to use.
+        scope: Optional scope name to restrict which hints are patched.
 
-        # Replace a regular hint
-        with patch_hint(program, 'ids.x = 5', 'ids.x = 10'):
-            ...
+    Yields:
+        The programs with the hint patched (same list for `Program`, new list for `RustProgram`).
 
-        # Replace hint only in specific scope
```
enitrat, line 62: we could keep this no?

python/cairo-addons/src/cairo_addons/testing/hints.py
```diff
@@ -28,111 +28,109 @@ def _debug_info(pc):
...
-                            )
-                        )
-                    else:
-                        new_hints.append(hint_)
-                else:
-                    # Handle regular hint patching
-                    if hint_.code.strip() == implementations.get(
-                        hint.strip(), hint.strip()
+    if not programs:
+        raise ValueError("No programs provided")
```
enitrat, line 59: I don't think we should validate that, the arg type does not allow for None anyway

python/cairo-addons/src/cairo_addons/testing/hints.py
Eikix: ah yes thx

enitrat (commented):
nice 👍

ClementWalter (approved):

Eikix (commented):
⚠️ This PR make tests longer 
Observed on one test -> 6x longer to run

Eikix:
A couple of errors in tests seem to emerge from type differences between the Rust and Python VMs.

python/cairo-core/tests/src/cairo_core/test_comparison.py
```diff
@@ -1,10 +1,10 @@
 import pytest
 from hypothesis import given
 from starkware.cairo.lang.vm.relocatable import RelocatableValue
+from tests.utils.strategies import felt
```
ClementWalter, line 4: probably update the pyproject.toml known first party

ClementWalter (dismissed):

Issue #1319, opened 2025-04-12
bug: off-by-one dict_ptr_end when executing block `22223896`


enitrat:
When executing `22223896` the execution panics because inside a call to the `copy_tracker_to_new_ptr` hint, we provide as dict_ptr_end ` 26926:159` instead of `26926:162`. This likely means that at some point, we're doing a dict_write / dict_read that's not properly rebinded.

PR #1320, opened 2025-04-13
fix: state rebinding issue after get_account_code

enitrat:
Closes https://github.com/kkrt-labs/keth/issues/1319

The issue was that we were calling `get_account_code`, mutating the `state`, but not properly re-binding the `state` to the `environment`.  Shifting the lines orders fixes that.


Eikix (approved):
lgtm

PR #1328, opened 2025-04-14
feat: centralize patch system

### LINKED ISSUE #1233: dev: design an implement a coherent and documented strategy for our codebase patches
we're accumulating patches in `prove_block.py` and `conftest.py`, with definitions in `tests.utils.args_gen`. These definitions and patches feel a bit out of place and duplicated. We should group and properly document these.

The biggest issue is that it involves not only patching the items definitions in their respective modules, but also hot-patching them wherever they're called. We need to have a better understanding and management of python's module loading system to avoid this - and ideally only patch where items are defined.

enitrat:
Closes #1233
### Why Custom Types and Patching?

Keth's Cairo-based prover requires custom types to ensure compatibility with the
Ethereum EELS library. EELS types are incompatible with Cairo's data structures
and serialization needs for zero-knowledge proof generation. Keth's custom types
(e.g., `Account`, `Evm`, `Message`) override EELS defaults, adjusting equality
comparisons (e.g., ignoring `storage_root` in `Account`) and serialization to
match Cairo's expectations.

Patching replaces EELS's original types with Keth's custom types across the
Python codebase. This prevents inconsistencies between Python logic (for input
prep and testing) and Cairo logic (for proving).

### How Patching Works

Patching is centralized in `keth_types.patches`, which defines a `PATCHES`
dictionary mapping attribute names to Keth's custom types. Importing
`keth_types.patches` triggers `apply_patches()`, which iterates through all
sub-modules of the `ethereum` package and `mpt` modules in `sys.modules`, using
`setattr` to override original definitions with custom types. This approach
ensures that patches are applied automatically to all relevant modules,
regardless of loading order. If we had only patched in the module where items
are defined, we'd get some errors, as these modules were already loaded in
python's module system, and the patches would not be effective everywhere.

### Challenges with Python Module Loading

Python's module loading order poses a challenge. Modules are cached in
`sys.modules`, and if EELS modules (e.g., `ethereum.cancun.vm`) load before
patches, original types are used instead of Keth's. This did happen because of
pytest plugins that initialized earlier than `conftest.py` hooks.

### Proper Patching Methodology

To ensure consistent patch application, follow these guidelines:

- **Early Import**: Import `keth_types.patches` at the start of every entry
  point (e.g., `prove_block.py`, `conftest.py`) before EELS modules to avoid
  unpatched cached modules.
- **Custom Pytest Plugin (if needed)**: Because pytest plugins load EELS modules
  early, we created a plugin (see `pyproject.toml`) to apply patches during
  plugin init, ensuring it runs before, for example, hypothesis strategies.
- **Centralized Type Definitions**: Define all custom types in
  `keth_types.types` as a single source of truth for imports.

These practices ensure consistent use of custom types, preventing discrepancies
between Python and Cairo logic. Maintainers should audit entry points for patch
imports and extend `PATCHES` in `keth_types.patches` for new attributes or
types.


cairo/scripts/prove_block.py
```diff
@@ -3,88 +3,39 @@
 Fetches zkpi data, converts it to EELS/Keth format, and generates a proof.
 """
 
+# Apply EELS types patches before importing any other modules
+from keth_types.patches import apply_patches
+
+apply_patches()
+
```
ClementWalter, line 10: can put this in `__init__.py` instead

cairo/scripts/prove_block.py
```diff
@@ -3,88 +3,39 @@
...
-from ethereum.cancun.transactions import LegacyTransaction  # noqa
-from ethereum.cancun.vm.gas import calculate_excess_blob_gas  # noqa
-from ethereum.utils.hexadecimal import hex_to_bytes, hex_to_u256, hex_to_uint  # noqa
-from ethereum_spec_tools.evm_tools.loaders.fixture_loader import Load  # noqa
-from ethereum_types.bytes import Bytes, Bytes0, Bytes32  # noqa
-from ethereum_types.numeric import U64, U256  # noqa
-
-from cairo_addons.vm import run_proof_mode  # noqa
-from tests.ef_tests.helpers.load_state_tests import (  # noqa
+from tests.ef_tests.helpers.load_state_tests import (
```
ClementWalter, line 35: we should not import from `tests` in a prod script

ClementWalter (approved):

PR #1335, opened 2025-04-15
feat: U384_to_le_bytes

### LINKED ISSUE #1334: refactor: U384_to_le_bytes
# What
Create U384_to_le_bytes utils

# How

```
 let bytes_be = U384_to_be_bytes(value, length);
 let (local bytes_result: felt*) = alloc();
 reverse(bytes_result, length, bytes_be.value.data);
```

obatirou:
Closes #1334

### DIFF
--- a/cairo/ethereum/utils/numeric.cairo
+++ b/cairo/ethereum/utils/numeric.cairo
@@ -22,6 +22,7 @@ from cairo_core.maths import (
 )
 from cairo_core.comparison import is_zero
 from cairo_ec.uint384 import uint256_to_uint384
+from legacy.utils.array import reverse
 from legacy.utils.bytes import bytes_to_felt, bytes_to_felt_le, uint256_from_bytes_be, felt_to_bytes
 from legacy.utils.uint256 import uint256_add, uint256_sub
 from legacy.utils.utils import Helpers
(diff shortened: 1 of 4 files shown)


cairo/ethereum/utils/numeric.cairo
```diff
@@ -535,6 +536,17 @@ func U384_to_be_bytes{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
     return result;
 }
 
+func U384_to_le_bytes{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
+    value: U384, length: felt
+) -> Bytes {
+    alloc_locals;
+    let bytes_be = U384_to_be_bytes(value, length);
+    let (local bytes_result: felt*) = alloc();
+    reverse(bytes_result, length, bytes_be.value.data);
```
enitrat, line 545: cant we avoid a reverse here?

obatirou: As it leverages `U384_to_be_bytes` the reverse is necessary. 
It would be possible to duplicate most of the code from `U384_to_be_bytes` and adapt it for le though.

enitrat: > It would be possible to duplicate most of the code from U384_to_be_bytes and adapt it for le though.

That's preferable

enitrat: It's not really duplicating as you'd be iterating over the limbs in the reversed order directly. Reversing the array here is super inefficient when we can directly work in the correct endianess

obatirou: I was wrong on this indeed. Made the change with [0ebf1d6](https://github.com/kkrt-labs/keth/pull/1335/commits/0ebf1d69e7d86734b70ff537c7b5fb47c8e8d1cf)

cairo/tests/ethereum/utils/test_numeric.py
```diff
@@ -240,70 +135,210 @@ def test_Bytes32_from_be_bytes(self, cairo_run, bytes: Bytes):
...
+        @given(value=...)
+        def test_U384_to_le_bytes(self, cairo_run, value: U384):
+            int_value = value._number
+            if int_value == 0:
+                length = 1  # At least one byte for zero
+            else:
+                length = (int_value.bit_length() + 7) // 8
+
+            cairo_result = cairo_run("U384_to_le_bytes", value, length)
+
```
enitrat, line 341: In the test, length should be a parameter, not always strictly equal to the number of bytes required to read the word (as in the cairo function, length is an arbitrary number). What should be the behavior is length < number of bytes in the word?

obatirou: you are raising a good point. In the context of U384_to_le_bytes, it is always used for a length of 48bytes. So I think it should be refactored: change the name `U384_to_le_48_bytes` and remove the length parameter.

enitrat: Ok for now but generally I think we should make our type libraries generic

enitrat (approved):

PR #1361, opened 2025-04-17
fix: check minimal bytes U384_to_be_bytes

### LINKED ISSUE #1341: dev: U384_to_be_bytes should be
# What
U384_to_be_bytes do not check if length is < of number of bytes in the U384. This should be fix

# How
Calculate the minimum number of bytes to represent the U384. felt252_bit_length can be used. Assert is_le(min_bytes, length)

obatirou:
Closes #1341


cairo/ethereum/utils/numeric.cairo
```diff
@@ -536,6 +536,78 @@ func U384_to_be_bytes{range_check_ptr, bitwise_ptr: BitwiseBuiltin*}(
...
+        return 12 + d1_bytes;
+    }
+
+    let d0_bytes = calc_limb_bytes(value.value.d0);
+    return d0_bytes;
+}
+
+// Each limb is 96 bits (12 bytes)
+// Use binary search
+func calc_limb_bytes{range_check_ptr}(limb: felt) -> felt {
```
enitrat, line 563: Improvement proposals

1. There's an existing `Helpers.bytes_used_128` (`legacy.utils.utils`) function that is returning how many bytes are used in a felt
2. That function is seemingly legacy, as it's just doing dumb divisions until the quotient is 0
3. I remember us working on a similar issue with `felt252_bit_length` a while back - we can easily use non-determinism for this task

So what we can do is

- Create `felt252_bytes_length` in `cairo_core.maths`
- Inside `felt252_bytes_length` we simply get the amount of bytes in the felt in a hint and then verify that `pow256(bytes_len - 1) < value < pow256(bytes_len)`
- Remove `Helpers.bytes_used_128` and all its usages, replaced by `felt252_bytes_length`

This is typically the kind of operations that benefits a lot from non-determinism, try to leverage it as much as possible

obatirou: Indeed, non det should be used with felt252_bytes_length

python/cairo-core/tests/src/cairo_core/test_maths.py
```diff
@@ -182,11 +182,16 @@ def test_felt252_to_bytes_be_should_panic_on_wrong_output_noncanonical(
...
 
-    @given(value=st.integers(min_value=0, max_value=DEFAULT_PRIME - 1))
+    @given(value=felt)
     def test_felt252_bit_length(self, cairo_run, value):
-        res = cairo_run("felt252_bit_length", value=value)
+        res = cairo_run("felt252_bit_length", value)
         assert res == value.bit_length()
 
+    @given(value=felt)
+    def test_felt252_bytes_length(self, cairo_run, value):
```
enitrat, line 191: can you add an `@example(value=PRIME)` to ensure coverage of the bytes_length == 32 case?

enitrat: or anything that would trigger proper coverage

cairo/ethereum/utils/numeric.cairo
```diff
@@ -20,6 +20,7 @@ from cairo_core.maths import (
     felt252_bit_length,
     felt252_to_bits_rev,
     felt252_to_bytes_le,
+    felt252_bytes_length,
```
enitrat, line 23: now that we have it it's a good time to remove `test_should_return_bytes_used_in_128_word` (py), `test__bytes_used_128` (cairo) and in https://github.com/kkrt-labs/keth/blob/39cab2eca4f7ab8913e18cad1a41ee0c446de8c6/cairo/ethereum/cancun/vm/instructions/arithmetic.cairo#L549-L560 use this new `felt252_bytes_length` function

enitrat (approved):

PR #1365, opened 2025-04-17
feat: good keth cli

### LINKED ISSUE #1340: dev: make CLI more modular
The current CLI is called `prove-block`. We would want to release a single CLI tool that enables:

1. Generating the trace, independent of the proving step
2. Proving a trace given as input
3. Verifying a proof given as input

And any combination of the above three commands.

We could name it `keth` - `uv run keth -b <block_number> --generate-trace --generate-proof --verify-proof`

enitrat:
Closes #1340

Make the Keth CLI better

* different modes (trace gen, prove, verify, e2e)
image
* Reorganize our rust bindings for clarity (separate pure stwo bindings with runner logic)
* some reorg of rust code


crates/cairo-addons/src/vm/runner.rs
```diff
@@ -748,104 +743,55 @@ except Exception as e:
...
-        }
-        return Ok(());
-    }
-
-    // Create a performance tracing span for output writing
-    let output_span = tracing::span!(tracing::Level::INFO, "write_outputs");
-    let _output_span_guard = output_span.enter();
+    // Write prover input infos
+    std::fs::write(
+        output_dir.join("prover_input_infos.json"),
```
Eikix, line 751: will this always overwrite the last proof?

cairo/scripts/keth.py
```diff
@@ -0,0 +1,241 @@
...
+def prove(
+    prover_inputs_path: Path = typer.Option(
+        ...,
+        help="Path to prover inputs (prover_input_infos.json)",
+        exists=True,
+        dir_okay=False,
+        file_okay=True,
+    ),
+    proof_path: Path = typer.Option(
+        Path("output/proof.json"),
```
Eikix, line 118: will this always overwrite the last proof?

Eikix: i wonder if we should have proof_<block_number>?
maybe later with feedback from Nicolas

crates/cairo-addons/src/vm/runner.rs
enitrat: fixd

Eikix:
great job!! this is really cool

crates/cairo-addons/src/vm/runner.rs
```diff
@@ -748,104 +743,55 @@ except Exception as e:
...
 
-    tracing::info!("Writing air public input to {}", air_public_input.display());
-    let json = cairo_runner
-        .get_air_public_input()
-        .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?
-        .serialize_json()
-        .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
-    std::fs::write(&air_public_input, json)
-        .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+    let mut hint_processor = HintProcessor::default().with_dynamic_python_hints(false).build();
```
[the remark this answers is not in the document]
enitrat: check the function signature: it's whether we enable traces or not

Eikix:
How can we test this in the CI, so that we don't merge something that could break the CLI? We've broken the CLI many times before and i'm not sure test_main covers it 100%?

cairo/scripts/keth.py
```diff
@@ -0,0 +1,237 @@
...
+)
+logger = logging.getLogger("keth")
+console = Console()
+
+app = typer.Typer(
+    help="Keth - Generate execution traces and proofs for Ethereum blocks",
+    no_args_is_help=True,
+)
+
+CANCUN_FORK_BLOCK = 19426587  # First Cancun block
```
ClementWalter, line 39: dup from `fixture_loader`

PR #1374, opened 2025-04-18
feat: mpt test builder

### LINKED ISSUE #1314: dev: create a way to fuzz MPT inputs for `_compute_diff`
## Why

## What

## How

TBD, exploratory issue

Eikix:
closes #1314

### DIFF
--- a/python/mpt/tests/src/mpt/test_trie_diff.py
+++ b/python/mpt/tests/src/mpt/test_trie_diff.py
@@ -9,6 +9,7 @@ from ethereum.cancun.trie import (
     ExtensionNode,
     InternalNode,
     LeafNode,
+    bytes_to_nibble_list,
     encode_internal_node,
 )
 from ethereum.crypto.hash import Hash32, keccak256
@@ -22,7 +23,9 @@ from hypothesis.strategies import composite
 
 from cairo_addons.rust_bindings.vm import poseidon_hash_many
 from cairo_addons.utils.uint256 import int_to_uint256
-from mpt.ethereum_tries import EthereumTrieTransitionDB
+from keth_types.types import EMPTY_TRIE_HASH, encode_account
+from mpt.ethereum_tries import EthereumTries, EthereumTrieTransitionDB
+from mpt.trie_builder import TrieTestBuilder
 from mpt.trie_diff import StateDiff, resolve
 from mpt.utils import decode_node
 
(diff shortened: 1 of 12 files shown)


python/mpt/tests/src/mpt/test_trie_diff.py
```diff
@@ -62,6 +65,94 @@ def embedded_node_strategy(draw):
     }
 
 
+@composite
+def invalid_mpt_strategy(draw):
+    """
```
enitrat, line 70: I think we should rather split this into multiple strategies and have one test per strategy rather than rely on hypothesis to auto-generate all cases we want to cover.

Also, after we split in multiple test, we can target a specific error the test shuold raise in the pytest.raise section of the test.

Eikix: how important is it?

enitrat: quite important imo as this is typically the kind of strategy where we expect each case to be covered explicitly and raise its own error message

Eikix: > quite important imo as this is typically the kind of strategy where we expect each case to be covered explicitly and raise its own error message

made a proposal

python/mpt/tests/src/mpt/test_trie_diff.py
```diff
@@ -139,6 +227,44 @@ def test_trie_diff(
...
+        empty_ethereum_tries,
+        data,
+        invalid_case,
+    ):
+        invalid_post_trie = data.draw(invalid_mpt_strategy(invalid_case=invalid_case))
+
+        trie = EthereumTrieTransitionDB.from_pre_and_post_tries(
+            empty_ethereum_tries, invalid_post_trie
+        )
+        with pytest.raises(Exception) as cairo_error:
```
enitrat, line 254: match the error string

Eikix: the cairo error does not return a string, just ValueError when the check fails

enitrat: I'd rather have us return explicit error strings in this case as we want to make sure we're targetting this exact error. In the future we could interpret the string `ValueError: error_cause` as `ValueError("error_cause")` for these kinds of things

Eikix: how do we proceed?
we replace 
```
    if (bytes_variant.len != 0) {
        raise('ValueError');
    }
```
for instance by
```
with_attr error_message("Non empty bytes value") {
    assert 0 = 1;
}
```
?

enitrat: An idea would be
```
with_attr error_message("ValueError: Non empty bytes value") {
    assert_not_zero(bytes_variant.len)
}
```

Ideally even `raise('ValueError: Non empty bytes value')` but it wont fit in a short string. Or we use error codes.

enitrat: For the time being we could just `raise('NonEmptyBytesValue')`

Eikix: working on a fix

PR #1385, opened 2025-04-21
fix: remove unused revm dependency

reject-i:
Removed unused revm dependency from cairo_addons crate that was detected by cargo-udeps. The revm-precompile library is used in the project, but revm itself is not.


enitrat:
@reject-i  we should be importing some items from `revm::primitives` instead of using re-exports from `revm_precompiles`, thus we need the `revm` dependency

PR #1393, opened 2025-04-24
dev: support auto-generation of empty segments in runner

enitrat:
generate empty segments (useful for keccak_ptr: felt* or blake2s_ptr: felt*) when running `cairo_run`


ClementWalter (approved):
looks ok but failing

enitrat:
> looks ok but failing

still dont know what could cause a RAM leak with these changes...

PR #1395, opened 2025-04-24
Feat/invalid mpts

### LINKED ISSUE #1314: dev: create a way to fuzz MPT inputs for `_compute_diff`

Eikix:
closes #1314


python/cairo-addons/src/cairo_addons/testing/errors.py
```diff
@@ -65,25 +65,45 @@ def map_to_python_exception(e: Exception):
...
+                getattr(eth_exceptions, new_error_type, None)
+                or getattr(rlp_exceptions, new_error_type, None),
+            )
+        except Exception:
+            # Get the exception class from python's builtins or ethereum's exceptions
+            exception_class = __builtins__.get(
+                error_type,
+                getattr(eth_exceptions, error_type, None)
+                or getattr(rlp_exceptions, error_type, None),
+            )
```
enitrat, line 97: please improve this code

enitrat:
```
def map_to_python_exception(e: Exception) -> None:
    """
    Maps a generic Exception (potentially from Cairo) to a specific Python exception
    based on its string representation.

    The error string is expected to follow one of these patterns:
    1. An arbitrary error string.
    2. The name of a Python exception class (e.g., "ValueError").
    3. The name of a Python exception class followed by a message (e.g., "ValueError: error message").

    Args:
        e: The exception instance caught.

    Raises:
        A specific Python exception (if mapped and is an Exception subclass),
        AssertionError for specific Cairo assertion messages, or a generic Exception,
        preserving the original exception as the cause.
    """
    original_error_str = str(e)

    # Extract the core message, handling the "Error message: " prefix if present
    match = re.search(r"Error message: (.*)", original_error_str)
    error_content = match.group(1) if match else original_error_str

    parts = error_content.split(": ", 1)
    potential_type_name = parts[0]
    try:
        potential_type_name_int = int(potential_type_name)
        potential_type_name = (
            potential_type_name_int.to_bytes(31, "big").lstrip(b"\x00").decode()
        )
    except (ValueError, UnicodeDecodeError):
        pass

    message = parts[1] if len(parts) > 1 else None
    if message is not None:
        try:
            message_int = int(message)
            message = message_int.to_bytes(31, "big").lstrip(b"\x00").decode()
        except (ValueError, UnicodeDecodeError):
            pass

    exception_class = _find_exception_class(potential_type_name)

    if (
        exception_class
        and isinstance(exception_class, type)
        and issubclass(exception_class, Exception)
    ):
        if message:
            raise exception_class(message) from e
        else:
            raise exception_class() from e

    if (
        "An ASSERT_EQ instruction failed" in original_error_str
        or "AssertionError" in error_content
    ):
        raise AssertionError(error_content) from e

    raise Exception(error_content) from e


def _find_exception_class(name: str) -> Optional[Type[Exception]]:
    """Looks up an exception class by name in predefined modules."""
    import ethereum.exceptions as eth_exceptions
    import ethereum_rlp.exceptions as rlp_exceptions

    builtin_exc = __builtins__.get(name)
    if isinstance(builtin_exc, type) and issubclass(builtin_exc, BaseException):
        return builtin_exc

    for module in [eth_exceptions, rlp_exceptions]:
        custom_exc = getattr(module, name, None)
        if isinstance(custom_exc, type) and issubclass(custom_exc, Exception):
            return custom_exc

    return None

```

PR #1398, opened 2025-04-25
chore: update stwo

enitrat:
note: updating stwo now causes an issue with:

```
thread '<unnamed>' panicked at /Users/msaug/.cargo/git/checkouts/stwo-cairo-ef02e8e85a2fe399/68c22a3/stwo_cairo_prover/crates/adapter/src/memory.rs:283:37:
Cannot convert F252 to u128
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


crates/cairo-addons/src/vm/runner.rs
```diff
@@ -380,6 +383,27 @@ except Exception as e:
         self.inner
             .end_run(false, false, &mut hint_processor)
             .map_err(|e| PyErr::new::<pyo3::exceptions::PyRuntimeError, _>(e.to_string()))?;
+
+        if self.inner.is_proof_mode() {
```
Eikix, line 387: what is this for?
perhaps we need a lil comment

Eikix (approved):
gg!

PR #1399, opened 2025-04-25
tests: invalid mpt

### LINKED ISSUE #1314: dev: create a way to fuzz MPT inputs for `_compute_diff`

Eikix:
closes #1314


enitrat (dismissed):

enitrat (approved):

Eikix (commented):
lgtm (can't approve)

PR #1403, opened 2025-04-28
feat: blake2s hash commitment utils

### LINKED ISSUE #1366: feat: commitment utils for `_apply_body_inner`
## Why

We need to commit to `_apply_body_inner` parameters. 

## How
Global strategy: for each type in _apply_body_inner parameters, we need to have a hash util. We will hash arguments in the same order as the function signature except `index` and `len`.

```
func _apply_body_inner{
    state: State,
    transactions_trie: TrieBytesOptionalUnionBytesLegacyTransaction,
    receipts_trie: TrieBytesOptionalUnionBytesReceipt,
    blob_gas_used: Uint, 
    gas_available: Uint, 
    block_logs: TupleLog,
    range_check_ptr,
    bitwise_ptr: BitwiseBuiltin*,
    keccak_ptr: KeccakBuiltin*,
    poseidon_ptr: PoseidonBuiltin*,
    range_check96_ptr: felt*,
    add_mod_ptr: ModBuiltin*,
    mul_mod_ptr: ModBuiltin*,
}(
    index: felt,
    len: felt,
    transactions: TupleUnionBytesLegacyTransaction,
    gas_available: Uint,
    chain_id: U64,
    base_fee_per_gas: Uint,
    excess_blob_gas: U64,
    block_logs: TupleLog,
    block_hashes: ListHash32,
    coinbase: Address,
    block_number: Uint,
    block_gas_limit: Uint,
    block_time: U256,
    prev_randao: Bytes32,
    blob_gas_used: Uint,
) -> () {
```

For this, we use the modified `root` util with the `blake2s` hash function for State, transaction_trie, receipts_trie.

- [x] for `transaction_trie` and `receipts_trie`, create a new root util that uses blake2s instead of keccak

- [ ] pass the following parameters `blob_gas_used: Uint, gas_available: Uint, block_logs: TupleLog` as implicit arguments
- [ ] for LogStruct, hash each struct arg, and then hash them together.
- [ ] for` transactions: TupleUnionBytesLegacyTransaction` hash each members of the tuple, then the entire tuple.
- [ ] for `gas_available`, don't hash single felts
etc...

### RULES:
- For nested structures or Tuples, hash attributes and then hash the entire struct
- For single felt structs, do not hash

enitrat:
Adds __hash__ functions using blake2s.

Closes #1366

### DIFF
--- a/cairo/tests/utils/hash_utils.py
+++ b/cairo/tests/utils/hash_utils.py
@@ -1,0 +1,77 @@
+import hashlib
+from typing import List, Tuple, Union
+
+from ethereum.cancun.blocks import Log
+from ethereum.cancun.transactions import LegacyTransaction
+from ethereum.crypto.hash import Hash32
+from ethereum_types.bytes import Bytes, Bytes32
+
+
+def LegacyTransaction__hash__(tx: LegacyTransaction) -> Hash32:
+    nonce_hash = hashlib.blake2s(tx.nonce.to_bytes(32, "little")).digest()
+    gas_price_hash = hashlib.blake2s(tx.gas_price.to_bytes(32, "little")).digest()
+    gas_hash = hashlib.blake2s(tx.gas.to_bytes(32, "little")).digest()
+    to_hash = hashlib.blake2s(tx.to).digest()
+    value_hash = hashlib.blake2s(tx.value.to_bytes(32, "little")).digest()
+    data_hash = hashlib.blake2s(tx.data).digest()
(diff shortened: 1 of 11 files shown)


cairo/tests/utils/hash_utils.py
ClementWalter, line 1: we can also monkey patch the existing types to implement the hash __magicmethod__ and consequently just uses `hash(obj)` everywhere

enitrat: I was thinking about this too and came to the same conclusion about composite types 🤔

PR #1405, opened 2025-04-28
feat: create init.cairo

### LINKED ISSUE #1367: feat: create init.cairo file
Depends on #1366 

## Why

C.f. parent

## How
Inline all the code from `main.cairo` until the first call to `_apply_body_inner`.
Note: Do not use the code related to trie diffs.

- [ ] Instead of calling `_apply_body_inner`, commit to `_apply_body_inner` as per #1366 input arguments and add them to the `output_ptr`. 
- [ ] We also need to commit to the inputs of the `teardown` program that are not used in `_apply_body_inner` - which are `(block.header, block.withdrawals)`
- [ ] Add to `compile_keth`, the compilation of `init.cairo` output to `init_compiled.cairo`.
- At the end of `init.cairo`, verify that all dicts are properly squashed! (use `verify_squashed_dicts=True` from `cairo_run` in tests to verify this)
- [ ] Add to CLI `uv run keth` TBD
   - [ ] `uv run keth trace --step init`, `uv run e2e --step init` -> Add a `step` argument that controls what step is ran.

### LINKED ISSUE #1387: dev: adapt default settings in `uv run keth` for zkpig 0.5
Reporting from zkpig:
```
:tada: Announcing zk-pig v0.5.0 :tada:
This is a minor release
refactoring some internal code for future deployment as AWS Lambda
maintaining dependencies
:warning: Note that this release introduce some breaking changes
Change the path at which prover inputs and preflight data is stored:
Prover Inputs: /<data-dir>/<chain-id>/inputs/<block-number>.<extension> (e.g. ./data/1/inputs/2224534.json if default config)
Preflight Data: /<data-dir>/<chain-id>/preflight/<block-number>.<extension> (e.g. ./data/1/preflight/2224534.json if default config)
Stop storing preflight data by default. To store preflight you need to set the --preflight-enabled or PREFLIGHT_ENABLED=true when running zkpig commands
See [Full Release Note](https://github.com/kkrt-labs/zk-pig/releases/tag/v0.5.0)
You can run brew upgrade zkpig (edited)

```
This means we need to change our default settings for data-dir in the repo

Eikix:
closes #1367 
closes #1387

### DIFF
--- a/cairo/ethereum/cancun/init.cairo
+++ b/cairo/ethereum/cancun/init.cairo
@@ -259,8 +259,8 @@ func body_commitments{
     // Commit to the state
     let state_commitment = state_root(state, 'blake2s');
 
-    // Commit to the transaction, withdrawal and receipt tries
-    // Squash the receipts, transactions, and withdrawals dicts once they're no longer being modified.
+    // Commit to the transaction and receipt tries
+    // Squash the receipts and transactions dicts once they're no longer being modified.
     default_dict_finalize(
         cast(transactions_trie.value._data.value.dict_ptr_start, DictAccess*),
         cast(transactions_trie.value._data.value.dict_ptr, DictAccess*),
--- a/cairo/ethereum/cancun/init.cairo
+++ b/cairo/ethereum/cancun/init.cairo
@@ -194,29 +194,6 @@ func main{
         block, chain.value.chain_id, excess_blob_gas, block_hashes
     );
 
-    // Commit to _apply_body_inner arguments instead of calling it
-    //
-    // func _apply_body_inner{
-    //     state: State,
-    //     transactions_trie: TrieBytesOptionalUnionBytesLegacyTransaction,
-    //     receipts_trie: TrieBytesOptionalUnionBytesReceipt,
-    // }(
-    //     index: felt, <- we don't commit to index
-    //     len: felt, <- we don't commit to len
-    //     transactions: TupleUnionBytesLegacyTransaction,
-    //     gas_available: Uint,
-    //     chain_id: U64,
-    //     base_fee_per_gas: Uint,
-    //     excess_blob_gas: U64,
-    //     block_logs: TupleLog,
-    //     block_hashes: ListHash32,
-    //     coinbase: Address,
-    //     block_number: Uint,
-    //     block_gas_limit: Uint,
-    //     block_time: U256,
-    //     prev_randao: Bytes32,
-    //     blob_gas_used: Uint,
-    // )
 
     // Finalize the state, getting unique keys for main and storage tries
     finalize_state{state=state}();
--- a/cairo/ethereum/cancun/init.cairo
+++ b/cairo/ethereum/cancun/init.cairo
@@ -11,7 +11,7 @@ from starkware.cairo.common.cairo_builtins import (
     SignatureBuiltin,
     EcOpBuiltin,
 )
-
+from starkware.cairo.common.dict_access import DictAccess
 from starkware.cairo.common.default_dict import default_dict_new
 from starkware.cairo.common.registers import get_fp_and_pc, get_label_location
 from starkware.cairo.common.cairo_keccak.keccak import finalize_keccak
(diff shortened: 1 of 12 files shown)


cairo/ethereum/cancun/init.cairo
```diff
@@ -0,0 +1,427 @@
...
+            parent_dict=cast(0, MappingBytesOptionalUnionBytesWithdrawalStruct*),
+        ),
+    );
+    tempvar withdrawals_trie = TrieBytesOptionalUnionBytesWithdrawal(
+        new TrieBytesOptionalUnionBytesWithdrawalStruct(
+            secured=bool(0),
+            default=OptionalUnionBytesWithdrawal(cast(0, UnionBytesWithdrawalEnum*)),
+            _data=withdrawals_trie_data,
+        ),
+    );
```
enitrat, line 237: extract in a function `init_tries`

cairo/scripts/keth.py
```diff
@@ -174,25 +212,32 @@ def e2e(
...
+        help="Step to run: 'main' or 'init'",
+    ),
+    compiled_program: Path = typer.Option(
+        None,
+        help="Path to compiled Cairo program",
+        exists=True,
+        dir_okay=False,
+        file_okay=True,
+        callback=program_path_callback,
+    ),
```
enitrat, line 240: i'd still use `main` as the default program? and overload only if `step` is active?

cairo/ethereum/cancun/init.cairo
```diff
@@ -0,0 +1,345 @@
...
+    //     excess_blob_gas: U64,
+    //     block_logs: TupleLog,
+    //     block_hashes: ListHash32,
+    //     coinbase: Address,
+    //     block_number: Uint,
+    //     block_gas_limit: Uint,
+    //     block_time: U256,
+    //     prev_randao: Bytes32,
+    //     blob_gas_used: Uint,
+    // )
```
enitrat, line 219:
```suggestion
```

cairo/scripts/keth.py
enitrat: nevermind

cairo/ethereum/cancun/init.cairo
```diff
@@ -0,0 +1,345 @@
...
+        block_logs,
+        block_hashes,
+        gas_available,
+        chain.value.chain_id,
+        block.value.header.value.base_fee_per_gas,
+        excess_blob_gas,
+    );
+
+    assert [output_ptr] = init_commitment.value.low;
+    assert [output_ptr + 1] = init_commitment.value.high;
```
enitrat, line 243: let me know what you think about it but there should actually be two different kinds of commitments:

1. Commitments used as inputs for the next `body` step
2. Commitments used for the final `teardown` step

Why? Because:
1. `body` does not take as input certain values used in the commitments, such as the withdrawals trie for example
2. `teardown` will need to take some commitment value directly from the `init`,  not from the `body` - for example, the withdrawal trie

So there should be two different functions:

- `commitments_for_body`
- `commitments_for_teardown`

where:

in `commitments_for_body` we have all the inputs to the `body.cairo` part:

```
    local block: Block;
    local state: State;
    local transactions_trie: TrieBytesOptionalUnionBytesLegacyTransaction;
    local receipts_trie: TrieBytesOptionalUnionBytesReceipt;
    local block_logs: TupleLog;
    local block_hashes: ListHash32;
    local gas_available: Uint;
    local chain_id: U64;
    local base_fee_per_gas: Uint;
    local blob_gas_used: Uint;
    local excess_blob_gas: U64;
```

(im not sure whether the entire "block" should be commited at once, or just the `transactions` and the `header` separately. Perhaps the latter is better.

in `commitments_for_teardown` we have all inputs for the `teardown.cairo` part that are _not_ coming from the body part - ommers, withdrawals, etc

cairo/ethereum/cancun/init.cairo
```diff
@@ -0,0 +1,342 @@
...
+    gas_available: Uint,
+    chain_id: U64,
+    excess_blob_gas: U64,
+) -> Hash32 {
+    alloc_locals;
+    // Commit to the state
+    let state_commitment = state_root(state, 'blake2s');
+
+    // Commit to the transaction, withdrawal and receipt tries
+    // Squash the receipts, transactions, and withdrawals dicts once they're no longer being modified.
```
enitrat, line 263:
```suggestion
    // Squash the receipts and transactions dicts once they're no longer being modified.
```

enitrat (approved):

Issue #1368, opened 2025-04-17
feat: create body.cairo program


enitrat:
- Create a helper function that enables us to run EELS on the slice [0, i[ of transaction of this block to fill the input of this program
- Add in the output_ptr the commitment of the function args
- Create a new Cairo program that loads these inputs, and runs the `apply_body_inner` function from `i` to `index+len` 
- Add in the output_ptr the commitment from the updated args (see #1366), `index`, `len`.
- At the end of `apply_body_subset`, verify that all dicts are properly squashed! (use `verify_squashed_dicts=True` from `cairo_run` in tests to verify this)
- Add a step to CLI `uv run keth {trace|e2e} --step body --index <index> --len <len>`

gabito1451:
hi @enitrat may i handle this ???

PR #1408, opened 2025-04-29
feat: body program

enitrat:
Closes https://github.com/kkrt-labs/keth/issues/1368


Eikix (commented):
body programs need a pre state

yes that pre state is the post state from the n-1 body program run, but we don't output it along since all body programs are parallel

i think we discussed that body programs would all take as input the pre state from the start of the block, and perhaps inside the initialization hint of body, the EELS in hint would advance the prestate to the correct index

what is a bit unlucky is that body needs to be passed partial receipt and tx tries, which suggests the hint might not be so clean since EELS' `apply_body` doesn't expose those intermediary tries

cairo/ethereum/cancun/body.cairo
```diff
@@ -0,0 +1,207 @@
...
+        cast(transactions_trie.value._data.value.dict_ptr_start, DictAccess*),
+        cast(transactions_trie.value._data.value.dict_ptr, DictAccess*),
+        0,
+    );
+    default_dict_finalize(
+        cast(receipts_trie.value._data.value.dict_ptr_start, DictAccess*),
+        cast(receipts_trie.value._data.value.dict_ptr, DictAccess*),
+        0,
+    );
+    // let (post_args_commitment) = init_commitment(block, state, transactions_trie, receipts_trie, block_logs, block_hashes, gas_available, chain_id, base_fee_per_gas, excess_blob_gas);
```
Eikix, line 196: i dont think we need to commit to excess_blob_gas or base_fee_per_gas

enitrat: we do if they're the input of the next step

Eikix: they are not afaik

cairo/utils/fixture_loader.py
```diff
@@ -229,6 +257,227 @@ def load_zkpi_fixture(zkpi_path: Union[Path, str]) -> Dict[str, Any]:
...
+    transactions_trie: Trie[Bytes, Optional[Union[Bytes, LegacyTransaction]]] = Trie(
+        secured=False, default=None, _data=defaultdict(lambda: None)
+    )
+    receipts_trie: Trie[Bytes, Optional[Union[Bytes, Receipt]]] = Trie(
+        secured=False, default=None, _data=defaultdict(lambda: None)
+    )
+    Trie(secured=False, default=None, _data=defaultdict(lambda: None))
+    block_logs: Tuple[Log, ...] = ()
+
+    for address, account in state._main_trie._data.items():
```
Eikix, line 291: can you add doc / comments on those two pre-processing steps?

Eikix (approved):
frankly this is very nice, 

prepare_body_input is a big chunk of code that maybe deserves more documentation & cleaning, but otherwise this is nice

cairo/tests/ethereum/cancun/keth/test_body.py
```diff
@@ -0,0 +1,38 @@
...
+class TestMain:
+    @pytest.mark.parametrize(
+        "zkpi_path",
+        [Path("test_data/22188088.json")],
+    )
+    @pytest.mark.slow
+    def test_body(self, cairo_run, zkpi_path):
+        # Run all transactions of the body
+        body_input = load_body_input(
+            zkpi_path=zkpi_path, start_index=0, chunk_size=int(U64.MAX_VALUE)
```
Eikix, line 22: can we add a test where we start at index 6 and chunk size is like 3?

enitrat: I think it's best to do this in the final "end-to-end" test

PR #1410, opened 2025-04-30
feat: teardown.cairo

### LINKED ISSUE #1369: feat: create `teardown.cairo` program
- Start line 1130 and finish the execution of the program.
- Take as input the output of `apply_body_inner` - and put in `output_ptr` the commitment of these values. Take as input the `block.header` and `block.withdrawals`, and compute the commitment - this will later be used to glue `init.cairo` and `teardown.cairo`
- Run the final part of the execution (including trie-diff - for now)

For the root of the transaction/receipts/withdrawals tries - ensure that we compute the root with blake2s for the commitments, but with keccak256 to compute the new roots post-block update

Eikix:
closes #1369


enitrat (dismissed):
lgtm

enitrat (approved):

enitrat (approved):
gg!

cairo/ethereum/cancun/keth/teardown.cairo
```diff
@@ -0,0 +1,361 @@
...
+    // MPT diffs inputs
+    local node_store: NodeStore;
+    local address_preimages: MappingBytes32Address;
+    local storage_key_preimages: MappingBytes32Bytes32;
+    local post_state_root: OptionalUnionInternalNodeExtended;
+
+    // Fill-in the program inputs through the hints.
+    %{ teardown_inputs %}
+
+    // // Because in args_gen we want to generate a state with (prev, new) tuples, we pass an initial snapshot of the state.
```
Eikix, line 187: should we do this in args_gen instead of in the cairo code?

enitrat: we don't have a way to do this in args gen unfortunately

PR #1415, opened 2025-04-30
dev: add cairo-lint CLI

### LINKED ISSUE #1409: dev: create a python script to lint cairo imports
To automatically detect unused imports, we would like to have an automated tools that can reformat our files. One major advantage is that it lets us copy-paste a big list of imports from other files and then only keep the relevant ones, instead of thinking about what to import specifically.

Specification:

Cairo import statements are declared as:

```
from starkware.cairo.common.dict_access import DictAccess
from starkware.cairo.common.default_dict import default_dict_new
from starkware.cairo.common.registers import get_fp_and_pc, get_label_location
from starkware.cairo.common.cairo_keccak.keccak import finalize_keccak
from starkware.cairo.common.alloc import alloc

from ethereum.cancun.fork import (
    _apply_body_inner,
    state_transition,
    BlockChain,
    Block,
    keccak256_header,
    validate_header,
    get_last_256_block_hashes,
    BEACON_ROOTS_ADDRESS,
    SYSTEM_ADDRESS,
    SYSTEM_TRANSACTION_GAS,
)
```

There cannot be more than two identic identifiers in the same cairo file.

All identifiers imported can be considered unused if they're not used in the file, OR if they're not reimported from another file.
Generally, we try to avoid re-exporting imports, but there are some cases where it happens in the codebase.

As a first task, we can write a CLI script that we will invoke through `uv`, kind of like ruff:

```
uv run cairo-lint format                   # Format all files in the current directory.
uv run cairo-lint path/to/code/     # Format all files in `path/to/code` (and any subdirectories).
uv run cairo-lint path/to/file.cairo   # Format a single file.
```

Similar to ruff, running `uv run cairo-lint /path/to/file.py` will format the given file or directory in-place, while `uv run format --check /path/to/file.py `will avoid writing any formatted files back, and instead exit with a non-zero status code upon detecting any unformatted files.

For the full list of supported options, run ruff format --help.

The logic of the script can simply be, in the first iteration:

1. Collect all imported identifiers
2. Identify whether that identifier is used in the program or not (>=1 occurence besides the import in the current file
3. If not, delete it

The program will need to properly handle multi-items import lines, meaning that if an unused import is part of a larger import statement, only the unused one will be deleted, and the line will deleted.

enitrat:
Closes #1409

### DIFF
--- a/dictionary.txt
+++ b/dictionary.txt
@@ -315,3 +315,5 @@ unsquashed
 ptrs
 pyerr
 samply
+typer
+syscall
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
 
 [tool.uv]
 dev-dependencies = [
+  "cairo-lint",
   "gprof2dot>=2024.6.6",
   "jupyter>=1.1.1",
   "snakeviz>=2.2.2",
@@ -54,6 +55,7 @@ cairo-core = { workspace = true }
 eth-rpc = { workspace = true }
 keth-types = { workspace = true }
 mpt = { workspace = true }
+cairo-lint = { workspace = true }
 garaga-zero = { git = "https://github.com/kkrt-labs/garaga-zero", rev = "08e80a99c60a11adc5715c329a4390937d12d77d" }
 
 [tool.uv.workspace]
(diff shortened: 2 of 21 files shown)


Eikix (approved):
yolo

PR #1417, opened 2025-04-30
dev: add cairo-lint to trunk (and format files)

enitrat:


Eikix (dismissed):
magnificent, 
do we notice a reduction in compilation time btw?

lastly, trunk CI fails (lol)

enitrat:
> lastly, trunk CI fails (lol)

I'll need to bring this CI job back into the keth repo, otherwise, the trunk action is not properly installing the venv dependencies.

PR #1440, opened 2025-05-10
dev: move steps test in test_e2e.py

enitrat:
Moves these tests in test_e2e so as to not re-compile the same file 4 times.


Eikix (approved):
lgtm

